Issue #1, opened 2016-01-27
NPM


mickr:
Will you be releasing this as an npm module anytime soon? I would love to try the full framework out, but trying to stay away from Bower.

xeno:
+1

timothyallan:
I'm in the same boat, no bower on this box.

jgthms:
Yeah absolutely. I've never registered a npm module but it should be trivial.

jonny-improbable:
+1, kill support for bower.

SamHH:
+1

The creator of Bower actually recommended moving to npm a few months ago.

cavneb:
+1

jgthms:
Done! https://www.npmjs.com/package/bulma

PR #3, opened 2016-01-27
Create package.json to allow installing with NPM

hummlas:
Quickfix for #1: Just adding a package.json to the repository will allow it to be installed through NPM by providing the git user and repo name, e.g. `npm install jgthms/bulma --save`


jgthms:
Cheers man :+1:

PR #4, opened 2016-01-28
Update variables.sass

Kakoum:
Double $grey-lighter ?


jgthms:
Ah yes, it was for testing purposes. But you removed the wrong one!

Kakoum:
Okey :)

PR #10, opened 2016-01-28
Remove arrows from select element in IE

Ronamo:
IE handles the arrows in select elements with a pseudoclass.

More info here:

http://stackoverflow.com/questions/20163079/remove-select-arrow-on-ie


jgthms:
I think you need to indent it a step further, so that it applies to `select` (the tag), not `.select` (the class).

Ronamo:
You're absolutely correct -- dropping request and fixing.

PR #12, opened 2016-01-28
Adding modifiers on form elements.

wlepinski:
cc @jgthms

### DIFF
--- a/bulma/elements/controls.sass
+++ b/bulma/elements/controls.sass
@@ -222,3 +222,15 @@
         border-radius: 0 $radius $radius 0
     &.is-centered
       justify-content: center
+  @each $name, $pair in $colors
+    $color: nth($pair, 1)
+    $color-invert: nth($pair, 2)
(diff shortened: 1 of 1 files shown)


wlepinski:
I also added a `.control-help` class to display messages below/above inputs.

jgthms:
What about textareas and dropdowns? Also, these modifiers would make more sense directly on `.input`, `.textarea`, and `.select`, rather than on the container.

How do the hover/active/focus states look like?

I like the `.control-help` idea. I forgot to implement `<label class="label"></label>` which would have a similar role.

arkist:
also, would need to implement things like `.is-small`. @jgthms do you have any plan for?

jgthms:
@arkist `.is-small` already exists for inputs.

arkist:
@jgthms sure, but not for `select` :joy:

jgthms:
You need to rebuild the css files with `npm run build`.

PR #15, opened 2016-01-29
Bug select arrow ie

Ronamo:
This should now have the correct indentation level to apply to `<select>` as intended, and hide the dropdown arrows in IE10+


jgthms:
I can't test but I trust you =)

Schnouki:
The selector is incorrect: it should be `&::-ms-expand`, not `&::ms-expand`. Right now it doesn't work.

PR #30, opened 2016-01-31
Prevent `.buttons` being part of a user selection.

brendanmoore:
It is possible to select the text of a button, especially with a double click, this is ugly IMO.
eww

### DIFF
--- a/bulma/base/helpers.sass
+++ b/bulma/base/helpers.sass
@@ -20,8 +20,4 @@
   margin: 0 !important
 
 .is-unselectable
-  -webkit-touch-callout: none
-  -webkit-user-select: none
-  -moz-user-select: none
-  -ms-user-select: none
-  user-select: none
+  +unselectable
--- a/bulma/elements/buttons.sass
+++ b/bulma/elements/buttons.sass
@@ -15,6 +15,7 @@
 
 .button
   +control
+  +unselectable
   padding: 3px 10px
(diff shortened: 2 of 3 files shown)


inexist3nce:
Although I do like this change, I believe selecting text is the correct behaviour -- `.button` should be applied only to a `<button>` tag, so it works like a proper button and disallows user selection.

brendanmoore:
`.button` ~~should~~ **can** be applied ~~only~~ to ~~a `<button>`~~ **any** tag, so it works like a proper button and disallows user selection.

jgthms:
Yeah, this seems sensible. There's already the `.is-unselectable` helper that does exactly that. Maybe `.button` should extend that. Or maybe a mixin would do.

Also, drop the semi-colons in Sass!

KeizerDev:
If you look at http://bulma.io/documentation/elements/buttons/
On the `.is-loading` class, selection should be disabled by default because now you can simply select the loading text underneath the loading icon and at `.is-disabled` it should be `cursor: not-allowed`. 
What's your opinion about it? 

I can do a new PR for this, if you want?

brendanmoore:
@jgthms Updated. Sorry about those semi-colons, I am programmed to default to `scss`

jgthms:
@brendanmoore Great job! This is exactly how I imagined it :smiley:

PR #31, opened 2016-02-01
Add correct import statements in main entry file. This will enable to successfully import the module.

mickr:
Add correct import statements in main entry file. This will enable to successfully import the module.

### DIFF
--- a/bulma.sass
+++ b/bulma.sass
@@ -1,9 +1,9 @@
-@charset "utf-8"
+@charset "UTF-8";
 
-@import bulma/utilities/utilities
-@import bulma/config/variables
-@import bulma/config/generated-variables
-@import bulma/base/base
-@import bulma/elements/elements
-@import bulma/components/components
-@import bulma/layout/layout
+@import "bulma/utilities/utilities";
+@import "bulma/config/variables";
+@import "bulma/config/generated-variables";
+@import "bulma/base/base";
+@import "bulma/elements/elements";
+@import "bulma/components/components";
+@import "bulma/layout/layout";


msikma:
:+1: quoted imports are optional in Sass but they're better for compatibility.

jgthms:
I'm up for the quotes but not for the semi-colons in a `.sass` file.

embik:
Is there any reason for `@charset "UTF-8";`? It doesn't seem to be the "official" way and doesn't really offer anything over the current `@charset "utf-8"`. The [SASS documentation](http://sass-lang.com/documentation/file.SASS_REFERENCE.html) points to the [W3 specification for CSS encoding](https://www.w3.org/TR/2013/WD-css-syntax-3-20130919/#determine-the-fallback-encoding) when talking about `@charset`, which in turns uses `utf-8` (lowercase).

msikma:
CSS2 mandated explicitly that [UTF-8](https://www.w3.org/TR/CSS2/syndata.html#charset) be used in all caps and checked for an ASCII byte sequence. But you're right that this is changed in CSS3. The `@charset` rule is now case insensitive and `"utf-8"` is recommended by W3, so that's probably best (assuming there's no Sass compiler idiosyncrasy that needs to be taken into account).

passcod:
Shouldn't this be an issue filed on Webpack? If it has a problem importing valid Sass, it shouldn't be the responsibility of Sass users to change to fit.

msikma:
Webpack doesn't really do anything, it simply defers to a Sass preprocessor (either Libsass or Ruby Sass, usually the former as it's used by sass-loader/node-sass, which also happens to be used in the vanilla build script of this project). So the bug would very likely be in Libsass, node-sass or sass-loader.

jeffijoe:
@msikma that's not entirely true, the `sass-loader` will use Webpack's resolver, which may be incompatible with the sass way of loading things.

I am having issues with it myself.

PR #38, opened 2016-02-03
Add full-width modifier to card component

leesiongchan:


jgthms:
It's good but you need to rebuild the css and map files.

PR #41, opened 2016-02-05
fix is-withicon when used with is-grouped on controls

PuerkitoBio:
Bonjour Jeremy,

First let me say that I love that CSS framework, I'm using it for a personal project right now and it looks great (I love the fact that it is CSS-only for now - I hear you intend to add some javascript components, but I hope it stays "usable" as CSS-only).

This PR fixes a couple of issues when using a `.control` with `.is-grouped` combined with `.is-withicon`.

1. The icon on the input was disappearing on focus and hover, because of the z-index property. I set it to `4` so that it is higher than the focus and hover styles, but maybe an even higher value would make sense.

2. When using e.g. a button with an icon in the "grouped" component, and with an icon on the input, the `is-withicon` styling was bleeding through to the other `.fa`. This PR limits `is-withicon` to the `.fa` immediately following the `.input`.

To illustrate the issue, this is the HTML that displays both problems:

```
<p class="control is-withicon is-grouped">
  <input id="username" type="text" class="input">
  <i class="fa fa-user"></i>
  <a class="button is-danger">
    <i class="fa fa-times"></i>
  </a>
</p>
```

The PR fixes both issues, but disclaimer, only tested on Chrome and Safari on OSX, there may be unintended side-effects.


jgthms:
Nice one.

I think the `z-index: 4` should work.

For the bleeding through to the other `.fa`, a better fix would be `& > .fa` instead.

Also, rebuild the css and maps when submitting please. :+1:

PuerkitoBio:
Update as requested (good point for the `& > .fa` fix). I added postcss-cli to the dev deps, it was missing. Rebuilt the css (`npm run build`), apparently it did not generate a diff in the maps.

jgthms:
Perfect thanks.

PR #42, opened 2016-02-05
Improve customizability of variables when importing bulma

embik:
When importing bulma, it's not possible to set variables (located in `bulma/config/variables.sass`) without modifying bulma's original source code.

This branch allows to set any variables from `bulma/config/variables.sass` before importing `bulma.sass` by using the `!default` flag for those variables in question.

### DIFF
--- a/bulma/config/variables.sass
+++ b/bulma/config/variables.sass
@@ -1,43 +1,43 @@
 // Colors
 
-$grey-darker: #222324
-$grey-dark: #69707a
-$grey: #aeb1b5
-$grey-light: #d3d6db
-$grey-lighter: #f5f7fa
+$grey-darker: #222324 !default
+$grey-dark: #69707a !default
+$grey: #aeb1b5 !default
+$grey-light: #d3d6db !default
+$grey-lighter: #f5f7fa !default
 
-$blue: #42afe3
-$green: #97cd76
-$orange: #f68b39
-$purple: #847bb9
-$red: #ed6c63
-$turquoise: #1fc8db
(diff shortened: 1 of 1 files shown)


jgthms:
Mmmh that's an interesting one. I didn't know `!default` worked like that.
So does it update the **generated** variables correctly then?

embik:
Taking a look at the generated variables it should. My "personal" stylesheet modifies `$family-sans-serif` and the new font shows up correctly (considering `$family-primary`).

embik:
Okay I screwed up, it actually does _not_ work. Generated variables do not work correctly, I'm sorry. I'll close this PR and take a closer look.

jgthms:
No worries. Thanks for trying it out.

It would have been _so_ useful though, because customizing Bulma wouldn't require importing all files individually, like that:

```sass
@import "bulma/utilities/utilities"

@import "bulma/config/variables"

$grey-dark: #666
$grey-lighter: #f2f2f2
$blue: #08c
$primary: $blue

$family-sans-serif: "Myriad Set Pro", "Myriad Pro", "Helvetica Neue", Helvetica, Arial, sans-serif

$size-5: 20px
$size-6: 16px

$header-height: 44px

@import "bulma/config/generated-variables"

@import "bulma/base/base"
@import "bulma/elements/elements"
@import "bulma/components/components"
@import "bulma/layout/layout"
```

embik:
Yes exactly, that's what I was aiming for. Maybe I can persuade the generated variables to honor custom variables.

PR #45, opened 2016-02-06
remove duplicate line

narikei:
Is it necessary?


artichokes:
Those are not duplicate lines. The second one is lighter; therefore I'm guessing the second variable assignment should be named grey-lightest.

jgthms:
This has been fixed.

PR #51, opened 2016-02-07
Converted left, right to start, end

hitautodestruct:
Converted left, right in padding, margin and text-align to start and end respectively.
This adds support for direction independent layout i.e. support for right-to-left languages.

Just add `direction: rtl;` to the `body`.

### DIFF
--- a/css/bulma.css
+++ b/css/bulma.css
@@ -280,7 +280,7 @@ table {
 
 table th,
 table td {
-  text-align: left;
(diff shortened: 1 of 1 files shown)


jgthms:
The feature is unfortunately not supported on any Microsoft browser: http://caniuse.com/#feat=css-logical-props

Also, you changed the `.css` files, not the source `.sass` ones.

Issue #46, opened 2016-02-06
Support for a fixed header/navbar


soc:
It would be nice if Bulma would offer support for that out-of-the-box (header/navbar stays at the top when scrolling).

benfilbey:
Yes! Would be very useful.

joeblew99:
+1 me too.

gbourgneuf:
+1 especially when working with Hero layout to keep navbar and site logo available after scrolling.

MauF:
Yes would be great!

n0daft:
+1 really need this. Until then, does someone has neat workaround?

core01:
+1

pchaganti:
👍

Poliuk:
I'd love to have this. There's even a PR almost ready https://github.com/jgthms/bulma/issues/56

carsol:
+1

thulioph:
+1

gcsomesz:
+1

ImreC:
+1

Poliuk:
@jgthms can we have an official answer about this?

ChinuxParibus:
Perhaps [this](https://github.com/jgthms/bulma/pull/678) could help @jgthms

Van-Vu:
+1

Poliuk:
Thanks for trying @ChinuxParibus, I really appreciate your effort.

jeanleonino:
👍 

Any updates on this feature?

sunewa:
+1

alexsandro-xpt:
+1

jgthms:
Working on it.

PR #56, opened 2016-02-08
Allow the header/navbar to be fixed to the top of the page.

VizuaaLOG:
As requested by issue #46 this push adds the ability for fixing the header or navbar to the top of the page.

This push adds:
.header.is-fixed-top
.navbar.is-fixed-top


bulma/components/navbar.sass
```diff
@@ -7,6 +7,13 @@
       margin-bottom: 10px
 
 .navbar
+  &.is-fixed-top
+    position: fixed
+    left: 0
+    right: 0
+    z-index: 1030
```
artyomtrityak: why 1030? Bumping z-index will move to `z-index: 100000` issue

bulma/components/navbar.sass
```diff
@@ -7,6 +7,13 @@
       margin-bottom: 10px
 
 .navbar
+  &.is-fixed-top
```
artyomtrityak: should it have different behavior for mobile?

jgthms:
The header is a good idea, but the navbar has no background so it shouldn't be "fixable" at the top.

I might rename that navbar actually.

bulma/components/navbar.sass
VizuaaLOG: There isn't a reason, I just chose 1030. What z-index issue?

jgthms:
Can you create a new PR just for the header fixed on top? This PR got a bit crazy :astonished: 

And maybe add a class like `has-fixed-header` to the html so that the content doesn't end up behind the header?

jbrooksuk:
Yes please!

Bshally:
What is the implementation status of this PR? Interested in using Bulma for a project that requires a sticky header and would like to use an official component rather than creating my own if possible.

mayurtolani:
So how do you use this fixed navbar property? It does not display anything on the bulma website.

jgthms:
The source repository does not exist anymore.

PR #60, opened 2016-02-08
Update border radius on tabs when only one tab present

brendanmckenzie:
When only one tab was present the border radius was only applied to one side of the tab.  This fixes that issue.


jgthms:
Nitpick: could you order the properties alphabetically? Also: can you run `npm run build` before pushing?

Issue #65, opened 2016-02-09
Bug: navbar hamburger menu dropdown is behind secondary tab menu


C14L:
Just noticed this bug on Bulma.io when switching to narrow screen, the dropdown of the right side hamburger menu is rendered *behind* the secondary tab menu. Expected: it should cover any other element.

Screenshots: On the second image, I extended the length of the menu to see what the problem is.

http://imgur.com/a/DJND7

jgthms:
This has been fixed.

Grelek:
Hi, I use Bulma 0.1.2 and Chromium 52.0.2743.116 (64-bit) and I experience the exact same issue.

First screen:
![1](https://cloud.githubusercontent.com/assets/2447438/17817815/c5aa0222-6640-11e6-8659-45cf3578531d.png)

Second screen:
![2](https://cloud.githubusercontent.com/assets/2447438/17817816/c5c4726a-6640-11e6-89fc-c5c3796e16f6.png)

PR #69, opened 2016-02-11
Fix dropdown menu behind header secondary menu - Issue #65

fabioscsilva:
As requested by issue #65 this PR fix the hamburger menu that is rendered behind the secondary tab menu.

This is my first PR, so let me know if it's necessary to do something more.

Changes:
Remove z-index from .header


jgthms:
The `z-index` is here for the header's box-shadow (if you have a header before a hero for example).

Problem is with 2 headers, the second one will appear above, no matter the z-index of its child `header-menu`.

The whole z-index needs to be thought out I think.

fabioscsilva:
I will review all the z-index and work on a solution.

I've seen that PR #56 is adding new z-index, so maybe it's better wait for this one. And work on it after. What you think?

jgthms:
The header menu is not absolutely positioned anymore so this z-index issue is no more relevant.

PR #70, opened 2016-02-12
replace tab's margin with padding so border-bottom looks consistent

IanWang:
WOW NICE WORK!! 
thanks for making this project!

I find there's a margin-left in tabs component when it's not centered, and I guess it's not intended?

BEFORE:
before

AFTER:
after


jgthms:
It is actually intentional. The margin is there when the tabs are left aligned as well.

3noch:
I'm curios what the reasoning is.

PR #74, opened 2016-02-14
Remove shadows for input elements

nndung179:
I think the "Input elements" is going to be better without shadows.


jgthms:
I disagree. They would look too much like the buttons.

But a PR that would set it as an option is preferable.

PR #75, opened 2016-02-14
Add .editorconfig to maintain code style consistency

wolfika:
Added an [.editorconfig](http://editorconfig.org/) file to the project root, to help contributors maintain code style consistency between different editors and IDEs.

Also added the `.idea/` to `.gitignore`. This is a folder that many JetBrains IDEs (WebStorm, PhpStorm, etc.) create to store various config files. I think these are fairly common IDEs nowadays, it might be useful for those who use them.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,12 @@
+root = true
+
+[*]
+indent_style = space
+indent_size = 2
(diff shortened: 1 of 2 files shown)


jgthms:
Good idea.

PR #77, opened 2016-02-14
Using gulp to mange build and develop Bulma

nndung179:
I have added gulp to project to minified npm commands and cross-platform compatibility problems


VizuaaLOG:
There is already a  watch command 'npm run start'. Honestly, I now prefer using just NPM commands, far less work and imho more reliable.

jgthms:
Exactly. Just use npm as a build tool: http://blog.keithcirkel.co.uk/how-to-use-npm-as-a-build-tool/

nndung179:
Thanks for provide me the link and feedbacks. :)

PR #90, opened 2016-02-18
add is-text-bold

diego-vieira:
```sass
.is-text-bold
  font-weight: bold
```


juanmnl:
Is this any better than a simple <b> or <strong> html tag?
I think the framework should enforce semantics whenever possible.

PR #96, opened 2016-02-22
Fixed link to documentation in README.md

BlueAccords:
documentation link was pointing to http://bulma.io/documentation/overview/ which gave a 404 error instead of http://bulma.io/documentation/overview/start which I guess was changed to be the new overview page at some point.


jgthms:
Oh yeah forgot to update that one. Thanks!

PR #101, opened 2016-02-24
extracted white backgrounds into a variable in variables.sass

billoran8:
…to allow users to have more flexibility in their color scheme. now it's a lil bit easier to use bulma styles with different color schemes.

### DIFF
--- a/bulma/layout/section.sass
+++ b/bulma/layout/section.sass
@@ -1,5 +1,5 @@
 .section
-  background: white
+  background: $white
   padding: 40px 20px
   & + .section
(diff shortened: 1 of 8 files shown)


bulma/layout/section.sass
```diff
@@ -1,5 +1,5 @@
 .section
-  background: white
+  background: $white
```
crodriguez1a: @billoran8  I tried providing a value for $white and sections are unaffected.  Is this maybe not specific enough? Should it be section.section?

PR #102, opened 2016-02-24
Fixed space under table icons on hover.

HeliumSquid:
When you hover over the table icons, there is a small space at the bottom.

http://bulma.io/documentation/components/tables/

Use to look like this:
hover

Now looks like this:
capture


jgthms:
Thanks but the problem is that there's a space on normal links now:
screen shot 2016-02-27 at 1 32 04 pm

Still trying to figure out why it happens.

HeliumSquid:
Sorry about that. I think I've found a solution now.
It should look like this now:

capture

andrkrn:
I think this is because the `fa` class has height fixed at `24px` so there 1px space at the bottom.
When you change the height to `25px` or `100%` it should be fine.

HeliumSquid:
Thanks @andrkrn that solved the issue.

bulma/components/table.sass
```diff
@@ -16,7 +16,7 @@
       padding: 0
       & > a
         display: block
-        padding: 8px 10px
+        padding: 8.8px 10px
```
jgthms: This feels like a huge hack.

jonnyreeves: CSS you mean? ;)

seripap: +1

HeliumSquid:
Oops, sorry about that. I've now set the `.table td.table-icon .fa`'s height to `25px`.

PR #104, opened 2016-02-25
Add not-allow tip

tsq:
I think It's better to add a `not-allow tip` When the input is disabled. So the user can understand this status more clear.


jgthms:
Why not. But please remove the semi-colon. Also, you need to rebuild the `.css` and `.map` files with `npm run build`.

tsq:
:sweat_smile: I'm sorry! I forgot it.

PR #108, opened 2016-02-28
Fix outlines on close buttons for modals, notifications and tags

kolybasov:
When you click on close button on tags, notifications or modals you will get next result:
##### Tags

tags bug
##### Notifications

notifications bug
##### Modals

modals bug

This PR fix default browsers outline styles.


klaskowski:
Personally, I'd rather not delete default outline behavior. It serves a purpose of improving accessibility by clearly static which active object is focused. That is what Web Content Accessibility Guideline suggests to help people with visual impairment, and people using keyboard only (no mouse, traversing through webpage by Tab button). You might consider setting default outline style for all browsers, if consistency is required.

sapher:
I don't think it's an issue. When you click on the close button the element get destroy.

PR #119, opened 2016-03-11
Added is-borderless

seripap:
Sometimes I like button without borders, could be helpful for others.

### DIFF
--- a/bulma/base/helpers.sass
+++ b/bulma/base/helpers.sass
@@ -70,3 +70,5 @@
 .is-unselectable
   +unselectable
 
+.is-borderless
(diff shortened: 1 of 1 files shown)


jgthms:
There's the new `.button.is-link` for that.

jsgv:
@jgthms 

The `is-borderless` class would be useful for the `message` component.

Border is removed when a `message-header` div is present.

```
.message-header+.message-body {
    border-width: 0;
    border-top-left-radius: 0;
    border-top-right-radius: 0;
}
```

I want to remove border without having to add `message-header`

PR #124, opened 2016-03-15
Add commit amend to pre-push hook

Chariyski:
This way the CSS folder will be always updated and committed before push.


jgthms:
Mmmh interesting. How does the `git commit --amend --no-edit` work exactly?

Chariyski:
Hi @jgthms,

> Use the selected commit message without launching an editor. For example, git commit --amend --no-edit amends a commit without changing its commit message. 

The block-quote above is from - [https://git-scm.com/docs/git-commit ](https://git-scm.com/docs/git-commit)

lostm1nd:
I think this is quite useful and should be integrated.
Currenly the pre-push hook does no real job. As you can see most of the PRs are still lacking the built css files.

PR #125, opened 2016-03-15
Use calc instead of negative margins for centering

lostm1nd:
I could not see the benefit of using margins instead of calc.
Share your concerns if any.

Thanks,
Deyan


jgthms:
I'm not sure why you would want to process the calculation every time the element is rendered, rather than pre-process it, as it is currently the case?

In this example, the loading spinner will always be 16x16 pixels, so letting the browser calculate that on fly seems pointless in this case.

PR #127, opened 2016-03-16
Add left spaces between two buttons

guillaumemorin:


jgthms:
Not sure how I feel about this one. I feel buttons should not be aware of their surroundings.

Also, let's say you add the `.is-fullwidth` modifier, then the margin should be above, not on the left side.

In any case, you need to rebuild all the css files before submitting a PR.

PR #134, opened 2016-03-24
Remove negative margins on gapless columns container

lostm1nd:
Not clearing top and bottom negative margins on .columns.is-gapless leads to undesired result. Check out the screenshot.

gapless-issue


jeppech:
I would really like this as well! Why is this necessary to do?

jgthms:
Happy to merge this but there's conflicts to resolve.

lostm1nd:
Fixed in v.0.0.18

PR #136, opened 2016-03-24
Add a option to justify text

Karklis:
Add a helper class to justify text.


jgthms:
Justified text is harder to read, as the line ends are all similar, and I don't want to promote that. :smile:

PR #141, opened 2016-03-25
Fixes #107: Inline-flex properly calculates its content size

### LINKED ISSUE #107: Tabs component alignment and small screens
Center aligned tabs can end up with their leftmost item cut off. Right aligned tabs get no scrollbar.

lostm1nd:
Block element does not overflow the tab container.


jgthms:
What is the purpose behind this one? It prevents the tab border to go across the whole width.

lostm1nd:
Could you share a screenshot of this because I do see the border across all the tabs.
tabs-border

jgthms:
Should be like this:

screen shot 2016-03-27 at 11 27 01 am

lostm1nd:
Totally overlooked this. Another way to stretch the .tabs ul is to use some of the proposed width properties (min-content does the job) but they are not that well supported.

jgthms:
The tab structure has changed.

PR #147, opened 2016-03-30
Added `!default` to variables

thegitfather:
Hi there! Just a small addition that I think is only useful.

Great work so far!
cu

### DIFF
--- a/bulma/config/variables.sass
+++ b/bulma/config/variables.sass
@@ -1,52 +1,52 @@
 // Colors
 
-$grey-darker: #222324
-$grey-dark: #69707a
-$grey: #aeb1b5
-$grey-light: #d3d6db
-$grey-lighter: #f5f7fa
+$grey-darker: #222324 !default
+$grey-dark: #69707a !default
+$grey: #aeb1b5 !default
+$grey-light: #d3d6db !default
+$grey-lighter: #f5f7fa !default
 
-$blue: #42afe3
-$green: #97cd76
-$orange: #f68b39
-$purple: #847bb9
-$red: #ed6c63
-$turquoise: #1fc8db
-$yellow: #fce473
+$blue: #42afe3 !default
+$green: #97cd76 !default
+$orange: #f68b39 !default
+$purple: #847bb9 !default
+$red: #ed6c63 !default
+$turquoise: #1fc8db !default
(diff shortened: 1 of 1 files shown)


jgthms:
What's the purpose exactly?

thegitfather:
This way I can keep your files in my bower_components folder untouched and still overwrite the vars within my style project folder. If I don't set a var of yours then your value will be the default. So no harm, only useful

jgthms:
What about the generated variables? Are they still correctly set, even if you set _your_ custom variables _after_?

https://github.com/jgthms/bulma/pull/42 had the same idea but it didn't work.

thegitfather:
To test this I swapped `$yellow` with `$red` in my `_vars.scss`:
```scss
$temp: $yellow;
$yellow: $red;
$red: $temp;
```
and expected two buttons (`button.button.is-warning` and `button.button.is-danger`) to switch the resulting style. And it worked..

So only important thing is to import the project vars before bulma, e.g.:
```scss
@import 'global-scss/vars';
@import 'global-scss/mixins';
@import 'global-scss/extends';
@import 'global-scss/fonts';
@import 'global-scss/layout';

@import '../bower_components/bulma/bulma';
...
```

jgthms:
You need to add `!default` to `generated-variables.sass` as well.

thegitfather:
Hmm not sure what setup you've tested.. As I see it, the generated variables get assigned with values from the variables file only. So it should be ok to add the `!default` statement only to the variables.sass.. At least it works just fine for my project (`"gulp-sass": "^2.0.1"`)

jgthms:
What about if someone updates the `$primary` color? Or the `$text` one, **before** including bulma?

jbrooksuk:
It wouldn't work, because Bulma would override it anyway?

thegitfather:
It works. E.g. if i add the line `$primary: red;` to my project `_vars.scss` then all links will be red. And as i stated before you always have to set your project vars *before* including bulma.

> `!default` will assign a value only if the variable doesn't already have one. In other words, it prevents a variable from being reassigned.

thegitfather:
has been merged

jgthms:
Hey @thegitfather thanks for your help btw. I restructured all the variables so I took the opportunity to add the `!default` as well.

PR #160, opened 2016-04-03
Add Bower to README install instructions

feimosi:


README.md
```diff
@@ -17,6 +17,12 @@ Bulma is in early but active development! Try it out now:
 npm install bulma
 ```
 
+### Bower
+
+```sh
+bower install bulma
```
jbrooksuk: Maybe `bower install --save bulma`

feimosi: I know, but I wanted to stay consistent with the `npm` part

jbrooksuk: Which should be updated too IMHO.

feimosi: That stays up to @jgthms. I can do that on a request.

jgthms:
At first, Bulma only used bower. But people suggested that npm was the way to go now.

I left the bower support as legacy but intend to remove it completely very soon, as even the creator of bower thought npm was a better option.

feimosi:
@jgthms I can understand your point, but I really wouldn't remove it. There're plenty of projects that are forced to use bower (I know a few) and it doesn't cost you any effort to keep it. 
With bower, you have that convenience, it automatically updates the package based on Git tags.

jbrooksuk:
I actively use Bower :)

PR #199, opened 2016-05-04
Fix loader in buttons when they are not default

necinc:
When I want to use .is-loading class name with button which has the .is-danger or other class names the color of text can not become transparent. It is the simplest solution to fix that trouble.

I mean that earlier was like this:
before

And now it is something like this:
after


necinc:
Any other changes have occurred after the `npm run build`.

jgthms:
Thanks!

PR #201, opened 2016-05-06
Fixed link to cdnjs bulma overview

yaph:
The previous link returned a 403 error.


jgthms:
Ah thanks, I missed that.

PR #216, opened 2016-05-26
nav-center to work without any other nav

Wizjonero:
Changes proposed:

Currently if you want to use nav-center you have to use all three nav classes (nav-left, nav-center, nav-right). If you use only nav-center it will be aligned to left. If you use two classes - it will be pushed either to the right or to the left. With just one line of Sass you can use just nav-center and it works perfectly well.
- [x] Add
- [x] Fix
- [ ] Remove
- [x] Update


ShaggyDude:
Sometimes two lines is better than one.

margin-left: auto;
margin-right: auto;

don't touch top and bottom margins for no reason...

Wizjonero:
@ShaggyDude You're 100% right, sir! Updated. :)

jgthms:
It's already there.

PR #240, opened 2016-06-10
Fix hero bodies having 0 height in IE

t2t2:
Changes proposed: Adds a not-null flex-basis value to .hero-body to fix IE showing it as 0 height

Before:

bulma-ie-hero-before

After:

bulma-ie-hero-after
- [ ] Add
- [x] Fix
- [ ] Remove
- [ ] Update


kvanlierop:
I'm not sure if the is the same issue or not, but I noticed that changing

```
.hero-body {
flex: 1;
}
```

to 

```
.hero-body {
flex: 1 1 auto;
}
```

makes `.hero-body` function properly in IE 11. Apparently IE doesn't read `flex: 1` the same as other browsers. No surprise there.

PR #255, opened 2016-06-24
Add help inverted style

gfpacheco:
Changes proposed:

Add help inverted style (`help is-inverted`).
- [x] Add
- [ ] Fix
- [ ] Remove
- [ ] Update


bb010g:
Is there any reason this is still open?

jgthms:
Ah yes it's invalid because it has the built CSS in the PR.

And now that I think about it, having `is-inverted` for the `help` element doesn't make sense because "inverted" is for an element that has both a background color and a color, and inverting swaps both.

In this case, using a color helper is the solution.

PR #295, opened 2016-09-04
Center icons properly on buttons that only have one icon

mcampo:
### Pull Request

Buttons that have only an icon (one icon and nothing else) are not centered properly.

``` html
<p className="control">
  <a className="button">
    <span className="icon"><i className="fa fa-smile-o"></i></span><span>Text on the right</span>
  </a>
</p>
<p className="control">
  <a className="button">
    <span>Text on the left</span><span className="icon"><i className="fa fa-smile-o"></i></span>
  </a>
</p>
<p className="control">
  <a className="button">
    <span className="icon"><i className="fa fa-smile-o"></i></span>
  </a>
</p>
```

Styles before the change:
image

Styles after the change:
image

Changes proposed:
- [ ] Add
- [x] Fix
- [ ] Remove
- [ ] Update

Fix the styles for icons that are the single child of a button, so they appear centered on the button.

PD: I'm not sure where all the `-webkit-flex` changes on the built css come from, maybe I got a newer version of the autoprefixer when I installed the dependecies?


gperdomor:
@mcampo, maybe the icon button should be squared instead of rectangular. WDYT?

mcampo:
@gperdomor that could be a posibility, but I don't really mind the button being rectangular as long as the icon is properly centered.

Issue #308, opened 2016-09-19
Docs: Table example preview differs from what is in the example code


petrvolny:
### Checklist
- [x] This is about the Bulma **Docs**
### Description

Preview of the Table example does not match the code in the documentation (table is filled with open source projects but there are some musicians and their instruments in the example code)
### Steps to Reproduce

Check http://bulma.io/documentation/elements/table/

PR #309, opened 2016-09-19
Unify table example preview with code

petrvolny:
Simply fixes the table example docs. (#308)

Not sure how the actual website page is being generated after the migration from the `bulma-website` repo so let me please know if I make some other changes ;)

Cheers


jgthms:
Thanks! Can you fix the indentation though?

PR #366, opened 2016-10-28
Adds a new sticky footer class

judetucker:
### Pull Request

Fixes https://trello.com/c/o4WBYLih/11-add-sticky-footer

Changes proposed:
- [x] Add

This creates a `.sticky` class that can be used with footer to create a sticky footer and updates the docs to reflect it. This was one of your TODOs from Trello. Hope this helps!


VizuaaLOG:
I think it should be ```.is-sticky``` as it is a modifer and should follow the Bulma naming convention.

optikfluffel:
@judetucker Could you rebase on current master and make it `.is-sticky`? Would be really cool to have this one in 😸

docs/css/bulma-docs.css
```diff
@@ -6542,6 +6538,14 @@ a.panel-block:hover {
   border-bottom-color: #00d1b2;
 }
 
+.footer .sticky {
```
cheshire137: Shouldn't this be `.footer.sticky` since you're expecting the `sticky` class to be on the same element that has the `footer` class?

sobers-2002:
@judetucker Thanks for this!

However, I noticed some issues when using this:

1. If you are using hero with is-fullheight, with multiple scroll locations on the page, the footer will pop into all the screen. Not sure if this is intended when it's being sticky, but I would expect that a sticky footer should be at bottom of the screen even if the content on the page does not make it go down to the bottom

2. It can overlap other elements on the page (for eg. a form's input boxes) if they didn't warrant a scroll bar of their own

It would be nice if it would work properly as I believe this would be required by a lot of people :+1:

cheshire137:
The approach I usually use for a sticky footer is this:

```css
body {
  display: flex;
  min-height: 100vh;
  flex-direction: column;
}

#root {
  flex: 1 0 auto;
}
```

Where `#root` is the outer-most child of `body` that contains the rest of the content. It's the approach used by [Materialize](http://materializecss.com/footer.html).

surgiie:
Ive always had issues with this method @cheshire137. I use VueJs for some of my projects and usually make my outer-most child of body the div that lets Vue hook into the markup `#app`. For some reason this:

```
<div id="app">
      //navbar
       //content
      //footer
 </div>

```

and 

```
body {
  display: flex;
  min-height: 100vh;
  flex-direction: column;
}

#app {
  flex: 1 0 auto;
}


```

will not work for me. The footer does not stick.  This will however work:


```
//move header/nav outside
<div id="app">
      
       //content
     
 </div>
//move footer outside
```

The only problem with this is vue js wont be able to hook into any vue code i may have in markup outside of `#app`. I usually just use absolute positioning for a sticky footer. 


```

body
{
	min-height:100%;
}


footer
{
	position: absolute;
	right: 0;
	left: 0;
	bottom: 0;
}
```

Id honestly prefer to get the flex way to work for me as bulma is flex based but this is what i do for now:/

Hope it helps someone!

kimbaudi:
@cheshire137 - Your approached worked for me until I made my navbar fixed (to either the top or bottom). In that case, I had to change from `min-height: 100vh` to `min-height: calc(100vh - #{$navbar-height})` where `$navbar-height` is the sass variable for navbar height. So in my case, I modified the styling (I'm using sass):

```css
body {
  &.default {
    display: flex;
    min-height: 100vh; // use this if navbar is not fixed (default)
    flex-direction: column;
  }
  &.fixed {
    display: flex;
    min-height: calc(100vh - #{$navbar-height}); // use this if navbar is fixed
    flex-direction: column;
  }
}

main {
  flex: 1 0 auto;
}
```

hope this helps anyone else who plans to use fixed navbars and still get sticky footers

PR #367, opened 2016-10-30
Add a modifer to allow the navbar to be fixed to the top of the page

VizuaaLOG:
Finally gotten round to redoing this commit. This is related to #56 and #46 .

This adds a `.is-fixed-top` modifer to the `.nav` component. Fixing it to the top of the page. There is also a `.has-fixed-nav` modifer which adds a `padding-top` of the navs height. Wanted to put this modifer in the helpers file, but as it depends on the `$nav-height` variable this was not possible.

The modifer does not fix the nav when on a mobile device.


flexd:
Any progress on this being merged in? :)

wilomgfx:
+1 yes please :)!!

sass/components/nav.sass
```diff
@@ -131,10 +131,22 @@ a.nav-item
...
     & > .container,
     .container > &
       & > .nav-left
         & > .nav-item.is-brand:first-child
           padding-left: 20px
+
+.has-fixed-nav
+  padding-top: $nav-height
+  +mobile
+    padding-top: 0
```
jgthms: Why no padding on mobile?

sass/components/nav.sass
```diff
@@ -131,10 +131,22 @@ a.nav-item
...
+      position: static
   // Responsiveness
   +touch
     & > .container,
     .container > &
       & > .nav-left
         & > .nav-item.is-brand:first-child
           padding-left: 20px
+
+.has-fixed-nav
```
jgthms: This should only affect the `html` element.

sass/components/nav.sass
```diff
@@ -131,10 +131,22 @@ a.nav-item
   // Modifiers
   &.has-shadow
     box-shadow: 0 2px 3px rgba($black, 0.1)
+  &.is-fixed-top
+    position: fixed
+    left: 0
+    right: 0
+    top: 0
+    +mobile
+      position: static
```
jgthms: The behavior should be similar on mobile.

sass/components/nav.sass
```diff
@@ -131,10 +131,22 @@ a.nav-item
   // Modifiers
   &.has-shadow
     box-shadow: 0 2px 3px rgba($black, 0.1)
+  &.is-fixed-top
+    position: fixed
+    left: 0
```
jgthms: Properties need to be alphabetically ordered.

sass/components/nav.sass
VizuaaLOG: I don't know about this. I think that on mobile the nav bar should return to a normal non-sticky state. Otherwise, you'll be using up precious screen space.

sass/components/nav.sass
VizuaaLOG: So, replace this with `html.has-fixed-nav`?

sass/components/nav.sass
VizuaaLOG: Due to the nav bar returning to a non-sticky state on mobile, padding-top is not required.

nicolastakashi:
Some news?

octoshrimpy:
@jgthms and @VizuaaLOG I think that the navbar should have an option of sticking it to the top on mobile. best of both worlds.

Hillcow:
Any news?

VizuaaLOG:
Due to the changes in Bulma core, and the `Nav` component being deprecated/removed in favor of the new (and much better) `Navbar` component, this PR is no longer valid.

PR #398, opened 2016-11-29
Use tablet layout for printing

### LINKED ISSUE #363: Printing uses mobile layout
### Description

Since all responsive media queries are formed along the lines of `@media screen ...`, printing a page defaults to the mobile-first, everything is vertical style, making most of the layout features unusable. There is also no way to override this with configuration. 

Are there any plans regarding this? Including `@media print` in the `tablet` mixin seems like a good fit.

yscik:
Forgot about making a pull request for this earlier. It adds the `print` media type to the `tablet` mixin, so that horizontal layout is used for printing. (No min-width condition as it's not very useful for paper pages)

Fixes #363


hdra:
Hi, any update if this is going to get merged? I happen to be working on something that could make use of this change...

jgthms:
Can you only update `sass/utilities/mixins.sass` and I'll merge this.

yscik:
Sure, I removed the generated files.

jgthms:
🤔 There seems to be a conflict for this single line change.

yscik:
Maybe the revert isn't working out, I've remade the PR against the current master without the generated files. #606

PR #402, opened 2016-12-05
Not-allowed cursor added to disabled buttons

caiokawasaki:
Now it will be indicated to the user when a button is disabled.

Changes proposed:

* [ ] Add
* [ ] Fix
* [ ] Remove
* [x] Update


jgthms:
Not allowed !== not able to

I believe it's a matter of opinion, but I don't want to impose this on Bulma users.

PR #418, opened 2016-12-22
Update columns.html

ekanna:


jgthms:
I want to keep the code and the style separate.

PR #438, opened 2017-01-04
Unsplashed

Epistol:
### Pull Request

Fixes #

Updated from placehold.it which is unavailable from unsplash.it

Changes proposed:

* [ ] Add
* [x ] Fix
* [ ] Remove
* [ ] Update


jgthms:
Thanks. I've actually decided to host the images myself instead of relying on a third party service.

PR #450, opened 2017-01-09
[fix] Remove explicit font-size for ".select select"

kkkrist:
This was overwriting the font size inherited from =control and therefore causing inconsistencies, see screenshots.

Before:
Before

After:
After

Changes proposed:

* [ ] Add
* [x] Fix
* [ ] Remove
* [ ] Update


jgthms:
It's needed when sizing the `select` parent:

image

kkkrist:
I see. Thanks for looking into it!

PR #454, opened 2017-01-10
adding back old templates as a placeholder

yawetse:
### Pull Request

Fixes

Changes proposed:

* [x] Add
* [ ] Fix
* [ ] Remove
* [ ] Update


jgthms:
As they are not ready, it is better to leave them out for now.

PR #459, opened 2017-01-12
fix select element with is-expanded class

### LINKED ISSUE #458: select is-expanded not working as expected
* [x] This is about the Bulma **CSS framework**
* [x] I'm using Bulma **version** [0.3.0]
* [x] My **browser** is: Chrome 55.0.2883.95 (64-bit) Mac OSX

### Description

The wrapping span.select is expanded but the actual select element is kept at its current width, so it ends up looking like this:

screen shot 2017-01-12 at 11 22 13

### Steps to Reproduce

```html
<p class="control has-addons"><span class="select is-expanded"><select  name="something"><option value="xxxxx">Xxxxx</option></select></span><button type="submit" class="button">Run</button></p>
```

### Expected behavior

The select element is correctly stretched to match the span.select

dalaidunc:
to fill the full width of its parent span

### Pull Request

Fixes #458

Changes proposed:

* [x] Fix

### DIFF
--- a/sass/elements/form.sass
+++ b/sass/elements/form.sass
@@ -200,6 +200,8 @@ $input-radius:              $radius !default
         select
           border-radius: 0 $input-radius $input-radius 0
       &.is-expanded
+        select
+          width: 100%
         flex-grow: 1
         flex-shrink: 0
     .select select


dalaidunc:
Why is this invalid? I searched PRs/issues and could not find a duplicate.

jgthms:
You need to rebuild the css files.

dalaidunc:
Ah ok, sorry about that.

I have pushed some changes now to address that and to match the SASS style guide.

PR #465, opened 2017-01-14
Add Text Align Justify to helpers

DavidLambauer:
Added Text Align Justify to helpers


jgthms:
I feel like justified text is quite unreadable. I don't want to support that behavior.

PR #476, opened 2017-01-18
Update docs to match the current version

yoostk:
#### Update documentation

- Use `.nav` instead of `.header` in the layout/container section

`.header` was deprecated and renamed in 0.0.26 

- Remove `/documentation/grid.html` and `documentation/navbar.html`

Each is now `/documentation/grid/columns` and `/documentaion/components/level/`. 
They seem to be not in use. But currently you can visit http://bulma.io/documentation/grid/ and http://bulma.io/documentation/navbar/.


yoostk:
BTW, guides / steps for pull requests to the documentation would help.

jgthms:
Nice 👍 

Yeah I'm gonna add some instructions for doc PRs. But if it's not CSS changes, you can just update the HTML as it's rebuilt on push anyway.

PR #501, opened 2017-02-01
add 'is-widescreen' modifier on the 'columns' container

badalsurana:


sass/grid/columns.sass
```diff
@@ -243,9 +243,13 @@
     align-items: center
   // Responsiveness
   +tablet
-    &:not(.is-desktop)
+    &:not(.is-desktop):not(.is-widescreen)
```
jgthms: This is not scalable in this case. It should be `is-tablet`.

sass/grid/columns.sass
```diff
@@ -243,9 +243,13 @@
...
       display: flex
   +desktop
     // Modifiers
     &.is-desktop
       display: flex
+  +widescreen
```
jgthms: We'll need `is-desktop-only` and `is-widescreen-only` and `is-fullhd`. now.

sass/grid/columns.sass
badalsurana: if i make it `is-tablet` and remove all `:not()` columns will not be activated by default from tablet onwards.

sass/grid/columns.sass
badalsurana: ok i will also add these 3 modifiers.

prograhammer:
When can we see this added? Widescreen is super super common. I have a lot of use-cases where I need to stack columns if below widescreen.

Love Bulma, such great work (will have to drop some $$ gratitude as I get deeper into using it)!

mathvdh:
I have had some success by changing column.sass responsiveness part to :

`  // Responsiveness
  +tablet
    &:not(.is-desktop):not(.is-fullhd):not(.is-widescreen)
      display: flex

  +desktop
    &.is-desktop:not(.is-fullhd):not(.is-widescreen)
      display: flex

  +widescreen
    &.is-widescreen:not(.is-fullhd)
      display: flex

  +fullhd
    &.is-fullhd
      display: flex`

fredDesign:
i tried to use some class like is-10-widescreen without success . 
Have you fixed that breakpoint , or have i to do it localy ? 
Thx

fabiobulgarella:
News about the merge of 'is-widescreen' modifier on .columns class? Thanks.

janosrusiczki:
I just ran into a situation when I want to stack columns if viewport is below widescreen. If this was not merged since 2017 is it possible to use some workaround?

michaeljsalo:
I'm using Bulma 0.9.0 and finding that 'is-widescreen' modifier does not work on 'columns' class.  is-mobile, is-tablet, and is-desktop all work as expected.  is-widescreen has no effect.

Issue #508, opened 2017-02-07
Body not using $background color in v0.3.1


reustonium:
The html body doesn't seem to be setting the background-color: $background in version 0.3.1

This is about the Bulma **CSS framework**
I'm using Bulma **version** [0.3.1]
My **browser** is: Chrome Version 55.0.2883.87

I am sure this issue is **not a duplicate** but it might be related to #504 

### Description

I'm not sure if this is supposed to be a feature or if it's a bug.

### Steps to Reproduce

1. Create an html page using Bulma version 0.2.3
2. Observe that the `background-color: whitesmoke;`

![image](https://cloud.githubusercontent.com/assets/1432186/22712491/f3762b74-ed52-11e6-9d3f-6afd91815ce8.png)

3. Create an html page using Bulma version 0.3.1
4. Observe that the `background-color: white;`

![image](https://cloud.githubusercontent.com/assets/1432186/22712295/483cb142-ed52-11e6-94ab-c4e2b513f53d.png)

### Expected behavior

I would expect the `<html>` background-color should be equal to `$background` which defaults to whitesmoke.

### Actual behavior

The `<html>` background-color equals white.

yoostk:
The `<html>` background-color is defined as `$body-background` variable, which is white color. I think this is supposed to be a future. The `$background` variable is used for other elements.

https://github.com/jgthms/bulma/blob/master/sass/base/generic.sass#L5

ForsakenHarmony:
yeah, some of the other elements are not using $background though, which I find really annoying

jgthms:
`$body-background` is for the page background.
`$background` is the light grey background used for some elements.

cyrilchapon:
And $body-background is not in derived-variables.scss 😋

PR #509, opened 2017-02-08
Fixes Issue 508 - background-color definition to match implementation

reustonium:
This short fix updated the documentation to match the implementation of `$body-background` to set the `<html>` `background-color` property.

https://github.com/jgthms/bulma/blob/master/sass/base/generic.sass#L5

References [Issue 508](https://github.com/jgthms/bulma/issues/508)

### DIFF
--- a/docs/documentation/overview/variables.html
+++ b/docs/documentation/overview/variables.html
@@ -40,7 +40,7 @@ doc-subtab: variables
         <li>
           <strong>Generated variables</strong> where variables are <strong>calculated</strong> from the values set in the previous file. For example, you can have:
           <ul>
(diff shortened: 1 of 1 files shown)


yoostk:
`$body-background`'s color is not `$white-ter` but `$white`. `$background`'s color is `$white-ter`. You can check those variables in [sass/utilities/variables.sass](https://github.com/jgthms/bulma/blob/master/sass/utilities/variables.sass).

reustonium:
good catch, I updated the color, hopefully this is more clear.

Thanks!

ForsakenHarmony:
I feel like you should also add the old one back and make it clear that it's for other elements

PR #511, opened 2017-02-08
Prefix for classes

### LINKED ISSUE #302: please namespace all classes
Please **namespace** all your classes with the prefix "bulma-" or "b-", because there are a lot of conflicts when using other css frameworks as well. For example, common css-class names such as .card, .panel etc are in conflict. So please namespace all the classes with "b-" ("b-card", "b-panel", etc.).

Thanks.

hgezim:
### Proposed solution
Fixes #302 

### Tradeoffs
More complex docs.

### Testing Done

It does work.


jgthms:
I've thought about this one. Although it can be useful, I feel it adds too much noise in the source code.

But the main thing is that namespacing should not be solved this way, but through components, or CSS modules.

PR #518, opened 2017-02-14
Prevent Chrome to suggest translation from Latin

termosa:
Add `lang` attribute to `html` tag


noformnocontent:
Lorem ipsum Dolores was here

PR #541, opened 2017-03-05
Fixes #336 - is-gapless columns and offset in column

### LINKED ISSUE #336: is-gapless columns and offset in column
### Checklist
- [ ] This is about the Bulma **Docs**
- [x] I'm using Bulma **version** 0.2.1
- [x] My **browser** is: Firefox 49 Ubuntu 16.04
- [ ] This is a **Sass** issue: I'm using version [x.x.x]
- [x] I am sure this issue is **not a duplicate**?
### Description

Using gapless columns and a column with an offset will cancel the offset. I believe this is related to CSS precedence.

These are the rules that are applied by the browser. The `margin-left` on the offset class is cancelled by `is-gapless`.

```
.columns.is-gapless > .column {
    margin: 0;
    padding: 0;
}
.column.is-offset-half-desktop {
    margin-left: 50%;
}
```

In my case I wanted to have the offset on desktop but not on mobile.
### Steps to Reproduce

The following code will replicate the issue

```
<div class="container is-fluid">
    <div class="columns is-gapless">
        <div class="column is-half-desktop is-offset-half-desktop">Offset</div>
    </div>
</div>
```
### Expected behavior

I expect the offset to be respected and maintained. `is-gapless` should only remove the margins and paddings and classes related to positioning should not be affected by the removal of margins.
### Actual behavior

Using gapless columns and a column with an offset will cancel the offset making the column move to the left and destroying the layout.

Thank you ;)

Saboteur777:
### Proposed solution
Fixes #336 - is-gapless columns and offset in column by narrowing the targeted 

### Tradeoffs
No that I can think of. 

### Testing Done
Yes.

### DIFF
--- a/sass/grid/columns.sass
+++ b/sass/grid/columns.sass
@@ -222,7 +222,7 @@
       margin-bottom: 0
     &:not(:last-child)
       margin-bottom: 1.5rem
-    & > .column
+    & > .column:not([class*="is-offset"])
       margin: 0
       padding: 0
   &.is-grid


Saboteur777:
@jgthms why invalid or duplicate?

cyrilchapon:
What is the status of this ?

Benjaminsson:
I'm also experiencing problems with this. A fix would be very welcome.

augnustin:
:+1: for fixing this.

JonPugsley:
If you want to work around this, use an empty column

So

`<div class='column is-8 is-offset-4'>content</div>`

Becomes

`<div class='column is-4 is-dummy'></div><div class='column is-8'>content</div>`

PR #546, opened 2017-03-06
Add new feature: Dropdown menu

gaillota:
Adding a dropdown menu (inspired by Bootstrap's one).


robin900:
Would you be willing to add a Dropdown item to the Components`docs/` with documentation and example code?

robin900:
I seem to notice that in Bulma, "elements" are via a single class (with one or more modifiers) applied to a single element; "components" involve a class for the topmost tag and then additional classes for child tags. I think your Dropdown qualifies as a "component" since it involves multiple classes to style the child tags properly.

gaillota:
Actually, I didn't where to put it at first between "elements" and "components". I thought the "elements" was for the stylized HTML elements but then I saw the boxes.
Anyway, thanks for the clarification, I will update this ASAP.
And yes, I can update the documentation as well, with some examples.

davmrtl:
See PR https://github.com/jgthms/bulma/pull/546

gaillota:
Question: What is the template engine used for the documentation ? 

I've blindly updated the `docs` with some examples but I would like to verify that everything is fine before I can push.

mightyiam:
@gaillota maybe take a look at the scripts:
https://github.com/jgthms/bulma/blob/0395dc59d8b147f2f47d057a6ffde2eb2966db49/package.json#L29-L42

gaillota:
I have, and there's nothing regarding the `docs` templates. Only for building sass.

robin900:
@gaillota It's jekyll. There's a PR to add how-to for the documentation, including how to view the local documentation for your local checkout.

https://github.com/jgthms/bulma/pull/561/files

gaillota:
Nice, thank you ! Updating ASAP

sass/components/_all.sass
```diff
@@ -1,6 +1,7 @@
 @charset "utf-8"
 
 @import "card.sass"
+@import "dropdown"
```
Remo: You might want to add `.sass` here for consistency reasons, great PR though!

gaillota:
I finally checked the documentation rendering and I think it's explicit enough on how to use dropdowns.

wolfgang42:
:+1: for this&mdash;it looks great! The one thing I've added for my use is a drop shadow on the list to make it stand above the background, like this:


I copped this style from Bootstrap:
```css
.dropdown-list {
	box-shadow: 0 6px 12px rgba(0,0,0,.175);
}
```

gaillota:
I thought of that, but I choose not to because it would not suit with the overall aspect of Bulma. But it was

gaillota:
I thought of that, but I chose not to because it wouldn't suit the overall aspect of Bulma IMO. 
It was clearly a personal call but if you think it may be better, why not.

wolfgang42:
I don't have any opinion on the way it should look&mdash;borrowing Bootstrap's shadow was simply the most expedient way to get *something.* The problem is that there's some other stuff in my app right underneath the dropdown's button, so when you open the menu there's a lot of clutter underneath; the drop shadow makes it much easier to see what's going on.

davmrtl:
I think it could be possible to reuse the box-shadow of the .box element, which is currently:
```css
.box {
    box-shadow: 0 2px 3px rgba(10, 10, 10, 0.1), 0 0 0 1px rgba(10, 10, 10, 0.1);
}
```
This way, it would be compatible with the overall aspect of Bulma.

gaillota:
Yes it could be, but it wouldn't make sense to use the `box` class along with `dropdown-list`.
Plus, if the `box` class gets other CSS rules added someday, it would affect the dropdown component as well, and I don't think that's a good idea.

**However** we could make the shadow optional by adding another class like`has-shadow`, which would, when coupled with `dropdown-list`, add a `box-shadow` property to the list.

What do you guys think ?

GustavsR:
Any plans on merging this?

Log1x:
Looking forward to this getting merged.

Nezt thing to tackle would be a Mega Menu with mobile support. Usually if the mobile is a mega, the mobile fallback would be a toggle able sidebar navigation.

jgthms:
Needs to be rebased against master before I can review this.

ForsakenHarmony:
yeah, this is way to big

gaillota:
I agree with you, this isn't very clean. I going to try to clean this up when I get home.

gaillota:
Cleaned everything up.

gaillota:
Hi @Wikiki 
Sorry for the delay,

1) You're right, I didn't think of an icon-only nav-item. What do you suggest then, move the arrow to a different class, for example `has-arrow`, which can optionally be added to the dropdown button ?
2) It actually felt inappropriate to use ìnline-block` when Bulma is based on flexbox. Will change to ìnline-flex`ASAP.
3) There's a class for that, `dropdown-list-right` which align the list along with the right side of the button. Is that what you meant ?

Thank you for your suggestions.

sass/components/dropdown.sass
```diff
@@ -0,0 +1,126 @@
...
+  &:disabled
+    background-color: transparent
+    color: $dropdown-link-disabled-color
+    cursor: not-allowed
+
+.dropdown-list-right
+  left: auto
+  right: 0
+
+.dropdown-list-left
```
ludo237: Shouldn't this be `has-list-left` ? @jgthms

sass/components/dropdown.sass
```diff
@@ -0,0 +1,126 @@
...
+    background-color: darken($dropdown-bg, 2.5%)
+    text-decoration: none
+
+  &.is-disabled,
+  &:disabled
+    background-color: transparent
+    color: $dropdown-link-disabled-color
+    cursor: not-allowed
+
+.dropdown-list-right
```
ludo237: Same issue as [The above one](https://github.com/jgthms/bulma/pull/546/files#r111597211)

RafaelKeramidas:
Any hope of seeing this getting merged? This is something I would love to see in Bulma

ludo237:
@RafaelKeramidas maybe next tag

sass/components/dropdown.sass
gaillota: Waiting for an answer to finalize work :) @jgthms

wilburpowery:
Any plans on merging this request @jgthms ?

jgthms:
I just reviewed the code. Thanks for your contribution but I don't like it. This code is basically copied from Bootstrap, without attributing it to the original authors, which goes against the open source mindset.

It uses floats, which is barely used in Bulma. And the caret should be an arrow, because that's what Bulma uses.

In any case, I'm closing just because it's not good enough for Bulma. I will probably implement the dropdown myself.

PR #549, opened 2017-03-08
Fixes #490: Removes padding-related directives for .delete nested selectors

### LINKED ISSUE #490: V0.3.1 - Close Button Is Not Displayed Correctly in IOS Safari
It is about Bulma.
It is most likely a bug.

### Overview of the problem
I'm using Bulma **version** [0.3.1]
My **browser** is: IOS Safari

### Description
Close button is not displayed correctly in IOS Safari, as shown in the images attached below. Images are captured on Bulma Official Documentation Website using IOS Safari.

### Steps to Reproduce
1. Find an IOS Device with latest version of IOS and Safari
2. Browse http://bulma.io/documentation/components/modal/ OR http://bulma.io/documentation/elements/notification/

### Expected behavior
Close button is expected to be shown correctly, as per below (These images are captured using Chrome):

### Actual behavior
Close button is not shown correctly in latest version of IOS Safari:

robin900:
padding is inherited by delete selectors inside components and elements that need their own padding directives.
The inherited padding for nested .delete cause no problems for
desktop browsers, because their user agent stylesheets have
overriding directives for padding for the relevant <button>, etc.
But on iOS Safari there are no such directives.

Fixes jgthms/bulma#490.

### Proposed solution

Fixes weird delete element appearances (elongated from top-left to bottom-right) when delete element inside tag, notification, or message.

### Tradeoffs

I don't know. I don't see any padding directives that are relevant for .delete elements, so I thought zeroing out the padding directives was the best way forward.

### Testing Done

Solution confirmed via xCode iOS Safari simulator; no regression on macOS Chrome desktop browser.


jgthms:
I'm not sure in which case the `.delete` element inherits the padding, since it's usually a simple `<a>` or `<button>`?

In any case, the reset should be set on `.delete` itself, not on each instance of its use.

PR #555, opened 2017-03-09
Fixes #535: default button conforms to is-outlined, is-inverted, and combined

### LINKED ISSUE #535: Button .is-outlined does not work without a color declaration.
I'm using Bulma **version** [0.3.2]
My **browser** is: Chrome 55

### Description

Assigning an element the class .button.is-outlined does not work for the default button - it remains with a white background. It requires a colour class in order to be applied.

Find below an image from the Bulma documentation but with the background changed to black.

screenshot 2017-02-27 at 11 40 43 am

### Steps to Reproduce

1. Create a link element
2. Assign the class .button & .is-outlined
3. Use a coloured background

### Expected behavior

Like all other .is-outlined elements this should have a transparent background

### Actual behavior

The element has a non-transparent background

robin900:
### Proposed solution
Default button style did not seem to address the requirements of `is-outlined`, `is-inverted`, and the combination `is-inverted is-outlined`. 

Fixes #535, which merely asked that default button style look different when modified by `is-outlined`, but issue poster mixed that up with `is-inverted`. The definition of the modifier styles as written in the documentation are:

* `is-outlined`: transparent background, and the principal color of the button style is used for border and text. hover/focus revert the styling back to non-outlined styling.
* `is-inverted`: background color made into the text color, and vice versa. border color becomes transparent.
* `is-inverted is-outlined`: background color becomes the text and border colors; background color becomes transparent.

Documentation is updated to display the default button style for the three modifier applications in an illustrative way. I also included `&:is-loading` sections where they were omitted for  `is-outlined` and added an extra row of loading buttons to illustrate.

### Tradeoffs
Only concern is that users might have somehow used the default style with is-outlined/inverted/both and somehow wanted the previous, undocumented behavior.

### Testing Done

Browsing local version of docs/_site.


amir20:
@jgthms I just ran into this issues where `.is-outlined` doesn't work unless some color defined. Any chance of getting it merged?

niuage:
I'm curious as to why this hasn't moved forward, it's been such a long time. Buttons are pretty core to a CSS framework ;)

iliaparvanov:
Transparent white buttons are still broken :(

PR #573, opened 2017-03-17
fix ie select default style

superwf:
fix ie select tag default style


jgthms:
You should replace the `&::ms-expand` instead.

jgthms:
Done: https://github.com/jgthms/bulma/pull/619

PR #608, opened 2017-03-28
Fixes #227 - prevent $colors names from being minified

### LINKED ISSUE #227: .button.is-white becomes .button.is-#fff
In bulma.css for example the `.button.is-white` is correct but in the minified version all `.is-white`, `is-black`, ... is converted to `is-#fff`, `is-#000`, ... 

You should check the minifier settings to prevent that.

quagliero:
### Proposed solution
By setting `white` and `black` as strings in the $colors map in variables.scss, external minifiers/compilers don't class them as literal colours and change them to `#fff` and `#000`.

This allows usage of, e.g. `.button.is-white` as expected, instead of the `.button.is-#fff` that is built.

### Tradeoffs
N/A

#### Before:
before-devtools
before-minified

#### After:
after-devtools
after-minified

### DIFF
--- a/sass/utilities/variables.sass
+++ b/sass/utilities/variables.sass
@@ -146,6 +146,6 @@ $size-large: $size-4 !default
 ////////////////////////////////////////////////
 // 4. Lists and maps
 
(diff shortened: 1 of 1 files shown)


jgthms:
Needed double quotes and might as well do it for all.

PR #610, opened 2017-03-31
Padding not right

bykof:
If I try to build bulma with react-create-app build the select field doesn't look properly.
Somewhere it overrides the padding-right to 0.625em and then it looks gross.


jgthms:
I can't add an `!important` statement, as it will mess up other people's styles. Priority should be fixed in context. You need to investigate what part of your app overrides that padding.

PR #614, opened 2017-04-01
Remove incorrect/unclear documentation

maxstr:
### Proposed solution
Documentation currently suggests importing `variables.sass` to override default colors and then import all of Bulma. This strategy doesn't work as the initial setting of `$colors` will not be overwritten on the second import.

This commit simply removes that suggestion from the documentation.

### DIFF
--- a/docs/documentation/overview/start.html
+++ b/docs/documentation/overview/start.html
@@ -98,9 +98,6 @@ npm install bulma
         <p class="title is-5">
           <strong>Set</strong> your variables:<br>
 {% highlight sass %}
-// Import Bulma's initial variables
-@import "bulma/sass/utilities/variables.sass"
-
 // Override initial variables here
 // You can add new ones or update existing ones:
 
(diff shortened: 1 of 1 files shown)


ForsakenHarmony:
it shouldn't override colours, because they are marked with `!default`

whatever you use to transpile it seems to be at fault

but the variables.sass file should be split out imo

maxstr:
The current documentation recommends importing `variables.sass`, setting some colors, and then importing all of Bulma. You will overwrite the specific color variables, eg `$primary`, `$turquoise`, but you won't override `$colors`(because it already got set the first time and is marked with ~default) which is necessary for the colors to be used when generating the classes.

ForsakenHarmony:
oh, yeah that, that's why I want it split out, I had to copy paste that into my own code again

maxstr:
The fix may also be to remove `!default` from the colors tag. It seems unlikely that you would want to manually modify that one.

PR #615, opened 2017-04-01
Fix/276

jgthms:


dvLden:
The anchor does not support `disabled` attribute, so this is just a styling fix applied where you're targeting a button class with disabled attribute... `.button[disabled] { ... }`

With the styling you applied, it looks like the button is disabled, but would still need to be handled with JS to prevent re-submitting or so. 

`pointer-events: none` - did great job if you ask me, one should not care about the cursor styling, if the anchor (button) has been clicked and then disabled on/for purpose...

Issue #298, opened 2016-09-08
breadcrumb


ManUtopiK:
Any way to make a breadcrumb ?

dashed:
Wondering about this as well.

pchaganti:
👍

Mubeen-Inamdar:
👍

xereda:
👍

jgthms:
It's shipped thanks to @vinialbano

PR #632, opened 2017-04-03
Add new feature: Breadcrumb component

vinialbano:
### Proposed solution

Add a breadcrumb component, as requested in https://github.com/jgthms/bulma/issues/298
It was based on bootstrap v4's breadcrumb and includes automatically created separators in the ::before pseudo-elements.
I also created modifiers for alternatives separators, which is originally a forward slash "/".

### Tradeoffs

The component wraps for big breadcrumbs on mobile. I haven't found a better way to solve it.

### Testing Done

screen shot 2017-04-02 at 10 11 14 pm
I tested this solution on google chrome for mac os. I also made tests using lists instead of nav elements and it works well.

### DIFF
--- a/docs/_includes/subnav-components.html
+++ b/docs/_includes/subnav-components.html
@@ -1,6 +1,9 @@
 <nav class="nav has-shadow">
   <div class="container">
     <div class="nav-left">
+      <a class="nav-item is-tab {% if page.doc-subtab == 'breadcrumb' %}is-active{% endif %}" href="{{ site.url }}/documentation/components/breadcrumb/">
+        Breadcrumb
+      </a>
       <a class="nav-item is-tab {% if page.doc-subtab == 'card' %}is-active{% endif %}" href="{{ site.url }}/documentation/components/card/">
         Card
       </a>
--- a/docs/_layouts/documentation.html
+++ b/docs/_layouts/documentation.html
@@ -43,7 +43,7 @@ route: documentation
             <a href="{{ site.url }}/documentation/elements/box/">Elements</a>
           </li>
           <li {% if page.doc-tab == 'components' %}class="is-active"{% endif %}>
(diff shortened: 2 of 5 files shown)


jgthms:
Pretty good. Thanks for making a proposition.

I feel like it could be a simple `ul` though, without the `breadcrumb-item`. The same way tabs work, with only a single CSS class. And I'm not sure if flexbox is really needed here (only to remove empty spaces I guess).

I'd like to see how it looks like on several lines.

Also, it doesn't follow the [Sass styleguide](https://github.com/jgthms/bulma/blob/master/.github/CONTRIBUTING.md#bulma-sass-styleguide).

vinialbano:
I made the suggested corrections and included the overflow-x in the container as used in tabs so it doesn't need to wrap.

ludo237:
This is really good

KevinSjoberg:
Is this good to go?

emielmolenaar:
Really nice!

vinialbano:
I am just waiting for @jgthms feedback. If any improvement is needed, let me know!

jgthms:
Great job @vinialbano ! Merging it and then I'll clean up a bit if needed.

Rainbowlemon:
Could you bump the version number? I've been sat here scratching my head for 10 mins 'cause my version is the same as the repo's yet I can't make breadcrumbs! (Great addition btw, just in time for a project I'm working on).

AdelMahjoub:
Not working, but it is on the docs page.

AlgusDark:
@AdelMahjoub actually, we need to bum the version number, you can check the [repo css](https://raw.githubusercontent.com/jgthms/bulma/master/css/bulma.css), although version is 0.4.2, `.breadcrumb` is there, but it's not on [CDNJS](https://cdnjs.com/libraries/bulma) because of the version name. 

@jgthms should we move the version for the deployed css to 0.5.0?

PR #650, opened 2017-04-07
Add RTL support (fix #6)

### LINKED ISSUE #6: Don't forget about RTL direction please

mohammadrafigh:
### Proposed solution
This PR adds support for RTL languages and layout and resolves issue #6.

### Tradeoffs
the only noticeable thing is after this PR maintainers should consider using $right and $left variables to add bidirectional styles. actually where there is a differentiate in styles like margins, paddings, etc.

### Testing Done
I tested this PR with the included docs project and everything worked as expected including RTL texts and layouts. but docs should be updated for RTL languages. for example in docs we should show how to right align and left align (both) something in a level. currently we show only the opposite alignment (right) for LTR languages so it doesn't sense when we use RTL and there we need to show "left" alignment too. also I didn't reverse the meaningful classes so .*-left(right)-* and .*-left/right-* classes are working as expected. because of this behavior if you see card and media component in RTL in docs it looks weird, becuase the example uses left alignment and the classes in html docs should be changed to right alignment when we are using RTL. so media-right should be used instead of media-left.

if you need more help with RTL or anything I can do in the future let me know.


bulma-rtl.sass
```diff
@@ -0,0 +1,10 @@
+/*! bulma.io v0.4.0 | MIT License | github.com/jgthms/bulma */
+@charset "utf-8"
+
+$right: left
+$left: right
+
+html
+  direction: rtl
```
mightyiam: Directionality should be declared in the content itself, not as a style, I think.

bulma-rtl.sass
```diff
@@ -0,0 +1,10 @@
...
+
+@import "bulma"
```
mightyiam: I'm not sure that having this extra file is at all a good approach.

sass/components/level.sass
```diff
@@ -46,6 +46,8 @@
   +block
   align-items: center
   justify-content: space-between
+  @if ($right == 'left')
+    flex-direction: row-reverse
```
mightyiam: Thank god for sass.

sass/components/nav.sass
```diff
@@ -42,10 +42,7 @@ a.nav-item
   &.is-tab
     border-bottom: 1px solid transparent
     border-top: 1px solid transparent
-    padding-bottom: calc(0.75rem - 1px)
-    padding-left: 1rem
-    padding-right: 1rem
-    padding-top: calc(0.75rem - 1px)
+    padding: calc(0.75rem - 1px) 1rem
```
mightyiam: To get rid of the words `right` and `left`. Got it.

sass/utilities/variables.sass
```diff
@@ -1,151 +1,155 @@
...
+$orange:       hsl(14,  100%, 53%) !default
+$yellow:       hsl(48,  100%, 67%) !default
+$green:        hsl(141, 71%,  48%) !default
+$turquoise:    hsl(171, 100%, 41%) !default
+$blue:         hsl(217, 71%,  53%) !default
+$purple:       hsl(271, 100%, 71%) !default
+$red:          hsl(348, 100%, 61%) !default
+
+// Typography
+$family-sans-serif: -apple-system, BlinkMacSystemFont, "Segoe UI", "Roboto", "Oxygen", "Ubuntu", "Cantarell", "Fira Sans", "Droid Sans", "Helvetica Neue", "Helvetica", "Arial", sans-serif !default
```
mightyiam: What's going on here? Is there a change here, at all?

sass/utilities/variables.sass
```diff
@@ -1,151 +1,155 @@
...
+$size-medium: $size-5 !default
+$size-large: $size-4 !default
+
+////////////////////////////////////////////////
+////////////////////////////////////////////////
+// 4. Lists and maps
+
+$colors: (white: ($white, $black), black: ($black, $white), light: ($light, $light-invert), dark: ($dark, $dark-invert), primary: ($primary, $primary-invert), info: ($info, $info-invert), success: ($success, $success-invert), warning: ($warning, $warning-invert), danger: ($danger, $danger-invert)) !default
+
+$sizes: $size-1 $size-2 $size-3 $size-4 $size-5 $size-6 !default
```
mightyiam: What's up with this file? Why is it full of changes? It would be kind of hard to review like this.

bulma-rtl.sass
mohammadrafigh: thanks for reviewing @mightyiam .

moving this into base modules then I should apply @if to add this directional style only when we are making an rtl version. but if you want I can do that.

bulma-rtl.sass
mohammadrafigh: this is the way many css libs out there releasing RTL versions. generally we cannot have the smart BIDI behavior only with css and the BIDI is only applicable we the powers of js. so with css only approach we have 2 types of implementations:
* making a separate RTL version which will have the same KBs in size but it's standalone and there is no need to include the LTR version too. (my approach is this)
* another way to handle that is making an override css file so we should include LTR version in page and then override styles for RTL with another file. which I see is more dirty than first approach since you need to include bulma.css and bulma-rtl.css both in page. also overriding the styles are not optimal (computational and rendering wise).

sass/components/level.sass
mohammadrafigh: sorry didn't get your meaning :)

sass/utilities/variables.sass
mohammadrafigh: oops, sorry in this file I just added:
```
// Direction variables
$right: right !default
$left: left !default
```

but my IDE changed CR LF line break things...

sass/utilities/variables.sass
mohammadrafigh: I reverted CRLF to LF again so only these lines are new:

```
// Direction variables
$right: right !default
$left: left !default
```

bulma-rtl.sass
mightyiam: Both versions don't sound so pretty to me. Perhaps there's another way?

In the future, the solution would be that CSS for both directions is included and is chosen using the [`:dir`](https://developer.mozilla.org/en-US/docs/Web/CSS/:dir) pseudo selector.

Right now, what we could possibly do is similar, with the difference of selecting on `html[dir=rtl]`. How do you feel about that?

- Most RTL pages on the web will have `<html dir="rtl">`, right?
- Bulma is CSS for components. Components in an RTL page will always be RTL, right? There's no common reason to mix both RTL and LTR styled components, I think.
- Bulma has a component for styling content. I don't think it has any directionality-relevant properties, right? It is probably simmetrical.

So, if Bulma would "just work" for RTL when the HTML element has `dir="rtl"` perhaps that would be best?

If a slight increase in size is an issue, that could be resolved by splitting it into some imports. But that would create so many new files... I think not worth it.

Thoughts?

sass/components/level.sass
mightyiam: We don't have conditionals in CSS, right?

sass/utilities/variables.sass
```diff
@@ -27,8 +27,8 @@ $red:          hsl(348, 100%, 61%) !default
 
 // Typography
 $family-sans-serif: -apple-system, BlinkMacSystemFont, "Segoe UI", "Roboto", "Oxygen", "Ubuntu", "Cantarell", "Fira Sans", "Droid Sans", "Helvetica Neue", "Helvetica", "Arial", sans-serif !default
-$family-monospace: monospace !default
-$render-mode: optimizeLegibility !default
+$family-monospace: monospace !default
+$render-mode: optimizeLegibility !default
```
mightyiam: Seems like there are still meaningless changes here. But I think there's a bigger discussion going on.

bulma-rtl.sass
mohammadrafigh: your provided approach is nearly similar to what I did. the only difference between using html[dir=rtl] and using $right $left variables is in result and the implementation of both is the same. the result of my approach is 2 separated css files after build (bulma.css for LTR and bulma-rtl.css for RTL that library users just need to include one of these in the html page) and your approach will result in just 1 bulma.css which will have a bigger size (as you mentioned too) and includes both styles for RTL and LTR. actually my separation happens in SASS to CSS transpilation process and your separation happens in rendering time in browser.

I don't know if you are the one who can accept my PR or not (since you are not a member or contributor). so for now I accept your solution as an alternative and if the maintainers want me to edit my changes with this solution then I will update my PR. please let me know if you are the maintainer then I will apply your solution to the PR otherwise we can wait for a decision from maintainers on this.

finally in my opinion both approaches (mine and yours) have some drawbacks and I can't prefer one to another and I can't say one of them is pretty and another is not. I think all of these are dirty solutions until we have something like direction: auto; in css. we already have a dir=auto attribute in html and it's almost perfect but it's only applicable on divs/inputs and textual contents but doesn't affect in layout. by the way here I can reject your thoughts on: 

> Most RTL pages on the web will have <html dir="rtl">, right?

I can provide many examples that dir="auto" is the most used value for dir. so your approach is not usable in these cases.

sass/utilities/variables.sass
mohammadrafigh: as I said the only change in this file was only adding 2 variables that I mentioned and the IDE switched LF to CRLF so I reverted this IDE change in this commit! and as you can see at the top of the PR page the dirty lines are less than first commit now

bulma-rtl.sass
mightyiam: I'm just a user here.

The benefit of a single file is simplicity. In the case of a multilingual site, some pages are LTR and others are RTL. It is a more general approach to have just a single file, it seems to me. Feels more correct to me.

I didn't even know about the `auto` value.

I'd go with a single file, for compatibility and simplicity.

That
> something like direction: auto; in css

that we're waiting for, I think it is the [`:dir` selector](https://developer.mozilla.org/en-US/docs/Web/CSS/:dir).

mohammadrafigh: ok. for example I can say Twitter is the most known BIDI example. you can see support both for LTR and RTL on the same page. but they did this using js by switching between dir="ltr" and dir="rtl". ok I agree with you about :dir but as you said we need to wait (maybe 1-2 years at least).

that's my pleasure that you reviewed my work and I learnt about :dir now so we can use it later :) . so for now lets wait for maintainers to decide.

lordadamson:
Thank you @mohammadrafigh 
Could you please review and merge that @jgthms ? It would make my life much easier because now I have to patch this manually and stuff. thank you :D

seifsay3d:
@jgthms are you planning on adding rtl support to bulma's core?

bulma-rtl.sass
myGuruPitka: > Components in an RTL page will always be RTL, right? There's no common reason to mix both RTL and LTR styled components, I think.

@mightyiam actually there is a case for LTR components in RTL page.
for example, an email input field should be always LTR, even if the page is RTL.
this can be done by decorating the field specifically with `direction:ltr`.

dralshehri:
I think using two variables `$right` and `$left` for users to switch between LTR/RTL is a little bit confusing. Inspired from [Zurb Foundation for Sites 6](https://github.com/zurb/foundation-sites/blob/develop/scss/_global.scss) who has a long experience in RTL support, it would be better to enable RTL by one `$text-direction` variable and let sass get the right direction for `$right` and `$left` based on `$text-direction` value.

```sass
$text-direction: ltr;
$left: if($text-direction == rtl, right, left);
$right: if($text-direction == rtl, left, right);
```

SomeoneCares11:
Accept this please

mightyiam:
I stand by [my assessment on this](https://github.com/jgthms/bulma/pull/650#discussion_r110660088).

SomeoneCares11:
@mightyiam Check semantic RTL http://rtlcss.com/cdn/css-frameworks/semantic-ui/index.html
 and bootstrap RTL, they are very nice!! bulma must have the same.

mohammadrafigh:
"@SomeoneCares11 and everyone here". here is my fork you can use it by `npm install bulma-rtl` currently you can use it with sass modules only give me a few hours and I will update this fork with latest bulma changes and I may send another PR with a different approach. otherwise if the Bulma repo maintainers don't want to merge this PR then you can use bulma-rtl from tomorrow and so on... I try to fix rtl related issues there. let me know your thoughts. also I may do some tweaks and add some custom components later if I have time.

https://github.com/mohammadrafigh/bulma-rtl

kodeine:
Is this going to be merged?

eladcandroid:
Any news?

yitzhakbg:
Where does Bulma RTL stand today?

mohammadrafigh:
I'm going to update it tonight

madmadi:
We are still waiting...

yakovyarmo:
any updates?

sorousht:
@mohammadrafigh We are waiting for RTL

mohammadrafigh:
@sorousht @yakovyarmo @madmadi 
you can use https://github.com/mohammadrafigh/bulma-rtl  (`npm install bulma-rtl`) it's seems this PR is not going to be accepted.

yazeed:
> I think using two variables `$right` and `$left` to switch between LTR/RTL is a little bit confusing. Inspired from [Zurb Foundation for Sites 6](https://github.com/zurb/foundation-sites/blob/develop/scss/_global.scss) who has a long experience in RTL support, it would be better to enable RTL by one `$text-direction` variable and let sass get the right direction for `$right` and `$left` based on `$text-direction` value.
> 
> ```sass
> $text-direction: ltr;
> $left: if($text-direction == rtl, right, left);
> $right: if($text-direction == rtl, left, right);
> ```

I think it would be great if we can adopt the same approach on bulma, it's much simpler than the proposed solution in this PR.

yazeed:
Nice update @mohammadrafigh

mohammadrafigh:
0.8.0 is still WIP. I will release it in next few hours.

yazeed:
Check this guide for a proper RTL implementation
https://www.rtlstyling.com/

jgthms:
> Check this guide for a proper RTL implementation
> https://www.rtlstyling.com/

That's a fantastic resource. I'm going to work on a `rtl-support` branch.

mohammadrafigh:
Due to RTL support in Bulma 0.9.0 I close this PR

PR #480, opened 2017-01-20
Fixes #469 | fixing group buttons overlap on safari

andrewwwcollins:
Issue: #469 

Issue on both Bulma css and docs. Issue caused on Safari, IE11, and IOS where grouped buttons will overlap each other caused by a flex-basis of 0, changed to flex-basis auto with the initial fallback for partially supported auto property. Reference: https://github.com/jgthms/bulma/issues/469

- Added in 0 fallback for partially supported “auto” property

Issue #469, opened 2017-01-16
Button Group does not work on Safari nor IE 11


jeromu:
### Description

Bulma bug report

### Steps to Reproduce

Using iPad, or IE11:
1. Go to http://bulma.io/documentation/elements/button/
2. Scroll down to either section that describe button groups

### Expected behavior

Display properly

### Actual behavior

Buttons overlap

PR #659, opened 2017-04-10
Fix button width in is-grouped form fields (IE)

thijsvdanker:
### Proposed solution

When the `.button` class is used in a `.is-grouped` form control, the button width is wrong in IE11.
button-in-is-grouped

Earlier issue (https://github.com/jgthms/bulma/issues/469) and PR (https://github.com/jgthms/bulma/pull/480) where closed as the form structure was changed. This PR applies cleanly to the new structure.

### Tradeoffs
None that I can think of.

### Testing Done
Tested on my local machines with IE11 (Windows 7), Chrome, Safari and Firefox (on OS X).

### DIFF
--- a/sass/elements/form.sass
+++ b/sass/elements/form.sass
@@ -217,6 +217,7 @@ $input-radius:              $radius !default
     justify-content: flex-start
     & > .control
       flex-basis: 0
+      flex-basis: auto
       flex-shrink: 0
       &:not(:last-child)
         margin-bottom: 0


jgthms:
Fixed by removing it instead of overriding it: https://github.com/jgthms/bulma/commit/ffc66a0270e49a137c605805342a67ca416eb391#diff-ce3c2ee7e05f002260f45062123ff5c9

Issue #449, opened 2017-01-09
class "columns" too wide on mobile


larsvansche:
* [ ] This is about the Bulma **CSS framework**
* [ ] I'm using Bulma **version** [0.2.3]
* [ ] My **browser** is: Google Chrome
* [ ] This is a **Sass** issue: I'm using version [3.4.9]

On mobile, the element with class "columns" is wider than the actual viewport, making the website slightly horizontally scrolling. 
The element is given a negative margin on all sides except the bottom side, and the child element (class="column")

goodbytes-gb:
What do you have wrapped around your `.columns` container?

Bulma's containers are in 'sections' .. If you were to do something like:

```
    <div class="container">
        <div class="columns">
            <div class="column">
                Lorem ipsum dolor sit amet, consectetur adipisicing elit. Accusamus dolor doloribus esse ex expedita minus officiis voluptas. At consequuntur debitis dignissimos eum sed voluptas. Amet assumenda cupiditate earum molestias placeat.
            </div>
        </div>
    </div>
```

It would create a horizontal scroll bar. However if you were to do:

    <div class="section">
        <div class="container">
            <div class="columns">
                <div class="column">
                    Lorem ipsum dolor sit amet, consectetur adipisicing elit. Accusamus dolor doloribus esse ex expedita minus officiis voluptas. At consequuntur debitis dignissimos eum sed voluptas. Amet assumenda cupiditate earum molestias placeat.
                </div>
            </div>
        </div>
    </div>

Then the horizontal scrollbar would be fixed.

See documentation: http://bulma.io/documentation/layout/section/

stuartcusackie:
I'm getting a similar issue. Using a section solves the overlap but adds unwanted padding. If I use CSS to set the section padding back to zero then the overlap comes back. Having a padding of zero on the parent div clashes with the negative margins of the .columns div.

I got around this by overriding the padding on mobile columns like so:

```
.columns {
	margin: 0;
}

@media screen and (min-width: 768px) {

	.columns {
	    margin-left: -0.75rem;
	    margin-right: -0.75rem;
	    margin-top: -0.75rem;
	}

}
```

tudoanh:
Thanks @stuartcusackie, work like a charm.

frederikhors:
Any news on this?

And using this code instead?

```css
@media screen and (max-width: 1024px) {
  .columns {
    margin-left: 0;
    margin-right: 0;
  }
}
```

[...]

PR #660, opened 2017-04-10
Columns too wide on mobile

thijsvdanker:
### Proposed solution
When using `.columns` on a mobile device, the margin-right: -0.75rem causes the body to overflow.
This makes it possible for the user to scroll and see whitespace.
By changing the horizontal padding and margin on mobile, this issue is resolved.
Fixes: https://github.com/jgthms/bulma/issues/449

### Tradeoffs
People relying on the horizontal padding for mobile will lose this.

### Testing Done
Works on Chrome (OS X)

### DIFF
--- a/sass/grid/columns.sass
+++ b/sass/grid/columns.sass
@@ -242,6 +242,8 @@
   &.is-vcentered
     align-items: center
   // Responsiveness
+  +mobile
+    margin: 0
   +tablet
     &:not(.is-desktop)
       display: flex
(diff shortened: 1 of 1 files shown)


sass/grid/columns.sass
```diff
@@ -242,6 +243,9 @@
   &.is-vcentered
     align-items: center
   // Responsiveness
+  +mobile
+    margin-left: 0
+    margin-right: 0
```
jgthms: Bulma is mobile-first. So this should be undone on `tablet`.

sass/grid/columns.sass
```diff
@@ -41,6 +41,7 @@
     .columns.is-mobile > &.is-offset-#{$i}
       margin-left: ($i / 12) * 100%
   +mobile
+    padding: 0.75rem 0
```
jgthms: Not sure what this is for.

PR #666, opened 2017-04-11
Fixes #640 - Fix help text in .has-addons

### LINKED ISSUE #640: Help message for input with addons
### Overview of the problem
- [x] I'm using Bulma **version** [0.4.0] 
- [x] My **browser** is: Chrome
- [x] I am sure this issue is **not a duplicate**?

### Description
Showing a help message for a input with addons is not showing correctly in a horizontal form. As you can see in the picture below it it acts like the text is an addon. If it's possible to fix this with some extra html it would be nice to find it back in the docs.

### Steps to Reproduce
1. Create horizontal form
2. Add input with addon
3. Show help message.

### Expected behavior
The help message is shown under the input just like the rest of the form.

### Actual behavior
The help message is shown as an addon.

TNovalis:
### Proposed solution
`flex-wrap: wrap` on `.has-addons`, with `width: 100%` and `content: ''` in an after pseudo-selector. Any selector such as `.help` with `order: 1` (in commit) will be put on the next line.

### Tradeoffs
N/A

With this solution, no extra HTML is required to fix this and after testing I have found that no styling is affected as that only selectors with `order` set to 1 or greater are moved.

### Testing Done
**Before:**
before

**After:**
after

### DIFF
--- a/sass/elements/form.sass
+++ b/sass/elements/form.sass
@@ -156,6 +156,7 @@ $input-radius:              $radius !default
   display: block
   font-size: $size-small
   margin-top: 0.25rem
+  order: 1
   @each $name, $pair in $colors
     $color: nth($pair, 1)
     &.is-#{$name}
@@ -172,6 +173,7 @@ $input-radius:              $radius !default
   // Modifiers
   &.has-addons
     display: flex
+    flex-wrap: wrap
     justify-content: flex-start
     .control
       margin-right: -1px
@@ -204,6 +206,9 @@ $input-radius:              $radius !default
       &.is-expanded
         flex-grow: 1
         flex-shrink: 0
+    &::after
+      width: 100%
+      content: ''
     &.has-addons-centered
       justify-content: center
     &.has-addons-right


sass/elements/form.sass
```diff
@@ -156,6 +156,7 @@ $input-radius:              $radius !default
   display: block
   font-size: $size-small
   margin-top: 0.25rem
+  order: 1
```
jgthms: This is not needed.

sass/elements/form.sass
```diff
@@ -204,6 +206,9 @@ $input-radius:              $radius !default
       &.is-expanded
         flex-grow: 1
         flex-shrink: 0
+    &::after
```
jgthms: Remove this. You want `.field .help{ width: 100%;}` instead.

ReToCode:
Any updates on this issue? Seems like an easy fix, and lots of people having the issue.

balping:
> Seems like an easy fix

No, not really.  As you can see on the screenshot above, the textbox on the right hand side hasn't got rounded corners. The problem is that `:last-child` is used several times and that selector applies to the  help text instead of the input. Maybe a solution would be to replace it with `:nth-last-of-type()` together with other clever filters.

ReToCode:
Yeah, you are right. But still, I think it's better that it's on a new line even when the corners of the textbox are not round. I fixed my app by adding:

```
    .field .help {
        width: 100%;
    }

    .field.has-addons {
        flex-wrap: wrap;
    }
```

tggo:
any progress?

PR #678, opened 2017-04-15
Proposed solution to issue #46

### LINKED ISSUE #46: Support for a fixed header/navbar

ChinuxParibus:
### Proposed solution
Added `.is-fixed-top` and `.is-fixed-bottom` to get navbar fixed to top or bottom.

Fixes #46 

### Tradeoffs
This way, `.container` needs a margin top or bottom equals to `$nav-height`, because using `position: fixed;` rule gets content overlapped on top or bottom respectively.

### Testing Done
**Before: Fixed on top. Content gets overlapped (same with `.is-fixed-bottom`)**
without-adjust
**After: Fixed on top. Content gets a margin equals to `$nav-height`**
with-adjust

### DIFF
--- a/sass/components/nav.sass
+++ b/sass/components/nav.sass
@@ -119,3 +119,14 @@ a.nav-item
   // Modifiers
   &.has-shadow
     box-shadow: 0 2px 3px rgba($black, 0.1)
+  &.is-fixed
+    position: fixed
+    width: 100%
+    &-top
+      top: 0
+      & + .container
+        margin-top: $nav-height
+    &-bottom
+      bottom: 0
+      & + .container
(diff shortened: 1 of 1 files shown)


jgthms:
That is not the correct solution. This only works if a `.container` follows directly a `.nav`.

ChinuxParibus:
A simple workaround. Instead of:
```
...
  & + .container
```
Could be:
```
...
  & + *
```
On both rules. It works on build with Sass.

jgthms:
Performance-wise, using the universal selector `*` is quite bad.

Issue #293, opened 2016-09-03
Bug: .has-text-centered does not work on thead tr th


EdJoPaTo:
### Checklist
- [x] This is about Bulma the **CSS framework** and **not** about the Bulma website
- [x] I'm using Bulma **version** [0.1.2]
- [x] My **browser** is: Chromium, Firefox
- [x] I'm using the precompiled version from the cdn
- [ ] I am sure this issue is **not a duplicate**?
### Description

I'm using `.has-text-centered` within a table on `thead tr th` and on `tbody tr td`. In the `tbody` this does work as expected. In the `thead` the `.has-text-centered` got overridden by `.table th` and is left aligned.
### Steps to Reproduce

![screenshot from 2016-09-03 14-33-08](https://cloud.githubusercontent.com/assets/7953011/18224832/5f941302-71e3-11e6-9e10-9f1ebae9b95c.png)

``` html
<table class="table">
  <thead>
    <tr>
      <th class="has-text-centered">
        Small Header
      </th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td>
        Very big and long table cell to show the text alignment better
      </td>
    </tr>
    <tr>
      <td class="has-text-centered">
        Small Cell
      </td>
    </tr>
  </tbody>
</table>
```

_Expected behavior:_ the th should be centered.

_Actual behavior:_ the th is aligned left.

Njoror:
The only solution I could think of is a css hack where you increase specificity by chaining classes. For example if the css selector was .has-text-centered.has-text-centered it would increase the specificity enough to get past the .table th. Not elegant at all though. Well aside from an !important but generally not a fan of those. Helper classes do make a compelling case for it sometimes. :/

jgthms:
Yes the specificity of the cells is higher than the helper classes. What you can do is wrap the content of the cell in a separate element:

```
<td>
  <p class="has-text-centered">
    Small Cell
  </p>
</td>
```

Njoror:
Okay that works. Kind of like what Twitter Bootstrap recommends. I do want to note that that fix only works on block elements and does not work on inline elements. What's your take on !important'ing utility classes? It's one of the few decent arguments I've seen for an !important and one could argue that if you're applying a helper class to the markup you probably don't want it overriden.

jgthms:
The `text-align` property only works on block elements.

For the `!important` thing, most of the helpers have that, but text alignment gets applied to _all_ children, so it can have unexpected consequences. That's why it's not there.

[...]

PR #748, opened 2017-05-17
Make alignment modifiers !important

aheuermann:
### Proposed solution
Currently the text alignment modifiers cannot be applied to table headers.

```html
<table class="table">
  <thead>
  <tr>
    <th class="has-text-centered">Header</th>
  </tr>
  </thead>
  <tbody>
    <tr>
      <td class="has-text-centered">
        Data
      </td>
    </tr>
  </tbody>
</table>
```

### Tradeoffs
I realize `!important` is a bit of a no-no, but I think this usage is ok as you are deliberately applying the class to an element.

### Testing Done
Tested this branch on an private app I am developing.

### DIFF
--- a/sass/base/helpers.sass
+++ b/sass/base/helpers.sass
@@ -33,10 +33,10 @@ $displays: 'block' 'flex' 'inline' 'inline-block' 'inline-flex'
   +clearfix
 
 .is-pulled-left
-  float: left
+  float: left !important
 
 .is-pulled-right
-  float: right
+  float: right !important
 
 // Overflow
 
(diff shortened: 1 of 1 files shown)


aheuermann:
I mis-spoke - this does not fix #656; however it does solve my issue of wanting to apply these helper classes to elements and make sure they are applied and not overridden due to specificity of other styles.

timacdonald:
I'd agree with this as these are atomic helpers. If you look at a fully atomic CSS framework like [Beard](https://github.com/monarkee/beard) it's all set with `! Important`

timacdonald:
Looks like this might not be a thing. This is a related issue: #293

jgthms:
I think it's fair. Thanks for the contribution.

PR #761, opened 2017-05-24
Fixes #760 Issue that causes nav-right to overflow to be hidden on wide-screen

### LINKED ISSUE #760: Nav-right is displayed incorrectly on wide screen
### Overview of the problem

I'm using Bulma **version** [0.4.1]
My **browser** is: Opera 41
I am sure this issue is **not a duplicate**? Yes.

### Description

See this fiddle: https://jsfiddle.net/nngzwtje/2/

When you resize the window in wide-screen to a certain size then the nav-right is cut off. See screenshot attached.

This can be fixed with

```
flex-basis: auto;
```

in https://github.com/jgthms/bulma/blob/master/sass/components/nav.sass#L73

### Steps to Reproduce

Run JSFiddle and resize the bottom part like in screenshot.

### Expected behavior

Resize nav-left.

### Actual behavior

nav-right's overflow is hidden.

kgoedecke:
### Proposed solution

Solves the displaying problem of the nav-right on wide-screen display by setting the flex-basis to auto.

This issue is related to https://github.com/jgthms/bulma/issues/760

### Tradeoffs
No aware of any.

### Testing Done
No testing done.

### DIFF
--- a/sass/components/nav.sass
+++ b/sass/components/nav.sass
@@ -70,7 +70,7 @@ a.nav-item
   max-width: 100%
   overflow: auto
   +widescreen
-    flex-basis: 0
+    flex-basis: auto
 
 .nav-left
   justify-content: flex-start


jgthms:
This will break the `nav-center` unfortunately. I believe the solution is to get rid of `nav-center` completely, or have a modifier.

Issue #782, opened 2017-06-04
Half the height to hero


felipeas:
### Overview of the problem

This is about the Bulma **CSS framework** 

### Description

It would be nice to have:

A half height, just like `is-fullheight`. I wanted a screen that would show nothing else then two `hero`s, and wanted then to have half the height of the screen each.

tldr: `is-halfheight` selector.

jgthms:
Merged. Thanks!

PR #783, opened 2017-06-04
Add is-halfheight to hero

felipeas:
### Proposed solution
I proposed to have an `is-halfheight ` selector to the hero layout, this is just an implementation of what was suggested in #782.

### Testing Done
the hero with `is-halfheight ` has exctaly half the height of the screen, did not aftected `is-fullheight`.

### DIFF
--- a/sass/layout/hero.sass
+++ b/sass/layout/hero.sass
@@ -142,6 +142,14 @@
       .hero-body
         padding-bottom: 18rem
         padding-top: 18rem
+  &.is-halfheight
+    min-height: 50vh
+    .hero-body
+      align-items: center
+      display: flex
+      & > .container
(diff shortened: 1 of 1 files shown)


sass/layout/hero.sass
```diff
@@ -142,6 +142,14 @@
...
         padding-bottom: 18rem
         padding-top: 18rem
+  &.is-halfheight
+    min-height: 50vh
+    .hero-body
+      align-items: center
+      display: flex
+      & > .container
+        flex-grow: 1
+        flex-shrink: 1
```
jgthms: There's a lot of duplication. You can regroup most of `is-halfheight` and `is-fullheight`.

PR #796, opened 2017-06-12
Justified alignment helper class

webernir:
### Proposed solution
When using Right-To-Left and Left-To-Right layout flows you often need to use `text-align: justify`, this PR adds a helper class (`.has-text-justified`) just for that, using Bulmas' naming conventions.

### Tradeoffs
No tradeoffs here.

### Testing Done
Yes


jgthms:
Thanks but this had already been suggested: https://github.com/jgthms/bulma/pull/465

PR #799, opened 2017-06-12
Fixes #781 Add min/max height/width to delete class size modifiers

### LINKED ISSUE #781: Size attributes on the delete element have no effect.
### Overview of the problem

This is about the Bulma **CSS**
I'm using Bulma **version** [0.4.2]

### Description

Size attributes on the delete element have no effect.

### Steps to Reproduce

```html
<a class="delete is-small"></a>
<a class="delete"></a>
<a class="delete is-medium"></a>
<a class="delete is-large"></a>
```

### Actual behavior

image

### Expected behavior

image

I got the expected result if I remove thus attributes from the delete class.

```css
max-height: 20px;
max-width: 20px;
min-height: 20px;
min-width: 20px;
```

ZackWard:
### Proposed solution

This solves #781 by adding min-height and min-width properties to .delete .is-small elements, and by adding max-height and max-width properties to .delete .is-medium and .delete .is-large elements. These properties will override the min/max height/width properties that are set on the .delete class, allowing the delete element to be sized correctly.

### Tradeoffs

Ideally, the min/max width/height properties on the .delete class could simply be removed, and that would allow the width and height properties on the size modifiers to work correctly. However, those properties were added to the .delete element to solve another issue, and cannot be removed. 

### Testing Done

Forked repository
Cloned a local repository
Made changes
Ran `npm run build`
Copied bulma.css to my local dev server
Tested delete element sizes in Chrome 59 and Firefox 53, they seem correct

### DIFF
--- a/sass/utilities/mixins.sass
+++ b/sass/utilities/mixins.sass
@@ -74,12 +74,18 @@
   // Sizes
   &.is-small
     height: 16px
+    min-height: 16px
+    min-width: 16px
     width: 16px
   &.is-medium
     height: 24px
+    max-height: 24px
+    max-width: 24px
     width: 24px
   &.is-large
     height: 32px
+    max-height: 32px
+    max-width: 32px
     width: 32px
 
 =fa($size, $dimensions)
--- a/sass/utilities/mixins.sass
+++ b/sass/utilities/mixins.sass
@@ -74,6 +74,8 @@
   // Sizes
   &.is-small
     height: 16px
+    max-height: 16px
+    max-width: 16px
     min-height: 16px
     min-width: 16px
     width: 16px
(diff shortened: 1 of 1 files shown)


ZackWard:
Here is the result in Chrome:
image

sass/utilities/mixins.sass
```diff
@@ -74,12 +74,18 @@
   // Sizes
   &.is-small
     height: 16px
+    min-height: 16px
+    min-width: 16px
```
jgthms: You need `max-height` and `max-width` too.

sass/utilities/mixins.sass
```diff
@@ -74,12 +74,18 @@
...
     width: 16px
   &.is-medium
     height: 24px
+    max-height: 24px
+    max-width: 24px
```
jgthms: You need `min-height` and `min-width` too.

ZackWard:
@jgthms The requested changes have been made.

jgthms:
Thanks. It's a bit verbose but it will prevent issues.

PR #815, opened 2017-06-21
Fixes #814 - Make use of +fullhd mixin for columns

### LINKED ISSUE #814: Missing -fullhd version of column widths
### Overview of the problem

This is about the Bulma **CSS framework** **and** Bulma **Docs**
I'm using Bulma **version** 0.4.2

### Description

According to the [documentation](http://bulma.io/documentation/overview/responsiveness/) there is a **fullhd** modifier, but there are no .is-X-fullhd classes. Is that intentional? 

### Expected behavior

Just like there are classes for -mobile, -tablet, -desktop and -widescreen - according to this logic - there should be -fullhd equivalents of these, like e.g.: is-6-fullhd, which kicks in from 1384px.

Saboteur777:
Proposed fix for https://github.com/jgthms/bulma/issues/814

### Testing Done
It works (for me. :) ).

### DIFF
--- a/sass/grid/columns.sass
+++ b/sass/grid/columns.sass
@@ -202,6 +202,43 @@
         width: ($i / 12) * 100%
       &.is-offset-#{$i}-widescreen
         margin-left: ($i / 12) * 100%
+  +fullhd
+    &.is-narrow-fullhd
+      flex: none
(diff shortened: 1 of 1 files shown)


jgthms:
Thanks!

PR #830, opened 2017-07-03
Add yarn

samverneck:
### Proposed solution

Another option besides NPM and by the way has a very good performance

### Testing Done

Test done and satisfactory result


jgthms:
Was thinking of doing that. Thanks.

samverneck:
Thank you for accepting PR. How can I join the team of collaborators?
@jgthms

PR #833, opened 2017-07-04
Adding is-small modifier to hero

mccombs:
### Proposed solution
Just adding a simple is-small modifier to the .hero section to create tiny .hero areas, such as an a notice or warning placed at the top of a website.

### Tradeoffs
This modifier could easily become it's own pattern (a fixed position bar, similar to alert but not an alert) instead of just modifying .hero but I think this modifier is worth having as a simple solution.

### Testing Done
Tested on codepen.io : https://codepen.io/mccombs/pen/NRoRKg

### DIFF
--- a/sass/layout/hero.sass
+++ b/sass/layout/hero.sass
@@ -128,6 +128,13 @@
           .nav-item
             border-top-color: rgba($color-invert, 0.2)
   // Sizes
+  &.is-small
+    +tablet
+      .hero-body
+        padding-bottom: 1rem
+        padding-top: 1rem
+      .container 
+        margin: 0 auto;
   &.is-medium
     +tablet
       .hero-body


sass/layout/hero.sass
```diff
@@ -128,6 +128,13 @@
           .nav-item
             border-top-color: rgba($color-invert, 0.2)
   // Sizes
+  &.is-small
+    +tablet
+      .hero-body
+        padding-bottom: 1rem
+        padding-top: 1rem
+      .container 
+        margin: 0 auto;
```
jgthms: Why this change to the container? It doesn't seem specific to the `small` size.

timacdonald: I've done this in my projects, as when the viewport get's to tablet size, the container loses its center alignment. This keeps it center at the smaller breakpoint. I dare say that is why its been implemented in this PR

mccombs: Ideally I would like this to function like Hello bar (see screenshot). 

If you think this should be it's own unique pattern I'm happy to build one, but I thought a modifier to the hero element (like large, full screen etc.) was more appropriate.

hellobar

VizuaaLOG:
I think this should be its own element/component. A hero is generally (in my opinion) a large box to draw you eyes in, for either an image or a large 'Hello this is who we are' box.

This is kind of like a sticky notification. So thinking it could either be a modifier on the notification element or be its own thing entirely. Although not sure what to call it really.

timacdonald:
```
<div class="notification is-sticky"></div>
```

...I love that

PR #850, opened 2017-07-09
[Suggestion] Add column vertical alignment helper

timacdonald:
### Proposed solution
This allows the class `has-vertically-aligned-content` to be added to a `column`, if the column is taller than the contents, the contents will be vertically centered.

### Tradeoffs
I'm not CSS guru, so no tradeoffs I can think of. I've been using it in production happily.

### Testing Done
Yes - [working example](https://codepen.io/timacdonald/pen/vZVavL)

I haven't updated the documentation yet, but happy to add a note about this if you are interested in pulling this into Bulma.

### DIFF
--- a/sass/grid/columns.sass
+++ b/sass/grid/columns.sass
@@ -4,6 +4,7 @@
   flex-grow: 1
   flex-shrink: 1
   padding: 0.75rem
+  // Modifiers
   &.has-vertically-aligned-content
     display: flex
     flex-direction: column
--- a/sass/grid/columns.sass
+++ b/sass/grid/columns.sass
@@ -4,6 +4,10 @@
   flex-grow: 1
   flex-shrink: 1
   padding: 0.75rem
+  &.has-vertically-aligned-content
+    display: flex
+    flex-direction: column
+    justify-content: center
   .columns.is-mobile > &.is-narrow
     flex: none
   .columns.is-mobile > &.is-full


timacdonald:
I should give credit, I got this code from [CSS Tricks](https://css-tricks.com/centering-css-complete-guide/)

timacdonald:
On second thoughts, perhaps if this were to be accepted, it should be `has-content-vertically-aligned` to better match already established naming conventions such as `has-text-centered`

jgthms:
I don't really like the idea of using `flex-direction: column` since it goes against the normal flow. Also, this class looks more like a helper, and is not specific to a column.

I'm not sure what the solution is tbh.

timacdonald:
no worries @jgthms, if your not feeling it, you should probably go with your gut. Bulma is awesome and I wouldn't want to bring the standard down on such a great project.

Perhaps we close this and hope something else comes up. People can always just use this in their own styles if they want something that does it.

jgthms:
Sounds good 👍

fjelsted:
I've been using the following without a problem so far:
```
.column.is-vcentered {
  margin-top: auto;
  margin-bottom: auto;
}
```
Quite a few people have had this problem, so I'm somehow doubting that the solution is that simple... Any reason why I shouldn't be using the above?

joaobarcia:
@fjelsted centering via margins actually adds margins to the div instead of centering the content inside of it. This might raise some issues, i.e. background color being "cropped"

### Using margins:
screenshot_2018-07-26 sindreader 1

---

### Using flex:
screenshot_2018-07-26 sindreader

fjelsted:
@joaobarcia That totally makes sense. Thanks for the clear up. 👍

PR #863, opened 2017-07-13
Fixing using wrong variable for pagination disabled border

SecretBase:
### Changes

Just a small fix for the pagination disabled border.


jgthms:
Thanks!

PR #889, opened 2017-07-20
Fixes #6 | RTL display and a =rtl mixin

### LINKED ISSUE #6: Don't forget about RTL direction please

spacemudd:
### Proposed solution
Problem: In order to display right-to-left content properly, any attribute such as ('left', 'margin-left', 'padding-left') needs to be met with their mirrored version. Not only this, but also to unset the 'left' attributes. Why? Because the attributes will stack up if they are next to each other in sequences and then we will have both 'right' and 'left' styles when viewing the right-to-left content.

Solution: A conditional mixin that adds a right-to-left style to the element while having the flexibility for developers to ignore or have a style that utilizes 'left-' attributes be untouched when the page is mirrored for rtl content.

**sass/utilities/initial-variables.sass** - line 46
`$body-rtl: false`

**sass/utilities/mixins.sass** - line 196-199
```
=rtl
  @if $body-rtl
    @at-root [dir="rtl"] #{&}
      @content
```

**Usage**
```
.content
  margin-left: 30px
  +rtl
    margin-left: unset
    margin-right: 30px
```

### Tradeoffs

In the final css, each class name needs to be written again with [dir="rtl"] pre appended to it. Obviously this will result in a small rise in size. But then again, so are all updates. ;)

I created a boolean `$body-rtl `for when the mixin decides to add the rtl style or not. Obviously for projects that won't need right-to-left support.

Un-minified output: 161KB from 142KB. Is 19KB worth it to add rtl support? In my case, it's very much so.

### Testing Done

Nothing intensive yet. I'm currently developing a large project using Bulma and will be actively updating this fork.


spacemudd:
#1ff0f6f is my last commit.

spacemudd:
@jgthms Do you think this is a viable solution to add to Bulma? Because the way I see it, it is a bit of a workload. What are your thoughts also on having separate repo to for rtl-supported Bulma, does it merit one?

PR #901, opened 2017-07-23
[Suggestion] Make title and subtitle element responsive

BengtHagemeister:
### Proposed solution

I often want to use a smaller title on different screen sizes. With this solution you can add responsiveness to the title and subtitle class.

Example:

`<p class="title is-1-widescreen is-4-mobile">Hello World</p>`

### Tradeoffs

If you use the `<h1-7>` tags and not the `<p>` tag you are not consistent. 

Example: 

`<h1 class="title is-1-widescreen is-4-mobile">Hello World</h1>`

### Testing Done

I used this solution in my own project. It works perfect.

### DIFF
--- a/sass/elements/title.sass
+++ b/sass/elements/title.sass
@@ -36,6 +36,31 @@ $subtitle-weight:   $weight-light !default
     $i: index($sizes, $size)
     &.is-#{$i}
       font-size: $size
+  +mobile
+    @each $size in $sizes
+      $i: index($sizes, $size)
+      &.is-#{$i}-mobile
+        font-size: $size
+  +tablet
+    @each $size in $sizes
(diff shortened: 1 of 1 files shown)


jgthms:
Hey, thanks for your help, but it's something that already exists as a helper: https://github.com/jgthms/bulma/blob/master/sass/base/helpers.sass#L53

I haven't had time to document it yet.

BengtHagemeister:
Ah ok.
But it's not working with the `title` element because it brings his own fontsize. Or am I wrong with that? 

Maybe it will work with an `!important` tag?

jgthms:
Ah yes you're right, I forgot that.

jgthms:
Added!

PR #908, opened 2017-07-24
A11y: Enhance dropdown examples.

Muhnad:
- use button to can reach by keyboard and make a user with screen reader understand that's is a button.
- add aria-haspopup to make screen reader user understand this button have a drop-down.
- add aria-controls to make screen reader user understand this button control on items called x.
- should add aria-expanded by JS when user toggle button.

- add aria-hidden to icons.
- add href=# to links to make it a link we can interact with a keyboard.

- add role menu to menu drop-down to make sure this div is a menu.
- should add aria-hidden by JS and toggle it by button expanded.


jgthms:
Thanks, that's really helpful.

Issue #920, opened 2017-07-26
navbar-item min-width


Wikiki:
Hi,

why not setting a min-width to navbar-item in order to get a square item by default ?

```
.navbar-item,
.navbar-link
  min-width: $navbar-height
  .icon
    width: 100%
```

Regards

Wikiki:
I have created a PR for it: Set a min-width to navbar-item #926

PR #926, opened 2017-07-26
Set a min-width to navbar-item

Wikiki:
PR related to issue: navbar-item min-width #920


jgthms:
I don't think that's mandatory. Some users might want to implement a narrow navbar-item, like 3 vertical dots:

image

PR #928, opened 2017-07-26
Fix #910

### LINKED ISSUE #910: Hamburger icon is not aligned right on using container as first child of navbar
### Overview of the problem

Tried this fix #896 which brought out this issue.  The hamburger menu icon which appears in the mobile view is not aligned to the right when using a container inside a navbar and it looks odd.

Check this https://codepen.io/connectkushal/pen/prodge?editors=1100

### Steps to Reproduce

1. Copy the navbar example code as in docs
2. Wrap everything in the parent nav in a container div
3. See mobile view

Come2Daddy:
Fix #910

### DIFF
--- a/sass/components/navbar.sass
+++ b/sass/components/navbar.sass
@@ -112,6 +112,8 @@ a.navbar-item,
   margin: 0.5rem 0
 
 +touch
+  .container
+    flex-direction: column
   .navbar-brand
     .navbar-item
       align-items: center


sass/components/navbar.sass
```diff
@@ -112,6 +112,8 @@ a.navbar-item,
   margin: 0.5rem 0
 
 +touch
+  .container
```
jgthms: This needs to be `.navbar > .container`

joshmanders:
ping @Come2Daddy

connectkushal:
double ping @Come2Daddy

jgthms:
That's the wrong solution. Should be `display: block`.

Issue #931, opened 2017-07-26
Suggestion: remove overflow from html and margin from columns


nicolashuber:
### Overview of the problem

This is about the Bulma **CSS framework**

I'm using Bulma **version** [0.4.4]
My **browser** is: Google Chrome 59

I am sure this issue is **not a duplicate**

### Description

Currently, in version 0.4.4 the tag `html` and `columns` has:

```css
html {
  overflow-x: hidden;
  overflow-y: scroll;
}
.columns {
  margin-left: -0.75rem;
  margin-right: -0.75rem;
  margin-top: -0.75rem;
}
```
On `html` the property `overflow-x` prevents horizontal scroll, but ends up hiding the content and avoiding the user to see it. The property `overflow-y` always ends up showing a scrollbar. But if I set `overflow:visible`, the default value, the `.columns` create a horizontal scroll because it has a negative margin.

**Why `.columns` need a margin?
Is it similiar to `.row` on Bootstrap?**

```html
<body>
  <div class="columns">
    <div class="column">
      <div class="notification is-warning">Auto</div>
    </div>
    <div class="column">
      <div class="notification is-warning">Auto</div>
    </div>
  </div>
</body>
```
I'm using in the correct way? If yes, the `.column` padding it should appears on body like `.col-xx` on Bootstrap?

![example](https://user-images.githubusercontent.com/7015206/28635192-18776572-7210-11e7-947e-1e6ab55b49dc.jpg)

On Bootstrap the column has a padding of 15px that doesn't appears outside the `body`

I'm creating a suggestion pull-request to adjust this, but first I need to check if I'm not using it wrongly or maybe it's just how Bulma works.

jgthms:
See https://github.com/jgthms/bulma/pull/932

jgthms:
See https://github.com/jgthms/bulma/pull/932

FaezehZamiri:
thanks, it is worked for my issue

PR #932, opened 2017-07-26
Removed html overflow and .columns margin (#931)

nicolasluishuber:
### Proposed solution
This PR remove overflow from `html` and remove negative margins from `.columns`.

### Tradeoffs
Before
before
After
after

### DIFF
--- a/sass/base/generic.sass
+++ b/sass/base/generic.sass
@@ -4,8 +4,6 @@ html
   -moz-osx-font-smoothing: grayscale
   -webkit-font-smoothing: antialiased
   min-width: 300px
-  overflow-x: hidden
-  overflow-y: scroll
   text-rendering: $render-mode
 
 article,
--- a/sass/grid/columns.sass
+++ b/sass/grid/columns.sass
@@ -278,13 +278,6 @@
         margin-left: ($i / 12) * 100%
 
 .columns
-  margin-left: -0.75rem
-  margin-right: -0.75rem
-  margin-top: -0.75rem
-  &:last-child
-    margin-bottom: -0.75rem
-  &:not(:last-child)
-    margin-bottom: 0.75rem
   // Modifiers
   &.is-centered
     justify-content: center


jgthms:
You can't remove the negative column margin. This will cause alignment issues. As for the html overflow, why do you want to remove the horizontal one?

nicolasluishuber:
Look at this first [example](https://jsfiddle.net/aauksy4c/).

Like you can see, I'm trying create a 2 columns layout. But I found some issues:

1. I can't set  `.sideebar` to 100vh, because `.column` has a bottom margin
2. I need add a extra padding to sideebar, because the first `.columns` has invisible padding, hidden outside de body.
3. The browser window shows a scrollbar area.

Now, look this other [example](https://jsfiddle.net/aauksy4c/1/). Now I can use the default .column padding and set the sidebar to 100vh. I don't know, I think your have more control on your layout in this way. And if you don't want the padding, just put `.is-gapless` on .columns, like [this](https://jsfiddle.net/aauksy4c/2/).

Wikiki:
Hi @nicolasluishuber,

why you don't use is-marginless to .columns  in order to remove the bottom margin ?
That's what I do to get a full-height 2 columns layout.

Regards

jgthms:
I can't merge this PR to only solve your specific problem. The negative margin has worked for everyone since the beginning. You can try is-gapless and is-marginless.

atilkan:
@jgthms  I am not happy about learning  is-marginless here. Better docs needed.

VizuaaLOG:
@atilkan if you'd care to look at the [helpers](https://bulma.io/documentation/modifiers/helpers/) section of the docs. You'll find it's described there.

atilkan:
Thanks man.

frederikhors:
I think we should reconsider the idea of removing negative margins and html overflow.

@jgthms can you explain this sentence better?

> You can't remove the negative column margin. This will cause alignment issues.

In a project I need `html` `overflow-x` enabled for example.

I think the @nicolashuber PR is good. Maybe, @nicolashuber we can work together to fix all the possible "alignment issues".

@jgthms Having `overflow-x` "disabled" I think is a minus, it's a restriction. Don't you think?

PR #940, opened 2017-07-28
Fixed hardcoded localhost image url

jkniest:
### Proposed solution
In the current documentation the media object element image on the front page is hardcoded to `localhost:4000` which fails for example on the `bulma.io` page.

I simply use the variable `site.url`

Before:
wrong

After:
bildschirmfoto 2017-07-28 um 12 37 57

Have a nice day,
Jordan


jgthms:
Thanks!

Issue #837, opened 2017-07-06
New Navbar component is not inheriting Hero's color


rafaelpimpa:
### Overview of the problem

 This is about the Bulma **CSS framework** 

 I'm using Bulma **version** [0.4.3] 
 My **browser** is: Chrome/FF

 I am sure this issue is **not a duplicate**

### Description

New Navbar within Hero is not inheriting it's color

### Steps to Reproduce

See: https://codepen.io/anon/pen/jwpKXG

### Expected behavior

That the Navbar would be same color as the Hero, just like Nav.

### Actual behavior

Navbar is always white.

czamp:
It would appear that using the `.is-transparent` helper class with `.navbar` applies a transparent background-color to its direct children, the `.navbar-item` and `.navbar-link`, and their pseudo classes. 

It does not effect the background color of the `.navbar` itself, nor does it apply a transparent background color to any `.navbar-item` classes nested within a `.navbar-dropdown`.

If you want your navbar to match the color of your hero, you can include the following in any custom Sass you load after Bulma:

```
.navbar
  &.is-transparent
    background-color: transparent
```

If you want transparent dropdowns as well, the following should work:

```
.navbar
  &.is-transparent
    background-color: transparent
    .navbar-dropdown
      background-color: transparent
      box-shadow: none
      a.navbar-item
        &:hover
          background-color: transparent
        &.is-active
          background-color: transparent
```

geirmarius:
@czamp There's also the problem on where the text color should inherit and where it can't because of the white dropdown menu, which makes it a bit tricky.

I'd very much love to see support for navmenus inside heros.

gopumon:
If you are using scss version you can customize the navbar color scheme like this.
```
@import './path/to/bulma/sass/utilities/_all';

$navbar-background: $turquoise !default;
$navbar-item: $white !default;
```
and importing bulma main style.
```
@import "./path/to/bulma/bulma";
```

rogerschlachter:
Feels weird that the navbar `background-color` is set to white, but it doesn't have a set `color`.  In my case, it's using the color from .hero.is-primary which in my case was white.

miklb:
> I'd very much love to see support for navbars inside heros.

👍

heliosbayma:
Actually, 

We can achieve it by writing a sass (or scss) file and just rewriting Bulma's own code. This way we'll still be able to use the 'is-transparent' class (an approach that is not far from what gopumon posted).
[...]

PR #945, opened 2017-07-28
Add navbar support to hero

balping:
This PR makes it possible to have a navbar structure inside hero, as it has been possible with nav.

I just basically duplicated rules already present for nav, please double check if this is the way intended!!

I also replaced some `+mobile` to `+touch` where I thought it was necessary.

Again: **please review my changes** before merging. (I'm satisfied with the result though)

Related issue: https://github.com/jgthms/bulma/issues/837

### DIFF
--- a/sass/layout/hero.sass
+++ b/sass/layout/hero.sass
@@ -51,7 +51,8 @@
   display: flex
   flex-direction: column
   justify-content: space-between
-  .nav
+  .nav,
+  .navbar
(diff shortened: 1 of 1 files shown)


suth:
This doesn't look like it will work with new navbar features like dropdowns. Additional rules are needed, which means it might be time to ditch the old nav in hero support so it's not as messy trying to combine the two.

PR #947, opened 2017-07-29
Fixes #913: Safari button text-centering

### LINKED ISSUE #913: Safari .button Text Centering
### Overview of the problem

My **browser** is: Safari
Related: #727 

### Description

The following CSS breaks `<button>` text alignment in Safari _only_.

```
.modal-card-foot .button {
    width: 50%;
}
```

Here is an example Codepen showing this behavior (view in Safari): https://codepen.io/ellisthedev/pen/dzPbbv

### Expected behavior

Both `<button>` and `<a>` should center the text.

### Actual behavior

`<button>` text is left aligned, `<a>` text is centered.

tjkohli:
### Proposed solution

Solves problem #913 

### Tradeoffs

Sets display of button to table-cell. Seems like somewhat of a hack, but it's the only way text on `<button>` can be horizontally centered.

### Testing Done

Tested in Safari, works.

### DIFF
--- a/sass/elements/button.sass
+++ b/sass/elements/button.sass
@@ -33,6 +33,7 @@ $button-shadow-inset: inset 0 1px 2px rgba($black, 0.2) !default
   border-color: $button-border
   color: $button
   cursor: pointer
+  display: table-cell
   justify-content: center
   padding-left: 0.75em
   padding-right: 0.75em


jgthms:
Is this tested in other browsers? Because this will probably break many things.

tjkohli:
Then what should be the fix?

We can target only Safari using something like:

```
 _::-webkit-full-page-media, _:future, :root .safari_only {

   .button {
     display: table-cell;
   }

}
```

PR #961, opened 2017-07-30
add missing default statements to box and image

viqtor:
### Proposed solution
Fixes issue where you could not override variables in box.sass and image.sass

### Testing Done
Override image dimensions and get more classes i.e. `is-640x640`


jgthms:
Thanks!

PR #966, opened 2017-07-30
add "charset: utf-8" to starter template

bazitur:
### Proposed solution
Just added `<meta charset="utf-8">` tag, so browsers display page properly.

### Tradeoffs
No tradeoffs, actually.


jgthms:
Thanks!

PR #968, opened 2017-07-30
Fixes #949, is-size-7-helper not working

### LINKED ISSUE #949: .is-size-7 helper is not present
The docs says you can use the 7 sizes for typography helpers,
but actually, I only get classes from `is-size-1` to `is-size-6`

image

timkley:
### Tradeoffs
None

### Testing Done
yes


jgthms:
Thanks!

PR #977, opened 2017-07-31
Changed ES6 function syntax to ES5 in JS example

pbmarcano:
Simple change to bulma docs.

In Navburger JS example there is a blip of ES6 function syntax. I changed it to ES5 so people who copy and paste it into a vanilla JS file don't run into issues.


jgthms:
Thanks!

PR #982, opened 2017-07-31
This seemed to fix my issue that the burger show up when it wasent ne…

ArneAnka:
…ccessary


joshmanders:
`nav` component is being deprecated.

PR #984, opened 2017-08-01
Fix #833 Adding .is-sticky-top and .is-sticky-bottom modifiers to notifications

alexcanana:
##### Fixes [#833](https://github.com/jgthms/bulma/pull/833)

1. Add modifiers `.is-sticky-top` and `.is-sticky-bottom` to notifications.
2. Creates `=sticky()` mixin for future use in other elements/components.

##### Examples :

```html
<div class="notification is-dark is-marginless is-sticky-top">
  <button class="delete"></button>
  This site uses cookies. By continuing to browse the site you are agreeing to our use of cookies. <a href="#">Find out more here</a>.
</div>
```

```html
<div class="notification is-dark is-marginless is-sticky-bottom">
  <button class="delete"></button>
  This site uses cookies. By continuing to browse the site you are agreeing to our use of cookies. <a href="#">Find out more here</a>.
</div>
```

### Tradeoffs
- `.notification.is-sticky-top` must be the first element in the `<body>`.
I could have used `position: absolute` or `position: fixed`, but this would make the notification appear over the content. Keeping the position relative allows the following elements to appear after the notification.
[View Codepen](https://codepen.io/acanana/pen/pryaZV)

### Testing Done
[Testing Codepen](https://codepen.io/acanana/pen/wqGyrJ)

### DIFF
--- a/sass/elements/notification.sass
+++ b/sass/elements/notification.sass
@@ -35,7 +35,12 @@ $notification-padding: 1.25rem 2.5rem 1.25rem 1.5rem !default
       background-color: $color
       color: $color-invert
   // Modifiers
+  &.is-sticky-top,
+  &.is-sticky-bottom
+    border-radius: 0
   &.is-sticky-top
     +sticky(top, $notification-background-opacity)
+    border-bottom: 1px solid rgba($black, 0.2)
   &.is-sticky-bottom
     +sticky(bottom, $notification-background-opacity)
+    border-top: 1px solid rgba($black, 0.2)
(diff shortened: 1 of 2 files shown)


VizuaaLOG:
The `border-top` and `border-bottom` properties that are applied as part of the sticky mixing, should these not be applied on a per element/component basis? The different coloured notifications, for example, may have a darker version of their main colour for the border?

alexcanana:
I think you're right regarding the border properties, it should be part of the element/component. And, I think the `border-radius` too.

Regarding the notification's border color, it has the same value has the delete button background color. The result is a darker shade of their main color.

VizuaaLOG:
I thought that would be better. Ah okay, that makes sense.

jgthms:
I don't thinks that's the correct solution. Notifications can arrive in bulk, so there should be a container for these notifications rather than positioning only a single one.

I also want to keep notification as a single element.

Feel free to add an extension!

PR #1000, opened 2017-08-02
Update image.html

slaughtr:
Sizes badge set to true

### Proposed solution

Just a minor change to the HTML. Incorrectly said sizes were false. This is obviously not true, as the image container is almost entirely centered around size.

### DIFF
--- a/docs/documentation/elements/image.html
+++ b/docs/documentation/elements/image.html
@@ -39,7 +39,7 @@ variables:
     {%
       include meta.html
       colors=false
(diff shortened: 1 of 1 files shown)


jgthms:
So the reason I didn't add it is because this adds a link to an `#sizes` anchor which doesn't exist on that page. I usually add it for `is-small` `is-medium` `is-large` modifiers.

slaughtr:
Ahhh well don't I feel silly.

PR #1018, opened 2017-08-04
Fixed alignment of .navbar-burger

matteomanf:
### Proposed solution

Removed redundant transform lines and margins (lines 136, 140) while fixing alignment between sides of cross. Fixed alignment in the parent div by making it a size of 16px with a left margin of -8px (lines 117, 122).

### Tradeoffs

None.

### Testing Done

Testing by implementation in multiple browsers (Chrome, Safari...).

### DIFF
--- a/sass/utilities/mixins.sass
+++ b/sass/utilities/mixins.sass
@@ -114,12 +114,12 @@
     display: block
     height: 1px
     left: 50%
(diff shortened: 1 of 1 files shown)


matteomanf:
@jgthms why was this closed, without any explanation?

PR #1021, opened 2017-08-05
doc update for responsive alignment (when deploy 0.5.1)

kennedybaird:
doc update for commit for responsive text alignment

https://github.com/jgthms/bulma/commit/010b6d70a0e4baff1554b7c6406e71fd64881ca1

for when deploying 0.5.1


jgthms:
Thanks but I think a table like the ["Responsive size"](http://bulma.io/documentation/modifiers/typography-helpers/) one would be more useful.

kennedybaird:
I was wondering about that, thanks, I'll offer another commit later :)

PR #1028, opened 2017-08-06
Feature/button.is-rounded

bstashio:
### Proposed solution

Add `is-rounded` modifier to button.

A new `$radius-rounded` variable was also added to unify the value used for making fully rounded border radius.

### Tradeoffs

None that I can think of.

### Testing Done

Testing also includes button group with addons and can be viewed at https://codepen.io/bstashio/full/OjWjLm/.

### Additional Notes

I have set `$radius-rounded` to `290486px` as this was the previous value used for rounded `border-radius`; any particular reason for using this value instead of something like `1.1em`?

### DIFF
--- a/sass/elements/button.sass
+++ b/sass/elements/button.sass
@@ -197,3 +197,5 @@ $button-shadow-inset: inset 0 1px 2px rgba($black, 0.2) !default
     color: $button-static-color
     box-shadow: none
     pointer-events: none
+  &.is-rounded
+    border-radius: $radius-rounded
--- a/sass/elements/form.sass
+++ b/sass/elements/form.sass
@@ -396,12 +396,18 @@ $help-size: $size-small !default
         .select select
           border-bottom-left-radius: $input-radius
           border-top-left-radius: $input-radius
+        .is-rounded
+          border-bottom-left-radius: $radius-rounded
+          border-top-left-radius: $radius-rounded
       &:last-child
         .button,
         .input,
         .select select
           border-bottom-right-radius: $input-radius
           border-top-right-radius: $input-radius
+        .is-rounded
+          border-bottom-right-radius: $radius-rounded
+          border-top-right-radius: $radius-rounded
       .button,
       .input,
       .select select
--- a/sass/elements/form.sass
+++ b/sass/elements/form.sass
@@ -66,7 +66,7 @@ $help-size: $size-small !default
   max-width: 100%
   width: 100%
   &[type="search"]
-    border-radius: 290486px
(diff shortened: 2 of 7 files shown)


jgthms:
I'm not sure if this is needed. The value must be very high to work, so customizing it does not make much sense?

bstashio:
This was used in a design where round buttons were needed, and I thought it would be a good idea to provide it in case others need to apply it in their projects.

Regardless of whether the value is going to be customized or not, I think it's good to have it in one place - BTW, from my tests `1.1em` does the job.

VizuaaLOG:
Can't you add `border-radius: 50%` or 100% and still get the same effect? That really high `border-radius` number just smells imo.

_Sent from my Htc HTC 10 using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

jgthms:
`1.1em` won't always work. `50%` behaves differently.

bstashio:
Any examples on when `1.1em` won't work?

jgthms:
https://codepen.io/jgthms/pen/XapQZL

VizuaaLOG:
@jgthms Ah yeh I see. Was getting confused. Forgot `50%` is useful for creating a circle around an icon or something :)

bstashio:
Thanks for the example; so it basically wouldn't work when `height` is set to a high value, unless we set it the `border-radius` to something much higher.

That aside, I still think having `is-rounded` modifier for buttons is needed.

Wikiki:
I think having a is-rounded modifier for butto s is needed too.
It's very usefull, especially for social buttons

jgthms:
Why not. But it needs its own PR, with adjustments to the horizontal padding.

bstashio:
Why not use this PR? The updated code is only about adding `is-rounded` modifier to `.button`.

jgthms:
You could, but it needs more changes than that.

bstashio:
Ok, let's do it.

What are the changes?

bstashio:
I updated the demo, it now includes fully rounded buttons with social icons.  Have a look https://codepen.io/bstashio/full/OjWjLm.

connectkushal:
@bstashio looks nice. How about turning this into an extension ?

bstashio:
Thanks @connectkushal.

You can if you want, but I think it should be included by default as this is something that has many use cases; after all it's available for the [tag](http://bulma.io/documentation/elements/tag/#modifiers), and I don't see any valid reason why it shouldn't be available for `.button`.

PR #1029, opened 2017-08-06
Feature/layout-backgrounds

bstashio:
### Proposed solution

Add color modifiers to `.footer` and `.section` based on the ones used in `.hero`.

### Tradeoffs

A slightly larger css file.

### Testing Done

Interactive color selection can be viewed at https://codepen.io/bstashio/full/xLgYqd/.


jgthms:
Why not use the hero, which is quite similar to a section?

ctolkien:
Footer however is quite a bit different to hero...?

bstashio:
`.hero` has properties and children properties that may affect how a `.section` and it's children are displayed.  I think both `.footer` and `.hero` should inherit from `.section`.

To avoid having to add colors in 3 files, and the resulting larger css file, we can simply create a new file `./sass/layout/colors.sass` where we can set the colors for all 3 layout classes and import it in `./sass/layout/_all.sass`.  An even better approach, is to have a common file `./sass/layout/common.sass` and include in it all common properties used by the 3 classes, which will include colors too.

bstashio:
Any reason why this isn't merged yet?

jgthms:
* There's no clear separation. What if you only want the hero? You would have unused code.
* And if you only include `hero.sass`, you won't have the colors.
* Also, it references `.nav` which is deprecated.
* `.section` is only for layout purposes, not for styling. This is going to create lots of conflicts and side effects, that `.hero` already has to face.

bstashio:
Thanks for the feedback, Here's my take on it:

> There's no clear separation. What if you only want the hero? You would have unused code.

To the contrary; separation is done between colors and layout, and the only unused code is literally  2 classes `.footer` `.section` and 2 commas.

> And if you only include `hero.sass`, you won't have the colors.

Even better; a user can include colors only when needed, which also comes with an added benefit of allowing a user to create their own color file without the needed of loading additional unused default colors that they'll have to override.

> Also, it references .nav which is deprecated.

I simply used the code that is in `hero.sass` which includes the deprecated `.nav`; I created a new PR https://github.com/jgthms/bulma/pull/1048 to address this issue.

> `.section` is only for layout purposes, not for styling. This is going to create lots of conflicts and side effects, that `.hero` already has to face.

Quite simply, adding the color class is optional, and as with the proposed solution the colors are now in a separate file, which makes it even more efficient as the user would not need to load the colors file until they are needed and that they are not considered to cause conflicts.

PR #1043, opened 2017-08-08
Update typography-helpers.html

joecallon:
Div had been closed accidentally, moving text transformation section out of the container


jgthms:
Thanks but duplicate of #1040

PR #1048, opened 2017-08-09
Replace deprecated .nav with .navbar in .hero layout

bstashio:
### Proposed solution

Replacing deprecated `.nav` with `.navbar` in `.hero`.

### Tradeoffs

None.

### DIFF
--- a/sass/layout/hero.sass
+++ b/sass/layout/hero.sass
@@ -5,7 +5,7 @@
   display: flex
   flex-direction: column
   justify-content: space-between
-  .nav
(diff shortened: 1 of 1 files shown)


suth:
Like #945 this doesn't support new navbar features like dropdowns.

bstashio:
Thanks for pointing it out.

This PR was created in reply to https://github.com/jgthms/bulma/pull/1029#issuecomment-321202499, which gave me the impression that this issue wasn't previously addressed.

PR #1052, opened 2017-08-10
New Extension : Calendar

Wikiki:
Add new Bulma's extension to display a calendar. It can be used on page as large calendar with apointments or in modal/popup for datepicker.


docs/extensions.html
```diff
@@ -3,6 +3,11 @@
 layout: default
 route: extensions
 extensions:
+- name: bulma-calendar
+  url: https://github.com/Wikiki/bulma-calendar
+  slug: bulma-calendar
+  width: 764
+  height: 676
```
jgthms: These values should be divided by 2 for retina.

Wikiki:
Divided image size by 2

ctolkien:
@Wikiki  - these extensions are lit! 🔥 🔥 🔥 !

PR #1058, opened 2017-08-11
Changed the $tablet variable value to 768px.

poomv2:
### Proposed solution
Currently the $tablet variable has a value of 769px. This means that iPad and other tablets that have a portrait resolution of 768x1024 will show mobile styles instead of the expected tablet styles.
Amending this value to 768px fixes this issue.

### Testing Done
Testing complete. No issues. Behaves as expected.


jgthms:
That's the intended behaviour. `$tablet` is for tablet in landscape mode.

nbapp:
Is it possible to add a size between tablet and mobile in that case? Like the I use my iPad also this way often.

jgthms:
It would be possible. But I'm not sure if it's used enough to add another one for the whole framework.

Issue #911, opened 2017-07-24
bulma.styl


Log1x:
### Overview of the problem

This is about the Bulma **CSS framework**

### Description

Hi. I just wanted to make light on my 1:1 fork of Bulma for Stylus. It will keep 1:1 with dev-master and npm. There are similar projects to mine (of which this is even forked from)-- but they are not being maintained to keep parity with the official repository.

I'm not sure your take on Stylus, but feel free to add it to the related projects if you'd like.  :)

[Bulma.styl](https://github.com/log1x/bulma.styl)

It will be updated daily with any changes done to the master Bulma repo.

-Brandon

jgthms:
Hey thanks. Could you add it to the related projects?

PR #1067, opened 2017-08-12
Add Bulma.styl to related projects & make related project markup more consistent (#911)

Log1x:
### Proposed solution

* Add Bulma.styl to related projects.
* Fix related project markup to be more consistent  (use — for separator instead of a mix between `—`, `-`, and `:`)

### Tradeoffs

N/A

### Testing Done

N/A


Log1x:
Check #1068 -- if you like the table then just remove this PR.

PR #1068, opened 2017-08-12
Move related projects to a table

Log1x:
### Proposed solution

Move related projects to a table.

### Tradeoffs

Markdown tables aren't the prettiest thing syntax-wise, but it looks a lot nicer.

### Testing Done

N/A


jgthms:
Very nice, thanks.

PR #1071, opened 2017-08-13
chmod 0644 *.sass

danielbayerlein:
The `*.sass` files have incorrect rights (`0755`). This PR change the rights to `0644`.


balping:
Just to spell it out (might help for those, who are not fluent in base 8 and/or umask)

**Before**

* owner: read, write, execute
* group: read, execute
* other: read, execute

**After**

* owner: read, write
* group: read
* other: read

So this PR basically removes the executable flag, which is indeed a good idea.

balping:
@jgthms Please apply the same for this file too: https://github.com/jgthms/bulma/blob/master/sass/elements/container.sass

(It was added later, than this PR had been issued)

PR #1079, opened 2017-08-16
remove min-width in the html element.

chistel:
remove min-width of 300px in the html element, as it causes the page to break slightly at the left on mobile


jgthms:
How did you test this?

chistel:
not much, been using bulma for some months, a client complained about the break on mobile, I checked using some mobile phones and using the responsive settings of Mozilla.

jgthms:
Do you have an example or screenshot?

PR #1081, opened 2017-08-16
Fixes #1080 - Add responsiveness to table

### LINKED ISSUE #1080: Responsive table
Is it about Bulma or about the Docs? **Bulma**
Is it a bug/feature/question or do you need help? **feature**
If it's a bug, is it a browser bug? **N/A**

### Overview of the problem

I'm using Bulma **0.5.1**
My **browser** is: Google Chrome
This is a **Sass** issue: I'm using version [3.5.1]

### Description

Make a table horizontally scrollable on smaller screen widths.

Related issues: #939.

brunodplima:
Make a table horizontally scrollable on smaller screen widths.

### Proposed solution
Make a table horizontally scrollable on smaller screen widths.

### Tradeoffs
N/A

### Testing Done
There is a demo on: https://codepen.io/brunodplima/full/rzpaLO/

Closes #1080

### DIFF
--- a/sass/elements/table.sass
+++ b/sass/elements/table.sass
@@ -96,3 +96,82 @@ $table-striped-row-even-hover-background-color: $white-ter !default
           background-color: $table-striped-row-even-background-color
           &:hover
             background-color: $table-striped-row-even-hover-background-color
+
+  +until($desktop + $container-offset)
+    &.is-responsive
+      width: 100%
+      border-collapse: collapse
+      border-spacing: 0
+      display: block
+      position: relative
+
+      td:empty:before
+        content: '\00a0'
+
+      th,
+      td
+        margin: 0
+        vertical-align: top
+
+      th
+        text-align: left
+
+      thead
+        border-right: solid 2px $grey-lighter
+        display: block
+        float: left
+
+        tr
+          display: block
+          padding: 0 10px 0 0
+
+          th::before
+            content: "\00a0"
+        td,
+        th
+          border-width: 0 0 1px
+
+      tbody
+        display: block
+        width: auto
+        position: relative
+        overflow-x: auto
+        white-space: nowrap
+
+        tr
+          display: inline-block
+          vertical-align: top
+
+      th
+        display: block
+        text-align: right
+
+      td
+        display: block
+        min-height: 1.25em
+        text-align: left
+
+      th,
+      td
+        &:last-child
+          border-bottom-width: 0
+
+      tr
+        &:last-child
+          td:not(:last-child)
+            border: $table-cell-border
+            border-width: $table-cell-border-width
+
+      &.is-bordered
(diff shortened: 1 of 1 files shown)


carlosxjm:
Look good!

jgthms:
This is the wrong solution for several reasons:
* it's not responsive since it behaves similarly on mobile and desktop
* it requires a complete different markup
* it requires to duplicate the table
* the responsive table uses a single row (?!)
* it uses cells and rows with `display: block`, so why use table tags in the first place?

Responsiveness is about adapting the content to the viewport, while keeping the _same markup_. A scrollable `table-container` element is the solution here.

brunodplima:
Thanks for the response.
Nevertheless, I think you sticked too much to the "responsive" term. I mean, I just want each record be quickly visible and understandable.

> it's not responsive since it behaves similarly on mobile and desktop

**Wrong**. It's completely different.

> it requires a complete different markup

**Wrong**. Just one `<table class="table is-responsive"></table>` and you are done.

> it requires to duplicate the table

**Wrong**. Why do you think so? What would be the point?

> the responsive table uses a single row (?!)

**Wrong**. Why do you think so? What would be the point?

> it uses cells and rows with display: block, so why use table tags in the first place?

Table is to show tabular data. It's the semantic way to show that information. But it's hard to fit on a vertical screen. 

> Responsiveness is about adapting the content to the viewport, while keeping the same markup. A scrollable table-container element is the solution here.

**Correct**. That's why only one markup is needed. A scrollable table-container element is a poor solution because, the user has to scroll almost to infinity if the table is to wide. Additionally he wouldn't be able to quickly compare entries.


I think you misunderstood my demo. **So I updated it**. Feel free to resize your browser and/or open on a mobile device. (The former code was made that way in order to avoid the resizing).

Bulma is great. It's an excellent open source solution. I'm just trying to improve it. Please, consider this.

Best regards

gpip:
Is bulma considering a different approach for responsive tables?

nickwest:
Personally I'd like to see multiple options in Bulma for responsive tables. Different approaches are appropriate for different sets of data. The three examples given here are common approaches: https://elvery.net/demo/responsive-tables/

1) Having the ability to drop columns at different responsive steps (the solution in the PR)
2) Flipping scroll so the table is horizontal instead of vertical
3) Repeating the labels and display the table more like a list

All of these can be done without JS and without duplicating markup.

tony:
The pull request is better than nothing.

jgthms:
> The pull request is better than nothing.

That is the wrong attitude to have. Otherwise, Bulma would have "okay-ish" solutions for almost everything, increasing the size of the core, introducing lots of potential bugs, and promoting bad design.

The best solution so far is to add a `responsive-table` wrapper that scrolls the table.

PR #1090, opened 2017-08-18
Remove gap between tag and delete button

balping:
This PR removes the 1px wide gap between tag and delete tag.

| Before | After |
|:----------:|:--------:|
| before   |   after    |

It seemed to me that that gap was intentional (since it was explicitly added), but it feels somewhat unnatural to me. Plus there's no gap between plain attached tags.

### DIFF
--- a/docs/documentation/elements/tag.html
+++ b/docs/documentation/elements/tag.html
@@ -10,6 +10,8 @@ variables:
   value: $text
 - name: $tag-radius
   value: $radius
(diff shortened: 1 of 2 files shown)


jgthms:
I'd rather make it a variable then.

balping:
OK, I'll update the PR then.

balping:
@jgthms Updated. Docs updated too.

jgthms:
Thank you that's great.

PR #1098, opened 2017-08-21
moving icon to bottom to accomodate label.

hammygoonan:
### Proposed solution

If a text input has a label above it the left and right icons sit outside the input.

screen shot 2017-08-21 at 21 22 04

Align icon to bottom of container rather than the top.


jgthms:
The label should not be in the `.control` element.

PR #1118, opened 2017-08-26
Update bulma-steps extension preview image

Wikiki:
Just update bulma-steps extension preview image


jgthms:
Dimensions need to be updated.

PR #1125, opened 2017-08-28
Wrap labels if inside a field with addons. Fixes #777

### LINKED ISSUE #777: Addons field and label
### Overview of the problem

I'm using Bulma **version** [0.4.2]
My **browser** is: Firefox/Chrome
I am sure this issue is **not a duplicate** (at least I couldn't find)

### Description

When using label in a has-addons field, the label position is not set correctly.

### Steps to Reproduce

Create a field like this
```html
    <label class="label">Label</label>
    <p class="control">
        <input class="input" type="text" placeholder="Find a repository">
    </p>
    <p class="control">
        <a class="button is-info">
            Search
        </a>
    </p>
</div>
```

A **workaround** for this is wrapping the ``.has-addons`` on a ``.field`` with the label:

```html
<div class="field">
    <label class="label">Label</label>
        <p class="control">
            <input class="input" placeholder="Find a repository" type="text">
        </p>
        <p class="control">
            <a class="button is-info">
                Search
            </a>
        </p>
    </div>
</div>
```

### Expected behavior

expected

### Actual behavior

current

havgry:
Currently less elegant markup workarounds are needed when combining field addons and labels.

### Proposed solution
Set labels to full width and wrap if inside a field with the `.has-addons` modifier. Fixes #777

### Tradeoffs
I'd argue it retains the structure of the affected classes and adds very little complexity while removing the need for ugly workarounds.

### Testing Done
I've tested all relevant modifiers (`.is-expanded`, `.is-fullwidth`, `.has-addons-centered`, `.has-addons-right`) and the issue doesn't affect horizontal forms because they're divided into `.field-label` and `.field-body`.

### DIFF
--- a/sass/elements/form.sass
+++ b/sass/elements/form.sass
@@ -365,10 +365,10 @@ $help-size: $size-small !default
   display: block
   font-size: $size-normal
   font-weight: $label-weight
-  &:not(:last-child)
-    margin-bottom: 0.5em
   .has-addons &
     flex: 0 0 100%
+  &:not(:last-child)
+    margin-bottom: 0.5em
   // Sizes
   &.is-small
     font-size: $size-small
@@ -394,8 +394,8 @@ $help-size: $size-small !default
   // Modifiers
   &.has-addons
     display: flex
-    justify-content: flex-start
     flex-wrap: wrap
+    justify-content: flex-start
     .control
       &:not(:last-child)
         margin-right: -1px
--- a/sass/elements/form.sass
+++ b/sass/elements/form.sass
@@ -367,6 +367,8 @@ $help-size: $size-small !default
   font-weight: $label-weight
   &:not(:last-child)
     margin-bottom: 0.5em
+  .has-addons &
+    flex: 0 0 100%
   // Sizes
   &.is-small
     font-size: $size-small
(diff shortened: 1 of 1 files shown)


sass/elements/form.sass
```diff
@@ -392,6 +394,7 @@ $help-size: $size-small !default
   // Modifiers
   &.has-addons
     display: flex
+    flex-wrap: wrap
```
jgthms: The addons should never wrap.

sass/elements/form.sass
```diff
@@ -365,6 +365,8 @@ $help-size: $size-small !default
   display: block
   font-size: $size-normal
   font-weight: $label-weight
+  .has-addons &
+    flex: 0 0 100%
```
jgthms: Not a big fan of shorthand properties. A `width: 100%` would have sufficed.

havgry:
After having a second look I kind of agree with you. This was probably a bit rushed. 

> The label should be outside the field.

Looking at the field and label class, it makes sense as they're completely decoupled. However, everywhere in the documentation labels are placed within fields which is causing confusion when they're suddenly not working as expected with `.has-addons`.

chasegiunta:
@havgry @jgthms  Just ran into this as well. Building out a component and it's a little funky that labels need to be placed _elsewhere_ if they include an addon. Not a deal breaker by any means, but potentially helpful & time-saving if included in the docs since, as @havgry pointed out, every instance within the docs has `label`s **inside** fields.

PR #1126, opened 2017-08-28
Update customize.html

kikoseijo:
$family-primary its not used at all, its only setup in variables.sass and derived-variables.sass but not implemented after this, so with previous var name didn't reflected changes.

### DIFF
--- a/docs/documentation/overview/customize.html
+++ b/docs/documentation/overview/customize.html
@@ -25,7 +25,7 @@ $primary-invert: $pink-invert
 // Use the existing orange as the danger color
 $danger: $orange
 // Use the new serif family
-$family-primary: $family-serif
+$family-sans-serif: $family-serif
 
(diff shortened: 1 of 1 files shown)


docs/documentation/overview/customize.html
```diff
@@ -25,7 +25,7 @@
 // Use the existing orange as the danger color
 $danger: $orange
 // Use the new serif family
-$family-primary: $family-serif
+$family-sans-serif: $family-serif
```
jgthms: This doesn't make sense?

IsmaelDev:
.

docs/documentation/overview/customize.html
kikoseijo: Sorry, didn't see, problem its, if I declare as you have it will not propagate over the compilation.
If you search for $family-primary inside yours, does not get used.

Sorry for late reply.

PR #1129, opened 2017-08-30
[Documentation] Include example of widescreen and fullhd modifiers

KKSzymanowski:
The `widescreen` and `fullhd` modifiers are explained on the [Overview | Responsiveness](http://bulma.io/documentation/overview/responsiveness/) documentation page but there is no mention of them in [Columns | Responsiveness](http://bulma.io/documentation/columns/responsiveness/) which can be misleading.

I extended the responsiveness example by these two modifiers.

The last two columns were removed because in mobile, with the first column taking up 75% of the space, they were overflowing to the right.


jgthms:
Thanks!

Issue #1084, opened 2017-08-17
Update docs for Hero to use navbar


tony:
http://bulma.io/documentation/layout/hero/

According to http://bulma.io/documentation/components/navbar/ `nav` is deprecated.

PR #1140, opened 2017-09-01
update nav-* classes to navbar-* classes

phatchai:
change .nav-left to .navbar-start
change .nav-right to .navbar-end


tony:
Related: #1084

PR #1169, opened 2017-09-09
Fix undefined variable navbar-item in sass/components/navbar.sass. Fixes #1168

### LINKED ISSUE #1168: Bug Undefined variable: "$navbar-item" in sass/components/navbar.sass
### Overview of the problem

I'm using Bulma **version** [0.5.1]

### Description

When compiling using `sass bulma.sass bulma.css` we get the following error:

```
Error: Undefined variable: "$navbar-item".
        on line 61 of sass/components/navbar.sass, in `@content'
        from line 209 of sass/utilities/mixins.sass, in `desktop'
        from line 55 of sass/components/navbar.sass
        from line 12 of sass/components/_all.sass
        from line 6 of bulma.sass
```

### Solution

It's obvious that the variable `$navbar-item` is not defined and was mistakenly used to set the `color` property in `navbar.sass`, I think it should be changed to `$color` instead.  The code on line 61  of sass/components/navbar.sass would become as follows:

```
color: $color
```

bstashio:
### Proposed solution

Fixes issue #1168 

### Tradeoffs

None

### Testing Done

`sass bulma.sass bulma.css`

### DIFF
--- a/sass/components/navbar.sass
+++ b/sass/components/navbar.sass
@@ -58,7 +58,7 @@ $navbar-divider-background-color: $border !default
             border-color: $color-invert
         .navbar-dropdown a.navbar-item:not(.is-active):not(:hover),
         .navbar-dropdown .navbar-item:not(a)
-          color: $navbar-item
+          color: $color
         .navbar-item.has-dropdown:hover .navbar-link,
         .navbar-item.has-dropdown.is-active .navbar-link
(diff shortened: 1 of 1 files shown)


jgthms:
There's more issues with that navbar color PR. I've already fixed them. Just need to push.

bstashio:
cool

PR #1173, opened 2017-09-11
Add $icon-selector variable

Log1x:
### Proposed solution

I noticed that `.fa` is hardcoded in multiple places and thus makes using icon fonts other than Font Awesome a little frusterating. To combat this, I added a variable `$icon-selector` which allows us to change the `.fa` selector using interpolation.

### Tradeoffs

No tradeoffs whatsoever.

### Testing Done

 I have compiled and tested with everything working as expected.

### DIFF
--- a/docs/documentation/overview/variables.html
+++ b/docs/documentation/overview/variables.html
@@ -91,6 +91,8 @@ initial_variables:
   value: 5px
 - name: $speed
   value: 86ms
+- name: $icon-selector
+  value: fa
 derived_variables:
 - name: $primary
   value: $turquoise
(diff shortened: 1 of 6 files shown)


jgthms:
This has been deprecated since there's no selector anymore.

Log1x:
Fair enough.

Do you plan on removing `.fa` everywhere else it's used as well? It's used multiple times in `form.sass` for general forms as well as `file-icons`. It's also still in `panel.sass`.

jgthms:
Good shout. Yes they need to be replaced with `.icon`.

PR #1187, opened 2017-09-14
:memo: Redact note about local environment setup to include env key

peterblazejewicz:
This PR intends to improve development/contributors tooling setup.

The change allows to use `development` checks already used in the source - for example
exclude 3rd party scripts fetching on local machine removing errors from console.

Thanks!

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -9,7 +9,7 @@ The documentation HTML is produced with the Ruby-based `jekyll` tool.
 
 Then to view the documentation in your local checkout:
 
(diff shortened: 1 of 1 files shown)


jgthms:
The 3rd party icons are not in the repo, so this wouldn't work unfortunately.

peterblazejewicz:
> The 3rd party icons are not in the repo, so this wouldn't work unfortunately.

Not sure, I've made this exactly for the purpose after running locally, but OK
https://github.com/jgthms/bulma/blob/1e4ae9e3e9d2f7299445de08f23448dee4e42e2d/docs/_includes/bsa.html#L1

PR #1153, opened 2017-09-04
remove unwanted margin-right on .level-item at mobile breakpoint [Fixes #1152]

chriszamierowski:
### Proposed solution
This fixes [#1152](https://github.com/jgthms/bulma/issues/1152) by moving the `margin-right` to the tablet breakpoint

This fixes the alignment on mobile:

fixed_mobile

It also preserves the behavior at the tablet breakpoint:

fixed_desktop

PR #1193, opened 2017-09-15
Update level.sass to fix issue with icons

connectkushal:
Removing `margin-right: 0.75rem` from the `level is-mobile` in #1153 changed the spacing of icons used in the `level-item`. [See the docs for the example code](http://bulma.io/documentation/elements/box/) of `box`, in mobile view, the icons get squashed together. 

This pr restores the margin-right **for the icon class only** to restore the behavior as before while also keeping changes made in #1153.

### Proposed solution
added margin-right for icon class only

### Tradeoffs
none

### Testing Done
yes

### DIFF
--- a/sass/components/level.sass
+++ b/sass/components/level.sass
@@ -41,8 +41,6 @@
   +mobile
     &:not(:last-child)
       margin-bottom: 0.75rem
-    // margin for icons
-    & > .icon
       margin-right: 0.75rem
     
 .level-left,
--- a/sass/components/level.sass
+++ b/sass/components/level.sass
@@ -20,6 +20,7 @@
         margin-bottom: 0
       &:not(.is-narrow)
         flex-grow: 1
+      margin-right: 0.75rem
   // Responsiveness
   +tablet
     display: flex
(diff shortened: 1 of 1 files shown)


connectkushal:
fix#1153 

(checking if this would trigger travis ci)

jgthms:
That's not the correct solution. The margin should be on `level-item`.

connectkushal:
I hope i understood you correctly. I added it to `level-item` as you suggested and found that it adds extra margin to the icons in desktop and tab views. Here's the screenshot of the code and output:

On adding to level-item
under level-item

connectkushal:
did you mean this ?  

```
.level-item
  +mobile
    at-root &
       margin-right: 0.75rem 
```

connectkushal:
@jgthms done

jgthms:
No but it should be only when there's `.level.is-mobile`.

connectkushal:
@jgthms i made the changes and checked the css output. Getting margin on the `.level.is-mobile .level-item` 

Output:
```

.level.is-mobile .level-item {
  margin-right: 0.75rem;
}

```

PR #1203, opened 2017-09-17
Fix #935 Fix #1216

### LINKED ISSUE #935: Bug dropdown in hero (primary) menu items not visible
My **browser** is: Chrome 59

### Description

When a dropdown button is placed inside a hero container are the colors of the dropdown items being inherited.

### Steps to Reproduce

Place a dropdown button in a hero container that is primary (is-primary).

### Expected behavior

An exception should be made for the menu items in the button to not override their color.
`.hero.is-primary a:not(.button), .hero.is-primary strong {
    color: inherit;
}`

### Actual behavior

Menu dropdown items color are being inherited.

### LINKED ISSUE #1216: <a class="tag" href="..."> has white text on white bg, when in primary hero
### Overview of the problem

I'm using Bulma **version** 0.5.3
My **browser** is: Firefox Nightly 57.0a1 (2017-09-20)
I am sure this issue is **not a duplicate**? Yes, only #770 is similar but not quite.

### Steps to Reproduce

Create a page with Bulma and the following code (a minimal testcase):
```html
<section class="hero is-primary">
	<a class="tag" href="https://github.com/">Hello!</a>
</section>
```

### Expected behavior

A tag containing "Hello!" in a small hero should be displayed.

### Actual behavior

The tag is displayed indeed, but as white text on a white background.

Note that this also happens in more complex examples, as long as the hero is primary. This is also fixed by using `<span class="tag"><a>` instead.

connectkushal:
Added not(.dropdown) to fix #935

### Testing Done
yes, locally using laravel mix.

### DIFF
--- a/sass/layout/hero.sass
+++ b/sass/layout/hero.sass
@@ -17,7 +17,7 @@
     &.is-#{$name}
       background-color: $color
       color: $color-invert
-      a:not(.button),
+      a:not(.button):not(.dropdown-item):not(.tag),
       strong
         color: inherit
       .title


jgthms:
It's `:not`

connectkushal:
:not was failing on my end so added &:not.

jgthms:
That's not correct. It's going to output `navbar.is-primary:not(.dropdown)`.

connectkushal:
I am using laravel mix to compile sass. On searching the css output for not:(.dropdown), it shows 9 results. All are with .hero.is-* only, like
 `.hero.is-primary:not:(.dropdown)`

Have just checked it again and its the same.

connectkushal:
screenshot from 2017-09-18 08-15-50

connectkushal:
@jgthms any changes required on this pr ?

jgthms:
`.hero.is-primary:not(.dropdown` is not the desired output. It targets `<section class="hero is-primary">` but not `<section class="hero is-primary dropdown">`. But the latter never happens. 

Here's what we want:
```
<section class="hero is-primary">
  <a>Target this</a>
  <a class="dropdown">Not this</a>
</section>
```

connectkushal:
I understand now. So then i added `a:not(.dropdown-*)` which includes every class starting with `dropdown-` totally isolating it from the color of hero. Maybe there can be a feature for dropdown colors in future.

image

connectkushal:
@jgthms  I got your point, thanx for explaining it. The only part that should not inherit hero color is `dropdown-item` class for `a` tags. The last commit targets only that.

connectkushal:
Css output:
 
image

connectkushal:
edited:

@jgthms  I think I got your point, thanx for explaining it. The `dropdown` class inside hero should not inherit hero color.

adding `.dropdown` to `a:not(.button, .dropdown)` solves the issue and the css output shows 

```
.hero.is-black a:not(.button, .dropdown),
.hero.is-black strong {
  color: inherit;
}
```

connectkushal:
@jgthms hello. Please let me know if there are any more changes to be made to get this pr accepted, as i really need this issue fixed in the master branch :grimacing:

connectkushal:
@jgthms any changes required in this pr ?

sass/layout/hero.sass
```diff
@@ -18,7 +18,7 @@
     &.is-#{$name}
       background-color: $color
       color: $color-invert
-      a:not(.button),
+      a:not(.button, .dropdown),
```
jgthms: Wrong syntax. Should be `a:not(.button):not(.dropdown)`.

connectkushal:
@jgthms done !

connectkushal:
@jgthms I just tested this suggestion after i had committed the change and there seems to be a problem.

This, `a:not(.button):not(.dropdown)` is not compiling to css at all. The color of links inside hero body gets messed up nor do i find this in the output css.

I tried adding `,` to `a:not(.button):not(.dropdown),`. It shows this line in the css output, normal link works fine but issue of dropdown-menu background inside hero is not fixed, its just as before.

SimplyThomas:
@connectkushal it looks like you accidentally removed the trailing `,` after `a:not(.button):not(.dropdown)`

It should be `a:not(.button):not(.dropdown),`


Though, using this syntax instead of `a:not(.button, .dropdown),` does not fix issue #935 for me on Chrome 61. 

@jgthms The [W3C spec](https://www.w3.org/TR/selectors4/#negation) for :not() identifies that the negation is supposed to take in a selector list as the argument, so the more correct syntax should be `a:not(.button, .dropdown),` as was identified before.

connectkushal:
@SimplyThomas I did try adding the trailing `,` ! But had edited my comment as i was not sure if it was the correct syntax because even that did not fix the issue (though it was compiling and i could see it in the changes in the css)

Thanx for pointing out the W3C spec, `a:not(.button, .dropdown)` was working in previous commits, so will revert to that again.

connectkushal:
This fix was working well in version 5.4 but came across new issue on **version 6.0**. 

On some more digging up on the :not, i [found some things](https://developer.mozilla.org/en-US/docs/Web/CSS/%3Anot) on the developer.mozilla.org which i want to point out first

It says this `a:not(div):not(span)` is used for multiple elements with :not, where as for class list it suggests this syntax `a :not(.class1, .class2)` but **also there is a note in the comment that this feature is not well supported yet**.

Now the new issue that i faced was, that the link color inside the `hero is-info` should have been white as it was before, but after the fix it turned to the new link color, blue. 

Essentially i realized that doing `a:not(.button, .dropdown)` or adding anything after .button to the list completely ignores all the styling of `a` inside the hero even those inside other elements like tags. 
I tried removing the `.button` and making it `a:not(.dropdown)` even `a:not(.dropdown-item)` or `-menu` or `-content`, none of which worked at all. All these things happen when adding `.tag` class as well. 

**Something that did work** without breaking anything was adding `!important` to the dropdown-item color and same for tags. 

@jgthms @SimplyThomas Im still learning so would appreciate your thoughts on how to fix this properly.

SimplyThomas:
@connectkushal Good find. I didn't realize that I was looking at a draft specification.

For what it's worth, since you're still learning, you should do everything you can to avoid using `!important` in a framework. This will override any additional styles for end-users and should really only be used on a case-by-case basis in an individual page's CSS -- not even site-wide.

I was able to get the intended results using the following syntax:
`a:not(.button):not(.dropdown):not(.dropdown-item):not(.tag),`


I've made a [codepen](https://codepen.io/SimplyThomas/pen/gGqwXV) that demonstrates that this selects the intended <a> and <strong> elements, but not the other classes involved.

sass/layout/hero.sass
```diff
@@ -17,7 +17,7 @@
     &.is-#{$name}
       background-color: $color
       color: $color-invert
-      a:not(.button),
+      a:not(.button, .dropdown, .tag),
```
SimplyThomas: changing this to `a:not(.button):not(.dropdown):not(.dropdown-item):not(.tag),` produces the desired results.

connectkushal:
@SimplyThomas Thanx for the insight on `!important`. 

And yes it worked ! I think there is no need for `.dropdown` in the list, so omitted that from the commit

@jgthms made the changes you suggested and also added `.tag`

connectkushal:
@jgthms please merge this.

SimplyThomas:
@jgthms The latest commit looks good to me and has a proof-of-concept available on [codepen](https://codepen.io/SimplyThomas/pen/gGqwXV)

PR #1209, opened 2017-09-19
Remove binding between `$primary-invert` and `$turquoise-invert`

matteomanf:
### Proposed solution

In the derived-variables.sass `$primary-invert` is assigned the default primary inverted color, while if you change the primary it should change, it actually doesn't because it is hardcoded to equal turquoise. Modified so that `$primary` changes affect `$primary-invert`.

### Tradeoffs

None

### Testing Done

Works.

### DIFF
--- a/sass/utilities/derived-variables.sass
+++ b/sass/utilities/derived-variables.sass
@@ -18,7 +18,7 @@ $blue-invert: findColorInvert($blue) !default
 $purple-invert: findColorInvert($purple) !default
 $red-invert: findColorInvert($red) !default
 
-$primary-invert: $turquoise-invert !default
+$primary-invert: findColorInvert($primary) !default
 $info-invert: $blue-invert !default
 $success-invert: $green-invert !default
 $warning-invert: $yellow-invert !default


sass/utilities/derived-variables.sass
```diff
@@ -18,7 +18,7 @@ $blue-invert: findColorInvert($blue) !default
 $purple-invert: findColorInvert($purple) !default
 $red-invert: findColorInvert($red) !default
 
-$primary-invert: $turquoise-invert !default
+$primary-invert: findColorInvert($primary) !default
```
Florian-R: If this gets to be merged, it would be more consistent to do the same for all the variables below.

matteomanf: I will modify afterwards, on my mobile now! You are right...

jgthms:
I like the idea. You just need to remove all unused variables from the `.sass` files and update the docs too.

matteomanf:
@jgthms I have modified the docs, resolved the merge conflicts and done the same treatment for the `$link-invert` variable, binding the color the value of `$link`, instead of the default `$blue-invert`.

PR #1228, opened 2017-09-25
Programmatically set bulma's sass files count in docs

bstashio:
This is related to the [Modular Page](http://bulma.io/documentation/overview/modular/) in the Docs.

### Proposed solution

Currently, the sass files count is set manually to `29` files, which is an outdated count.  This solution aims at avoiding the need of manually setting this count.

This was achieved by creating a new [Jekyll plugin](http://jekyllrb.com/docs/plugins/), namely `Jekyll::Bulma::SassFilesCount`, which takes care of counting the imported sass files used in bulma.

### Tradeoffs

None that I can think of.

### Testing Done

Tested in local development environment using:
```bash
cd docs && \
jekyll serve --config _config.local.yml
```

### Additional Notes

If this is considered useful, then we can consider cooking up other Jekyll plugins to dynamically set values such as the version number and other related config settings.

### DIFF
--- a/docs/_plugins/sass_files_count.rb
+++ b/docs/_plugins/sass_files_count.rb
@@ -1,0 +1,61 @@
+module Jekyll
+  module Bulma
+    # A Jekyll Tag type that renders the count of Bulma Sass files.
+    # 
+    # It works by reading Bulma's primary Sass file `bulma.sass`
+    # from the `git` repository.
+    #
+		class SassFilesCount < Liquid::Tag
+			def render(context)
+			
+				# bulma primary sass file
+				file = '../bulma1.sass'
+				
+				if !File.readable?(file)
+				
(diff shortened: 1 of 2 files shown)


jgthms:
I believe Github Pages does not allow plugins to run?

bstashio:
You're right.

I've never used GitHub pages before and I wrongly assumed it would be supported.  Apparently, and as per their [documentation](https://help.github.com/articles/adding-jekyll-plugins-to-a-github-pages-site/), the only way for this to work is to generate the site locally and push the static files, which isn't ideal nor practical, especially as there are many users contributing to the docs.

I was hoping that we could save ourselves the headache of having to manually update the docs with such details, similar to how I'm doing it in [bulma customizer](https://bulma-customizer.bstash.io/), but unfortunately that doesn't seem it would work without creating another headache.

PR #1230, opened 2017-09-26
Adding ";" on @require statements to avoid  errors

JuanDMeGon:
If not using the semicolon at the end of the @import lines this might cause syntax errors like "Media query expression must begin with '('"

### DIFF
--- a/docs/documentation/overview/customize.html
+++ b/docs/documentation/overview/customize.html
@@ -7,7 +7,7 @@ doc-subtab: customize
 
 {% capture sass %}
 // 1. Import the initial variables
-@import "../sass/utilities/initial-variables"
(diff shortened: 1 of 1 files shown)


jgthms:
For `.scss` maybe but not in `.sass`

JuanDMeGon:
For sure, this might be problematic for .scss files, but on .sass files you can add the semicolon too without problem :)

bstashio:
sass would throw the following error if you add a semicolon in `.sass`:

```
Error: Invalid @import: expected end of line, was ";".
```

JuanDMeGon:
Ohh I get it, didn't realize that. Thanks :)
Closing

PR #1238, opened 2017-09-27
Update references to variables in docs

bstashio:
### Proposed solution

Update outdated references to variables in addition to adding missing ones such as the following:

- section variables
- footer variables
- form generic variables - `$control-*`, `$label-*`, `$help-size`
- generic variables from `./sass/base/generic.sass`

### Tradeoffs

Can't think of any

### Testing Done

Done locally using `jekyll serve --config _config.local.yml`

### Additional Notes

This process is done manually which is prone to errors; we should automate it.

This can be easily achieved using something like a custom Jekyll plugin, but that's not possible as custom plugins are understandably disabled in GitHub pages.  One way around this, is to use a [CI](https://en.wikipedia.org/wiki/Continuous_integration) tool, such as [travis-ci](https://travis-ci.org/), and setup a script that would handle the building of the static files every time a commit is pushed.

### DIFF
--- a/docs/_includes/variables.html
+++ b/docs/_includes/variables.html
@@ -2,7 +2,7 @@
 
 <div class="content">
   <p>
-    You can use these variables to <strong>customize</strong> this {%if include.element %}element{% elsif include.layout %}layout{% else %}component{% endif %}. Simply set one or multiple of these variables <em>before</em> importing Bulma. <a href="{{ site.url }}/documentation/overview/customize/">Learn how</a>.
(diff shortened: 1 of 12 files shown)


jgthms:
> One way around this, is to use a CI tool, such as travis-ci, and setup a script that would handle the building of the static files every time a commit is pushed.

I like this idea! It's not critical but it would definitely help. We should do it at some point. I'll look into it.

bstashio:
👍

PR #1241, opened 2017-09-28
Remove iTemplate from from related projects

mattpham:
I was checking out the related projects and the link to iTemplate project appears to have been hijacked and redirected to a suspect site. It should be removed until @bendo01 fixes the link, or alternatively change to project repository.

**Redirects to freenom.link**
screencap


bstashio:
I agree, it should be removed until @bendo01 fixes it.

PR #1245, opened 2017-09-29
Fix typo "modifer" to "modifier" in all docs

JoshuaCoquelle:


jgthms:
I really can't write that word correctly, can I? 😂

PR #1247, opened 2017-09-30
Fix #1235 , add one-fifth column

### LINKED ISSUE #1235: Support for one-fifth
Is this something that we could consider? I'm working on a design and it had to be 5 columns, I was sad to find out Bulma doesn't support this out of the box.

Naskalin:
fix this #1235 , add one-fifth column
This is really useful, for example, when you need to show more data for width fullHd (but not `<div class="column is-2-fullhd"></div>` for 6 column in row)

```html
<div class="columns is-multiline">
    <div class="column 
                is-half-tablet 
                is-one-third-desktop 
                is-one-quarter-widescreen 
                is-one-fifth-fullhd"></div>
    <!--column-->
    <!--column-->
    <!--column-->
    <!--...-->
</div>
`````


jgthms:
You would also need `two-fifths` `three-fifths` and `four-fifths`, like the other classes. But that would add a lot of code, hence why I didn't want to add it in the first place.

Naskalin:
Ok, I understand, pr updated: add two-fifths three-fifths and four-fifths

jgthms:
Thanks! Can you update the Changelog and the Docs as well?

Naskalin:
jgthms for change log, only line 7

PR #1248, opened 2017-10-01
Feature: allow default table styles to be opt in (instead of mandatory)

prog-rajkamal:
### Proposed solution

I am using bulma and I need to use "<table>" element without bulma's default styles.

However in generic.sass and content.sass, the styles are applied to "table" instead of ".table".
The styles in content.sass, in particular, can not be overridden without "!important" 

This PR changes the selectors from "table" to ".table", so users have to specifically opt in for the generic styles to trigger.
 
### Tradeoffs

There is no trade-off other than adding class "table" for generic styles to kick in.

Technically, this is a minor breaking change. However bulma docs advocate using ".table" for styling anyway. So  this should be standard practice anyway

### Testing Done

Yep, I have successfully built it and tested it against my project.


prog-rajkamal:
Due to line ending issues, this is showing a lot of changes in commit, should I make a new PR.

Issue #1205, opened 2017-09-18
Documentation links to source files


maurobringolf:
### Overview of the problem

This is about the Bulma **Docs**
I'm using Bulma **version** [0.5.2]

### Description

I am using Bulma through sass modules and import only the ones I need. I looked through documentation and found the **level** component, but couldnt find an immediate link or hint as to what file I should import to use it. It is documented under layout, but its source is under components. Why is that?

I would suggest adding a link to sass source files for each documentation page, so users can immediately see what modules they need to import.

jgthms:
Yes I moved it in the documentation as I realised it's more of a layout component. I might move it in the source but it might break builds.

maurobringolf:
Okay and what do you think of providing a link/ sass include path for each module directly on the doc pages?

jgthms:
yes why not

PR #1264, opened 2017-10-06
Update responsiveness breakpoints values in docs

### LINKED ISSUE #1252: Why are there two different set of breakpoints?
Why are there 8px difference for `desktop`, `widescreen` and `fullhd`?

In the [docs](http://bulma.io/documentation/overview/responsiveness/):
screen shot 2017-10-03 at 9 20 01 pm

In [`variables.sass`](https://github.com/jgthms/bulma/blob/0f94e971d8a147a434d0e0a7c61dd0c647606898/sass/utilities/variables.sass#L52-L58)
screen shot 2017-10-03 at 9 20 36 pm

bstashio:
### Proposed solution

This PR includes the following updates:
- Fixes #1252 by updating references to `variables` and `mixins` in [responsiveness docs page](http://bulma.io/documentation/overview/responsiveness/)
- Add related responsiveness variables reference table
- Add links to relevant source files - as mentioned in #1205 

### Tradeoffs

None

### Testing Done

Done locally using `jekyll serve --config _config.local.yml`


jgthms:
I like the idea of having "one source of truth" but this doesn't work since `_includes/bp/*` also has these values. I think the solution is to store them in `_data/breakpoints.json` or `.yml` and use it from there.

bstashio:
> I think the solution is to store them in _data/breakpoints.json or .yml and use it from there

Even better!

jgthms:
Hey thanks for your help! I think we can remove the duplication, and rely on the name rather than the index.

So if you have `_data/breakpoints.json` with
```
{
  "mobile": {
    "from": 0,
    "to": 768
  },
  "tablet": {
    "from": 769,
    "to": 1023
  },
  "desktop": {
    "from": 1024,
    "to": 1215
  }
}
```
then you can access it with `site.data.breakpoints.mobile` instead, and remove the duplication from `docs/documentation/overview/responsiveness.html`.

bstashio:
My pleasure!

Much better if we use the name instead of the index, thanks for pointing out.

I'm on the move right now, and won't be able to get to near my main machine until tomorrow, so please go ahead and update it any way you see fit.

bstashio:
Updated to `json`.

jgthms:
Great stuff!

docs/documentation/overview/responsiveness.html
```diff
@@ -27,81 +38,71 @@ <h2 class="subtitle">Bulma is a <strong>mobile-first</strong> framework</h2>
...
+      <p>Bulma has <a href="{{ variables_file_url }}" target="_blank">5 breakpoints</a>:</p>
       <ul>
-        <li><code>mobile</code>: up to <code>768px</code></li>
-        <li><code>tablet</code>: from <code>769px</code></li>
-        <li><code>desktop</code>: from <code>1008px</code></li>
-        <li><code>widescreen</code>: from <code>1200px</code></li>
-        <li><span class="tag is-success">New!</span> <code>fullhd</code>: from <code>1392px</code></li>
+        {% for breakpoint_hash in site.data.breakpoints %}
+          {% assign breakpoint = breakpoint_hash[1] %}
+          <li>{% if breakpoint.id == 'fullhd' %}<span class="tag is-success">New!</span> {% endif %}<code>{{ breakpoint.id }}</code>: {% if breakpoint.id == 'Mobile' %}up to <code>{{ breakpoint.to }}px</code>{% else %}from <code>{{ breakpoint.from }}px</code>{% endif %}</li>
```
bstashio: Sorry about that, I just noticed that I used `Mobile` instead of `mobile`;

Should be `{% if breakpoint.id == 'mobile' %}` instead of `{% if breakpoint.id == 'Mobile' %}`.

Can you update it or do you want me to create a new PR?

Issue #1295, opened 2017-10-12
Disable hover styling on tables within .content class by default


admench:
This is about **Bulma**.

### Overview of the problem

This is about the Bulma **CSS framework**
I'm using Bulma **version** [0.6.0] 

### Description

Referencing #1236 - this issue of the table row `:hover` behaviour is also set as default to `.table` inside `.content` class elements. I propose that we remove this behaviour as default, inline with #1236 as it caught me out after upgrading bulma just to avoid this. Will submit a PR.

### Expected behavior

When hovering on a table, by default nothing should happen, only if `.is-hoverable` is added should the visual state change.

### Actual behavior

When interacting with a `.table` within `.content` the hover state is applied by default.

Issue #1236, opened 2017-09-27
Considering is-hover for tables to decide wether hovering is applied or not


seb3s:
This is about the Bulma **CSS framework**
I'm using Bulma **version** [0.5.3]

### Description

Hi, I propose to add an is-hover modifier to tables.
The tr:hover style would only be applied to tables with that modifier.
In my usage, not all tables requires hovering style and it is better for me that way.

What do you think ?
Thanks a lot,

Sébastien.

PR #1296, opened 2017-10-12
Disable table hover in .content #1295

admench:
This is about **Bulma**.

### Overview of the problem

### Description

Referencing #1236 - this issue of the table row `:hover` behaviour is also set as default to `.table` inside `.content` class elements. I propose that we remove this behaviour as default, inline with #1236 as it caught me out after upgrading bulma just to avoid this. 

### Expected behavior

When hovering on a table, by default nothing should happen, only if `.is-hoverable` is added should the visual state change.

### Actual behavior

When interacting with a `.table` within `.content` the hover state is applied by default.

### DIFF
--- a/docs/documentation/elements/content.html
+++ b/docs/documentation/elements/content.html
@@ -26,8 +26,6 @@ variables:
   value: 0.5em 0.75em
 - name: $content-table-cell-heading-color
   value: $text-strong
-- name: $content-table-row-hover-background-color
-  value: $background
 - name: $content-table-head-cell-border-width
   value: 0 0 2px
 - name: $content-table-head-cell-color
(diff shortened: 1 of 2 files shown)


jgthms:
Thanks! Can you remove the `$content-table-row-hover-background-color` from the `.sass` file and the docs, and update the changelog?

admench:
Absolutely. Wow this is exciting. My first EVER PR!

admench:
I have done as requested and removed the unnecessary variable in the sass file and updated the docs to remove the variable mentioned. Does that cover it for the docs?

jgthms:
Just need to rebase.

admench:
Is there anything that I can do to help?

lmbrochu:
@admench you're the one supposed to rebase.

andrewsosa:
Since this PR is dead, I've opened a duplicate as #1707, also updating the new docs format.

PR #1298, opened 2017-10-12
Cleanup docs deprecated versions directories

bstashio:

### Proposed solution

I was conducting a local docs test by running `jekyll serve --config _config.local.yml`
as usual and encountered the following error:

```
FATAL: Listen error: unable to monitor directories for changes.
Visit https://github.com/guard/listen/wiki/Increasing-the-amount-of-inotify-watchers for info on how to fix this.
```

Investigating the issue revealed that it is caused by the `inotify` limit, which is usually set to a default `8192` on Linux distros; meaning, it will monitor a maximum of `8192` directories for changes - [details here](https://github.com/guard/listen/wiki/Increasing-the-amount-of-inotify-watchers#the-technical-details).  That was an indication that something is not right about that number, so when I checked the `./docs` directory structure I realized that old deprecated versions included directories of older deprecated versions directories (recursion); for instance `docs/versions/0.5.3` contained a `versions` directory that contained all the other previous deprecated versions.  As you can imagine, that dramatically increased the number of directories and files unnecessarily.

In addition to all the extra files and directories that were part of the deprecated versions of the docs, I also found that the `versions` directory contained `0.6.0`, the current doc's version.  Not only that causes an even bigger versions' directory, but it was also used as the URL to the current version, which redirects to `http://bulma.io/versions/0.6.0/` instead of `http://bulma.io/`. Here's a screenshot of the deprecated notice.

bulma_v0 6 0_shown_as_deprecated

In summary this PR fixes the following:

- Update current version URL in [docs versions page](http://bulma.io/versions/)
- Delete not needed current version (0.6.0) directory from `versions` directory 
- Delete all unnecessary `versions` directories in deprecated `docs/versions`

Some before and after data in relation to repo size:

| Before | After | Savings |
|--------|-------|---------|
| 419MB  | 93MB  | 326MB   |

### Tradeoffs

None detected.

### Testing Done

Tested on local machine using `jekyll serve --config _config.local.yml` and
`jekyll serve --incremental --config _config.local.yml`


jgthms:
Hey thanks, I was actually working on this myself. I'm planning on moving the old versions to another repo.

My local config file actually excludes those directories, and I've updated my versions config file to exclude them from now on.

As for the `0.6.0` version, it needs to be archived as soon as it's released, since it needs to match the release.

bstashio:
Sounds good.

> My local config file actually excludes those directories

That was the initial hack I used to get it running; I added `versions/*` to the `exclude` array in `_config.local.yml`.  Probably we need to add this to `docs/README.md`, what do you think?

PR #1309, opened 2017-10-14
Fixes #1308, rainbow button glitch

### LINKED ISSUE #1308: Small visual glitch when `bd-rainbow` class is used in buttons
This is about a Bulma component that was used only in the Docs. 

### Overview of the problem

This is about the Bulma **Docs** 
My **browser** is: Chrome
I am sure this issue is **not a duplicate**? Yes.

### Description
The `bd-rainbow` class is only available in the docs and not publicly available in the production release. I found a visual glitch when I used the same `animate`'s keyframes as that of yours. Furthermore, it's not a browser bug. I was able to reproduce this visual bug in all browsers, including mobiles.

### Steps to Reproduce
1. Use the `bd-rainbow` class in a button.
2. Observe the edges of the button. The gradient does not have a smooth transition.

### Expected behavior
The rainbow gradient should rather transition smoothly without those hard edges.

### Actual behavior
Closely observe the edges when the keyframes are at 0% and 100%. The higher the resolution, the more pronounced is the glitch.
<p align="center">

</p>

jeremyphilemon:
This is a **bugfix**.

### Proposed solution
Initially, the keyframes transition from 0% to 100%, traversing till the edges of the gradient background and producing the glitch. This PR fixes it by making the transition happen between 1% to 99%.

### Tradeoffs
I don't think there is a tradeoff.

### Testing Done
Manually tested on all browsers, including mobile.

### DIFF
--- a/docs/_sass/love.sass
+++ b/docs/_sass/love.sass
@@ -64,11 +64,11 @@
 
 @keyframes rainbow
     0%
-      background-position: 0% 80%
+      background-position: 1% 80%
     50%
-      background-position: 100% 20%
+      background-position: 99% 20%
     100%
-      background-position: 0% 80%
+      background-position: 1% 80%
 
 .bd-hug
   align-items: flex-start


jgthms:
Interesting! I’ll take it.

PR #1311, opened 2017-10-14
(RTL) right to left

khalil88:
Could you please add right to left to the framework. so we would be able to create our template from right for some countries it is really useful.


jgthms:
Did you write this code? Having 7000 files doesn’t seem right.

PR #1318, opened 2017-10-16
Active Bulma - Jump start your Bulma creation.

RuinSain:
ActiveBulma is a simple yet powerful Javascript implementation for most of bulma's components that needs javascript in order to interact with them. just drop it in your project and Your good to go with modal, dropdown, and the navbar burger. 
In order to to use the modal, call the "openModal()" function. everything ese just works out of the box.

This is a **JavaScript implementation for modal, burger, and dropdown**.

### Proposed solution
A tiny Javascript file to jump start your bulma creation.

### Tradeoffs
None.

### Testing Done
Used laravel dusk in order to check everythoing is working.
Compatible with all browsers.

### DIFF
--- a/ActiveBulma
+++ b/ActiveBulma
@@ -1,0 +1,36 @@
+/* ActiveBulma made by Gabriel Michaeli @RuinSain */
+
+
+//Here we will controll the dropdown.
+let dropDown=document.querySelector('.dropdown');
+dropDown.addEventListener('click', function(){
+ dropDown.classList.toggle('is-active');
+});
+
+
+//Here we will controll the modal.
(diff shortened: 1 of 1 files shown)


VizuaaLOG:
If this is merged in (because Bulma is CSS only) it would need to be transpiled. `let` may cause issues for older browsers that still support Bulma.

RuinSain:
Yes, i have changed it to var so this wont happen now.

jgthms:
This PR is a bit weird. The file has no extension?

I also believe this piece of JavaScript could exist as a Bulma extension instead. Feel free to add it to the https://bulma.io/extensions/ page!

PR #1324, opened 2017-10-17
Add list of buttons

jgthms:
New list of buttons.


paulvl:
Docs says that is available on version **0.6.1** but... there is no version **0.6.1** available yet.

PR #1328, opened 2017-10-17
Stop is-selected row from changing color on hoverable row

### LINKED ISSUE #1313: Table is-selected and is-hoverable styling issue
This is about **Bulma**.

### Overview of the problem

I'm using Bulma **version** [6.0.0] 

### Description

When using a table with is-hoverable and is-selected the selected items hover styles are unreadable.

### Expected behavior

Not sure maybe do not do hover on selected items. 

### Actual behavior

Selected Item
screen shot 2017-10-14 at 2 55 08 pm

Selected Item on Hover
screen shot 2017-10-14 at 2 55 15 pm

njgingrich:
This is a **bugfix**, fixing #1313.

### Proposed solution
Fixes #1313, where instead of using the base active hover color for a table, will only apply that color if the table itself is not hoverable.

### Tradeoffs
This doesn't show any visual difference when hovering over an row that has `is-selected` applied to it.

### Testing Done
I checked the style based on the example in the documentation by adding `is-hoverable` to the table demonstrating the `is-selected` class.

Side note: I'm not entirely sure why git considers this entire file to be changed, I only changed line 88.

### DIFF
--- a/sass/elements/table.sass
+++ b/sass/elements/table.sass
@@ -85,7 +85,7 @@ $table-striped-row-even-hover-background-color: $white-ter !default
     width: 100%
   &.is-hoverable
     tbody
-      tr
+      tr:not(.is-selected)
         &:hover
           background-color: $table-row-hover-background-color
     &.is-striped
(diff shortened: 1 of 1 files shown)


jgthms:
Thanks but you need to fix your PR. It should only have a one line change. Try to rebase against master first.

njgingrich:
I've fixed the PR, thank you! Was an issue with line endings (CRLF vs LF).

PR #1334, opened 2017-10-18
Fix progress bar colors in IE

### LINKED ISSUE #1333: Progress bar color cannot be changed in IE11
This is about **Bulma**.

### Overview of the problem

I'm using Bulma **version** 0.6.0
My **browser** is: IE11
I am sure this issue is **not a duplicate**? Yes

### Description

In IE11, the progress bar color does not change when the CSS color modifier classes are added:

ie_colors

### Steps to Reproduce

The following HTML reproduces the issue:

```
<progress class="progress is-success" value="60" max="100">60%</progress>
```

The issue is also visible by viewing the docs at https://bulma.io/documentation/elements/progress/.

### Expected behavior

The progress bar color respects the `is-info`, `is-success`, etc. CSS modifier classes.

### Actual behavior

The progress bar color is always the default color regardless of the modifier classes.

The workaround in my project is to use the following CSS.

```
.progress.is-success::-ms-fill {
    background-color: hsl(141, 71%, 48%);
}
.progress.is-info::-ms-fill {
    background-color: hsl(204, 86%, 53%);
}
...
```

I'll create a PR to add the necessary `-ms-fill` declarations.

rafeememon:
This is a **bugfix** for progress bar colors in IE.

### Proposed solution

This PR adds the necessary `::-ms-fill` declarations to style the progress bar in IE, as explained here: https://developer.mozilla.org/en-US/docs/Web/CSS/::-ms-fill

Fixes #1333.

### Tradeoffs

This slightly increases the size of the build, but improves support for IE.  There are other `-ms-*` declarations in the code base so I think it is reasonable to add this.

### Testing Done

I built my branch of Bulma locally, added it to my project, removed my workaround, and confirmed that the progress bar color can be changed:

ie_progress_fixed

### DIFF
--- a/sass/elements/progress.sass
+++ b/sass/elements/progress.sass
@@ -18,6 +18,8 @@ $progress-value-background-color: $text !default
     background-color: $progress-value-background-color
   &::-moz-progress-bar
     background-color: $progress-value-background-color
+  &::-ms-fill
+    background-color: $progress-value-background-color
   // Colors
   @each $name, $pair in $colors
     $color: nth($pair, 1)
@@ -25,6 +27,8 @@ $progress-value-background-color: $text !default
       &::-webkit-progress-value
         background-color: $color
       &::-moz-progress-bar
+        background-color: $color
+      &::-ms-fill
         background-color: $color
   // Sizes
   &.is-small


rafeememon:
Thanks!

PR #1343, opened 2017-10-19
Adds sup and sub to titles and subtitles

SimplyThomas:
This is a **new improvement**.

### Proposed solution
This PR fixes sizing when using either `<sub>` and `<sup>` elements inside a .title element.

### Testing Done
Compiled with yarn and tested across all of the various sizes and configurations.

Before:
screen shot 2017-10-19 at 11 31 59 pm

After:
screen shot 2017-10-19 at 11 32 15 pm
screen shot 2017-10-19 at 11 32 32 pm

### DIFF
--- a/sass/elements/title.sass
+++ b/sass/elements/title.sass
@@ -3,12 +3,16 @@ $title-size: $size-3 !default
 $title-weight: $weight-semibold !default
 $title-strong-color: inherit !default
 $title-strong-weight: inherit !default
+$title-sub-size: 75% !default
+$title-sup-size: 75% !default
 
 $subtitle-color: $grey-dark !default
 $subtitle-size: $size-5 !default
 $subtitle-weight: $weight-normal !default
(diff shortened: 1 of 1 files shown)


jgthms:
I’d rather keep it consistent with the `.content` sup and sub who have a font size of 75%.

SimplyThomas:
@jgthms this is also a better implementation. Tested in the same way as before, cycling through the various sizes. Also verified that the variables work as intended.

jgthms:
Why does it appear twice? Just add it to the first rule set.

SimplyThomas:
@jgthms One's for .title, the other is for .subtitle

jgthms:
I’m not sure we would want a different size for both title and subtitle since they can both be all sizes from 1 to 7 anyway.

SimplyThomas:
@jgthms I have updated it to reflect just one size.

Issue #963, opened 2017-07-30
Delete Button Bug in iOS Safari (inside Tags, Messages and Notifications)


inseldesign:
![img_0156](https://user-images.githubusercontent.com/5708900/28753761-96aaf428-7539-11e7-8277-1e8cbf7f786f.PNG)

### Overview of the problem

Delete button with messages on ios safari. Happens only inside tag, messages and notifications.

ialhashim:
I found a hack by setting padding to zero

alcovegan:
Looks like same as #514 and proposed solution in #549, but not merged.

johan13:
I think it's also the same as #490. Both #514 and #490 have been closed without fixing the problem. I hope this issue doesn't face the same fate.

mdavid88:
It needs font-size: 0.

https://stackoverflow.com/questions/3777494/is-it-possible-to-override-the-minimum-width-of-a-button-in-safari-on-ipad

PR #1357, opened 2017-10-22
Fix for issue #963.

mdavid88:
iOS <button> element's minimum width is dependant on its font-size, so the .delete width wasn't able to be as small as it's height, resulting in one arm of the X being longer than the other. font-size: 0 does the trick.

Bugfix. Fixes issue #963 

### Proposed solution
Fixes issue #963 

### Tradeoffs
Only drawback is if you put text inside your Button, which you shouldn't as it's an icon only.

### Testing Done
Tested in iOS, Safari, Firefox, and Chrome on Mac. No access to IE.

### DIFF
--- a/sass/utilities/mixins.sass
+++ b/sass/utilities/mixins.sass
@@ -42,7 +42,7 @@
   display: inline-block
   flex-grow: 0
   flex-shrink: 0
-  font-size: $size-normal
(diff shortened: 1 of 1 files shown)


jgthms:
Fantastic! Thanks.

Issue #1201, opened 2017-09-17
`.tag` and `.number` are used by code highlighting tools


znck:
### Overview of the problem

This is about the Bulma **CSS framework**

### Steps to Reproduce

```
<pre>
   <span class="token tag">foo</span>
   <span class="token number">12</span>
</pre>
```

### Expected behavior

Should not override highlight style.

jgthms:
That's tricky to do, since there's no way of knowing what a highlighter's CSS classes are. The only way is to include everything except the tags and numbers.

PR #1369, opened 2017-10-25
Code Highlighter Collision

mackhankins:
### Improvement
I see it was previously discussed in #1201.

### Proposed solution
This is a common class used by most code highlighters.  I can't find where we're using it in Bulma, but simply changing the class to .numbers instead fixes the problem.

### Testing Done
result

### DIFF
--- a/sass/elements/other.sass
+++ b/sass/elements/other.sass
@@ -24,7 +24,7 @@
 .loader
   +loader
 
-.number
+.numbers
   align-items: center
   background-color: $background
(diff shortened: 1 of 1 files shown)


jgthms:
You need to change the HTML code as well in the docs.

predic8:
I want also to use bulma together with prism. But the collision impede that. Hope we get a fix soon.

Thanks for the great framework. I really enjoy working with it!

PR #1372, opened 2017-10-26
Remove whitespaces

zainfathoni:
Remove whitespaces to make it consistent with other entries

I just spotted an inconsistency and want to fix it immediately.


jgthms:
Thanks but they should all have whitespace actually. My mistake!

jgthms:
I reckon it should live in a variable in the `_data` folder.

Issue #1278, opened 2017-10-09
is-selected for table columns


Cosbgn:
I understand that this might be complicated and probably would require js so won't be part of bulma. But is there an easy way to add "is-selected" to columns in a table, in a similar way we can add "is-selected" to rows?

rafaelrenanpacheco:
The `is-selected` could be available to cells as well, this way you would choose which cells should be marked as selected.

If your request does't make into bulma's souce, you can create your own css in your project very easily.

PR #1380, opened 2017-10-29
Allow is-selected class on <td> and <th> tags

pip36:
This is an **improvement**.

### Proposed solution
Allows individual cells in a table to be selected, instead of just rows.
Similar issue is mentioned in issue #1278. 

### Tradeoffs
Let me know if you think there are downsides.

### Testing Done
Have tested locally and doesn't seem to affect existing table classes.

tabletest


jgthms:
This results in `td td` `td th` `th td` and `th th` selectors which are incorrect.

jgthms:
Thanks!

pip36:
No problem. Thank You.

PR #1384, opened 2017-10-29
Add `is-centered|right` modifiers to `.tags`

timacdonald:
This is a **new feature**.

### Proposed solution

This allows `.tags .tag` groups to be aligned centre and aligned right. Currently they are only aligned left.

### Tradeoffs

None come to mind.

### Testing Done

Currently doing this in production, works great. Also, [here is a codepen](https://codepen.io/timacdonald/pen/VrwREG) I created to show the feature.

### DIFF
--- a/sass/elements/tag.sass
+++ b/sass/elements/tag.sass
@@ -8,16 +8,6 @@ $tag-delete-margin: 1px !default
   display: flex
   flex-wrap: wrap
   justify-content: flex-start
-  &.is-centered
-    justify-content: center
-    .tag
-      margin-right: 0.25rem
-      margin-left: 0.25rem
-  &.is-right
-    justify-content: flex-end
-    .tag
-      margin-left: 0.5rem
-      margin-right: 0
   .tag
     margin-bottom: 0.5rem
     &:not(:last-child)
@@ -35,6 +25,16 @@ $tag-delete-margin: 1px !default
       &:not(:last-child)
         border-bottom-right-radius: 0
         border-top-right-radius: 0
+  &.is-centered
+    justify-content: center
+    .tag
+      margin-right: 0.25rem
+      margin-left: 0.25rem
+  &.is-right
+    justify-content: flex-end
+    .tag
+      margin-left: 0.5rem
+      margin-right: 0
 
 .tag:not(body)
   align-items: center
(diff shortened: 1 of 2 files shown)


sass/elements/tag.sass
```diff
@@ -8,6 +8,16 @@ $tag-delete-margin: 1px !default
   display: flex
   flex-wrap: wrap
   justify-content: flex-start
+  &.is-centered
```
jgthms: Can you move these after `&.has-addons`?

sass/elements/tag.sass
```diff
@@ -8,6 +8,16 @@ $tag-delete-margin: 1px !default
...
+    justify-content: center
+    .tag
+      margin-right: 0.25rem
+      margin-left: 0.25rem
+  &.is-right
+    justify-content: flex-end
+    .tag
```
jgthms: You need to use `.tag:not(:last-child)`

timacdonald: @jgthms thanks for taking the time to review this. I might have misunderstood, but if I use `.tag:not(:last-child)` in this block:

``` sass
&.is-right
    justify-content: flex-end
    .tag
      margin-left: 0.5rem
      margin-right: 0
```

It will cause the last child to not get the appropriate left margin - which will look like this:

screen shot 2017-11-05 at 9 50 13 pm

If I'm missing something let me know and I'm more than happy to fix it up - thanks for the great framework!

jgthms: Ah sorry, it should be this I believe:

```
&.is-right
  .tag
    &:not(:first-child)
      margin-left: 0.5rem
    &:not(:last-child)
      margin-right: 0
```

timacdonald: @jgthms  yup - I get you. makes sense - more targeted. Thanks. I've updated that for you.

jgthms:
Thanks! That's great.

vybhavB:
Just a question but does this work? I was trying to use it in my div for tags but it doesn't seem to be pulling them to the right

timacdonald:
@vybhavB sure is, check out this Codepen for an example of it in action: https://codepen.io/timacdonald/pen/QVVoLw

vybhavB:
@timacdonald hmm interesting. I have a div inside a panel-block. Not sure if that's what is causing it not to work but none of my tags get pushed right. Heres an example codepen: https://codepen.io/vybhavb/pen/mGGoea

timacdonald:
ahhh, yup, you'll need to throw `width:100%` at the `tags` div: https://codepen.io/timacdonald/pen/KxGwpX

PR #1385, opened 2017-10-29
Fix .pagination-next alignment

### LINKED ISSUE #1364: .pagination-next without a .pagination-previous shows next on left side
https://bulma.io/documentation/components/pagination/

If i use:

```html
<nav class="pagination is-centered" role="navigation" aria-label="pagination">
    <a href="/features/learn-javascript/2" class="pagination-next">Next: Fundamental concepts</a>
  </nav>
```

With no ``pagination-next`` link, (such as the first page of an article), the next page link shows the left.

This doesn't happen on last pages (where there is no ``.pagination-next``.

screen shot 2017-10-24 at 9 31 01 am

I would expect a next link to always show on right.

leomarquine:
This PR fix #1364

Correct `.navigation-next` alignment when it is the only child of `.navigation`.

### DIFF
--- a/sass/components/pagination.sass
+++ b/sass/components/pagination.sass
@@ -54,6 +54,9 @@ $pagination-shadow-inset: inset 0 1px 2px rgba($black, 0.2)
   margin: 0.25rem
   text-align: center
 
+.pagination-next:only-child
+  margin-left: auto
+
 .pagination-previous,
 .pagination-next,
 .pagination-link


sass/components/pagination.sass
```diff
@@ -54,6 +54,9 @@ $pagination-shadow-inset: inset 0 1px 2px rgba($black, 0.2)
   margin: 0.25rem
   text-align: center
 
+.pagination-next:only-child
+  margin-left: auto
```
jgthms: This needs to be only on tablet onwards. You can add it here:

```
&.is-centered
  .pagination-next
    order: 3
```

PR #1431, opened 2017-11-09
Fix default sample code in customize.html

evantill:

Actually, the default code will result in an unexpected behavior: the `$color` map is defined in `derived-variables.sass` before setting the custom colors.

### Proposed solution
move the `derived-variables import`

### Tradeoffs
IMHO the lists and maps should be defined in a dedicated sass file to avoid confusion in customization.

### Testing Done
tested  in my project.


jgthms:
I like this.

PR #1432, opened 2017-11-09
Fix navbar-burger sample code in navbar.html

stern-shawn:

### Proposed solution
button elements with the class combination of 'button' and 'navbar-burger' display with a border. Updating the buttons in the documentation to divs, and removing button class.

Other parts of the documentation, as well as the documentation site itself, use a div with class="navbar-burger" instead of a button. Normalizing the examples so others don't experience the same headache that I did from using html button and classes as presented in the docs :)

### DIFF
--- a/docs/documentation/components/navbar.html
+++ b/docs/documentation/components/navbar.html
@@ -20,7 +20,7 @@ doc-subtab: navbar
 {% endcapture %}
 
 {% capture navbar_burger_example %}
-<button class="button navbar-burger">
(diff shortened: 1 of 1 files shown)


oleksiipotiekhin:
Good job! I think @jgthms will merge it before doing a release.

stern-shawn:
Thank you! Hope it helps.

stern-shawn:
Great catch, thanks @feute !

PR #1475, opened 2017-11-24
Clean up Wikiki extensions

Wikiki:
Clean up Wikiki extensions links


jgthms:
Why not update the URL for each of them? It's a bit confusing to have one single image for all your extensions.

Wikiki:
I find that having a static image for the extension (which evolve) it not the best way. And adding an image each time I develop a new extension and settings ratio for each update is not convenient.
Maybe we could review the way of presenting the extensions ?

Added to this, I think we could use a grid layout to display extensions, it's to big currently.

Wikiki:
What about the PR https://github.com/jgthms/bulma/pull/1487 as new extension page layout ?
No image size needed anymore (I have created a new image modifier <code>.is-contained</code> to force image to fit in its container)

Issue #1355, opened 2017-10-22
Warning --columnGap ! Lack of consistency in var...


jimblue:
This is about **Bulma**.

It's a bug, but not in browser

### Overview of the problem

This is about the Bulma **CSS framework**
I'm using Bulma **version** [0.6.0]
My **browser** is: google chrome
This is a **Sass** issue: I'm using version [3.5.2]
I am sure this issue is **not a duplicate*#

### Description

While trying to compile a custom version of Bulma by following the documentation step I'm getting warning in the console about `--columnGap` in `columns.sass`. This append because this var is not scoped to the top-level `:root`.

My question is why you're not using classic SASS variable here. This place using a different writing for var in all Bulma...

Can you replace `--columnGap` with `$columnGap` this will work the same.

Thank you

jgthms:
It’s a CSS variable. It makes it easier to update the variable value for the variable gap feature. 

It shouldn’t be scoped to the root level. I believe there was another issue mentioning his, but it’s not really an issue since CSS variables are supposed to be used that way.

jimblue:
Hi @jgthms and thank for answering that fast!

I know this is a CSS variable, it's even more the only one in all Bulma. That's why I was asking...

Keep everything working with the same way is always better.
Actually without going inside Bulma architecture it's impossible to know about this CSS variable in initial-variables as root. Also it's not document.

I hope you understand my point 😄

jimblue:
The warning happen in a really classic configuration (webpack - sass - postcss - cssnext - cssnano).

HEre is the log message:

```bash
 warning  in ./src/scss/build/global_style.scss

(Emitted value instead of an instance of Error) postcss-custom-properties: /Users/jim/Sites/jimblue/node_modules/bulma/sass/grid/columns.sass:328:6: Custom property ignored: not scoped to the top-level :root element (.columns.is-variable.is-5 { ... --columnGap: ... })
```
So I'm was wondering what's your configuration, do you have this warning too?
(it's pretty annoying when refreshing)

jimblue:
As you can see it's a lot of warning, pretty disturbing !

<img width="1920" alt="screen shot 2017-10-23 at 15 58 50" src="https://user-images.githubusercontent.com/10623367/31893213-4820fac2-b80b-11e7-9004-8fcfbade7b87.png">

jimblue:
Finally found a way to hide the warning, all good. tks

juanpmarin:
[...]

Issue #1283, opened 2017-10-10
[SASS warning] Sass 3.6 will change the way CSS variables are parsed 


guillecro:
This is about **Bulma | the Docs**.

### Overview of the problem

This is a **Sass** issue: I'm using version [3.5^] (I guess) 
I am sure this issue is **not a duplicate**? 

### Description

So I got this warning message when i was doing `compass watch`, thought you might want to know it

```
DEPRECATION WARNING on line 328 of /home/mapache/www/residencias-culturales-2018/bower_components/bulma/sass/grid/columns.sass:
Sass 3.6 will change the way CSS variables are parsed. Instead of being parsed as
normal properties, they will not allow any Sass-specific behavior other than #{}.
For forwards-compatibility, use #{}:

  --variable: #{$i * 0.25rem};
```

Ok bye!

jgthms:
Thanks for the heads up. The only instance is `--columnGap` for now. I'll update that.

ADTC:
Am I right to say that this is not applicable to Sass variables?

jgthms:
Works with Sass 3.5.3.

Issue #1210, opened 2017-09-19
SASS Version 3.5.1 compiles custom properties wrong


samuelmoser:
### Overview of the problem

This is about the Bulma **CSS framework**

I'm using Bulma **version** [0.5.3]

This is a **Sass** issue: I'm using version 3.5.1

### Description
I'm a total noob and I'm not sure if it's a Sass or a Bulma misbehavior. With the newest Version of Sass 3.5.1 the columns.isVariable part compiles to

.columns.is-variable.is-1 {
  --columnGap: $ i *0.25px;
}
[...]

instead of (node-sass Version 3.5.0.beta.2)
.columns.is-variable.is-1 {
  --columnGap: 0.25px;
}
[...]

it's in the last part of columns.sass.

### Steps to Reproduce
Compile bulma with the newest Sass Version.

### Expected behavior
.columns.is-variable.is-1 {
  --columnGap: 0.25px;
}

### Actual behavior
.columns.is-variable.is-1 {
  --columnGap: $ i *0.25px;
}

mofahead:
Seeing this here on Sass 3.5.3. Pretty annoying, as a Rails app with up to date rails-sass gem will have this problem with Bulma and break the variable column gap functionality. 

Apparently this is correct behavior according to Sass from 3.5 going forward; custom property values are read as literal text, so anything dynamic needs to be surrounded by #{}.

Doesn't break currently using node-sass (tried with v3.8.0), but looks like this will change in the future ([see this discussion at LibSass](https://github.com/sass/libsass/issues/2076)). 

The fix for Bulma is easy, simply change the last line of columns.sass from
`      --columnGap: $i * 0.25rem`
to
`      --columnGap: #{$i * 0.25}rem`

I've patched mine and it now works on both sass and node-sass. =)

mofahead:
Turns out this is a dupe of [1283](https://github.com/jgthms/bulma/issues/1283), the only new info is that this affects Sass v3.5.

jgthms:
Works with Sass 3.5.3.

Issue #1181, opened 2017-09-12
Strange css property: `--columnGap`


skyghis:
### Overview of the problem

This is about the Bulma **CSS framework**
I'm using Bulma **version** [0.5.2]

### Description

The provided css contains a strange key: `--columnGap` @ [bulma.css:8849](https://github.com/jgthms/bulma/blob/0.5.2/css/bulma.css#L8849)

My IDE (Netbeans) complains about it.  
I'm not a css expert but this seem strange to me too.

If this key is expected sorry for the false report.

This value was added by PR [Variable columns #1073](https://github.com/jgthms/bulma/commit/6fe43b8f945b2e714fe3e1f9a098f79ac710c52e#diff-137ef0f5e3de878e27d30916c1415c9aR322)

jgthms:
It's a CSS Variable: https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_variables

skyghis:
Ok, sorry I didn't know this css feature.

It's the first variable used in bulma. Why use a variable in this case and not directly the expected value as usual ?

I also note that css variables are note as `experimental` by MDN.

jgthms:
Variable gap is noted as experimental: http://bulma.io/documentation/columns/gap/#variable-gap

skyghis:
Sorry I didn't see that :disappointed: .
Thanks for taking time to reply.

jgthms:
CSS Variables are way more convenient as it only requires to change the variable value to have different layouts: https://github.com/jgthms/bulma/blob/master/sass/grid/columns.sass#L319

skyghis:
I get your point, but this would also be true for many other parts of Bulma who doesn't use variables.

Why use variable for some parts of the code only ?

PTcampbell:
Shopify parses .scss on the server and it won't compile due to this error, for what it's worth.

PR #1480, opened 2017-11-27
Fixes #1190 - Use Sass variable and Sass mixin to generate is-variable rules

### LINKED ISSUE #1190: Problem with columnGap property
Hello Bulma'ers!

Just catched this WARNING message when compiling my Sass with Webpack

`(Emitted value instead of an instance of Error) postcss-custom-properties:
/<project-path>/node_modules/bulma/sass/grid/columns.sass:328:6:
Custom property ignored: not scoped to the top-level :root element 
(.columns.is-variable.is-8 { ... --columnGap: ... })`

ADTC:
This is a **improvement | bugfix | documentation fix**. _(all)_
- Improvement because, it drops the dependency on browser support for CSS variables. Depends purely on Sass pre-processing, so it will be compatible with browsers that don't have CSS variable support.
- Bugfix because of the referenced issue.
- Documentation fix because we can remove the warning about browser support in the `is-variable` docs.

### Proposed solution
Instead of depending on CSS variable browser support, use a Sass variable and a Sass mixin to generate regular CSS for the is-variable class. This should increase browser compatibility as well as fix #1190, #1355 and #1181 where warnings are emitted when using a CSS variable. It will also make the deprecation warning in #1283 and #1210 completely irrelevant since that only applies to CSS variables.

### Tradeoffs
I guess you have to learn how Sass variables and Sass mixins work? But other than that, I can't think of anything.

### Testing Done
Injected the generated CSS code into Bulma Docs. Confirmed working:

bulma-is-variable-removed
bulma-is-variable-new-css

### DIFF
--- a/docs/documentation/columns/gap.html
+++ b/docs/documentation/columns/gap.html
@@ -201,13 +201,5 @@ doc-subtab: gap
 
     {% include klmn.html %}
 
-    <div class="message is-warning">
-      <div class="message-body">
-        <p>
-          This feature is only available in browsers that support <a href="https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_variables">CSS Variables</a>:
-        </p>
-      </div>
-    </div>
-
   </div>
 </section>
--- a/sass/grid/columns.sass
+++ b/sass/grid/columns.sass
@@ -464,13 +464,15 @@ $column-gap: 0.75rem !default
     &.is-desktop
       display: flex
 
-.columns.is-variable
-  --columnGap: 0.75rem
-  margin-left: calc(-1 * var(--columnGap))
-  margin-right: calc(-1 * var(--columnGap))
+@mixin columns-is-variable($columnGap: 0.75rem)
+  margin-left: #{-1 * $columnGap}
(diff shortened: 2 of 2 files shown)


jgthms:
I know I could have done it this way but I want to use CSS variables because that’s the future. It’s also more elegant, and requires way fewer lines of code.

The solution here is to simply fix those deprecation warnings.

ADTC:
I don't know about the future, but what do you mean by _"way fewer lines of code"_? Looking at the comparison, the original code has **10 lines**, vs the modified code has **11 lines** (plus a blank line, but I won't count that).

That's merely a single line of code additional for several advantages. I don't see how saving 1 line gives you "way fewer lines".

jgthms:
Surely you understand that I’m talking about the CSS output, not the source file…

Also, CSS variables have very good browser support, and it’s getting better. The only issue is the Sass compilation. Why not simply fix that?

Columns like this can also be implemented with floats. Why not use that then?

Bulma started with the idea of moving forward, with flexbox notably. This PR goes against that idea.

PR #1487, opened 2017-11-30
Feature/redesign extension page

Wikiki:
New Extension page layout idea (following discussion on PR https://github.com/jgthms/bulma/pull/1475)

No image size needed anymore (I have created a new image modifier .is-contained to force image to fit in its container)


docs/extensions.html
```diff
@@ -3,86 +3,66 @@
...
+  slug: bulma-iconpicker
+- name: bulma-slider
+  url: https://wikiki.github.io/form/slider
+  slug: bulma-slider
+- name: bulma-switch
+  url: https://wikiki.github.io/form/switch
+  slug: bulma-switch
+- name: bulma-tagsinput
+  url: https://wikiki.github.io/form/tagsinput
+  slug: bulma-tagsinput
```
jgthms: Can you keep the original order?

docs/extensions.html
```diff
@@ -109,23 +89,30 @@ <h1 class="title">
...
             <h4 class="subtitle is-6">
-              {{ extension.url | remove:'http://' | remove:'https://' | remove:'www.' }}
+            {{ extension.description }}
             </h4>
           </div>
-          <div class="column">
-            <img src="{{site.url}}/images/extensions/{{ extension.slug }}.png" width="{{ extension.width }}" height="{{ extension.height }}">
+          <div class="card-footer">
+            <div class="card-footer-item">
+              <a href="{{ extension.url }}"><i class="fa fa-eye"></i> Extension page</a>
```
jgthms: This makes the click zone way smaller than before. I don't see it as an improvement.

sass/elements/image.sass
```diff
@@ -29,6 +29,10 @@ $dimensions: 16 24 32 48 64 96 128 !default
     padding-top: 56.25%
   &.is-2by1
     padding-top: 50%
+  &.is-contained
+    img
+      object-fit: contain
+      object-position: center center
```
jgthms: This is nice but should be in a separate PR with proper documentation.

docs/extensions.html
```diff
@@ -109,23 +89,30 @@ <h1 class="title">
...
-    {% for extension in page.extensions %}
-      <a class="box" href="{{ extension.url }}">
-        <div class="columns">
-          <div class="column is-one-third">
-            <h3 class="title is-4">
-              <strong>{{ extension.name }}</strong>
-            </h3>
+    <div class="columns is-multiline">
+      {% for extension in page.extensions %}
+      <div class="column is-3">
```
jgthms: This makes the layout very narrow. You can't really see what the extension is about. The image is the most important part and should be the bigger element here.

screen shot 2017-11-30 at 18 19 40

PR #1323, opened 2017-10-17
is-rounded

oleksiipotiekhin:
This is a **new improvement**.

### Proposed solution
 #1028 Class **is-rounded** was added to other _elements/components_ where it didn't break an esthetics and gave additional functionality to them. The reason behind this is to have ability to create something similar to those UI elements:

Angular

### Testing Done
Tested by running `npm run build` successfully. Also as you can see in attached picture all _elements/components_ were checked manually. Plus manual regression testing was done by checking that _elements/components_ that previously were using **is-rounded** class are still working.

### Additional Notes
I've set `$radius-rounded` to `290486px` as this was the previous value used in **tag** and **delete** _elements/components_. Please let me know if you like this idea and i will update documentation also. This pull request contains only `sass` changes

is-rounded
Buttons

PR #1492, opened 2017-12-01
 $radius-rounded: 290486px !default

oleksiipotiekhin:
This is an **improvement**.

### Proposed solution
By adding `$radius-rounded: 290486px !default` variable we will have a possibility to control **rounding** in a single place. Also we will be able to use this variable in extensions etc.

### Tradeoffs
There are no tradeoffs.

### Testing Done
Tested by running `npm run build` successfully.


pavi2410:
Can't we use `$radius-rounded: 50% !default`?

oleksiipotiekhin:
@pavi2410 You can't use `$radius-rounded: 50% !default` for `<input type="text" name="">` for example. But by using `$radius-rounded: 300000px !default` you are able to set round corners for **all** elements. 
Example

oleksiipotiekhin:
@jgthms you can close this one, because it initially was a smaller part of https://github.com/jgthms/bulma/pull/1323

Issue #1502, opened 2017-12-05
PLEASE! - use SCSS by default on bulma or implement 2 filetypes...


leandroluk:
Is it about Bulma

### Overview of the problem

I'm using Bulma **version** [0.6.1]

### Description

Hello, I'm starting to study the bulma to apply in a project but I'm having difficulties to use only the SASS framework. is there any way to create it for SCSS? I think it would be interesting to convert since the build API already keeps this native ...

jgthms:
Is there a particular reason why you would need Scss?

leandroluk:
Hello, first of all I'm sorry for my English, Google Translator is helping me ...

I have been working with other frameworks for a long time and to this day I have not found any that use (still) SASS files ... and also I have not found any viable and functional solution to import SASS into SCSS. So far, the only solution I have is to use this package:

[https://github.com/CyCoreSystems/bulma](https://github.com/CyCoreSystems/bulma)

That is no more than a copy of the original project but already transformed into SCSS ... I believe it is not so difficult to natively implement in the project files in SCSS since it is possible to compile the file from SASS to SCSS ...

jgthms:
You can import sass files into your scss files. It’s the same engine, just a different syntax.

feute:
As @jgthms states, you can import sass files into scss. If you just want to convert them to scss for learning purposes and study Bulma with that syntax you can convert every sass file with `sass-convert -T scss bulma.sass` (or whatever file you want), and it will print out the result.

oleksiipotiekhin:
If you ask me, I prefer to use `SCSS` because of the `C#` background and because i think `SCSS` is more readable than `SASS`. But in the same time the `SASS` is not that different. @leandroluk you have a chance to learn not only _Bulma_ but `SASS` in the same time.

leandroluk:
not really, I use SASS because I need to put this inside a project with the webpack and unfortunately I can not find any plugin that allows me to read SASS within SCSS.

I'm sure that to implement this in the original bulma project is not at all difficult, it is only necessary to add 1 new step in the deploy process to execute the command that the @feute showed

oleksiipotiekhin:
@leandroluk I'm using **Angular CLI** with `SCSS` most of the time and it's using **webpack** behind the scene and I do not have any problem with importing Bulma.

leandroluk:
Hello, I created a fork of the master project and inserted the implementations necessary for the project to automatically create the SCSS files when the project build is done. I hope it's accepted! 

[...]

PR #1506, opened 2017-12-06
Increment bulma package to auto-generate SCSS files

leandroluk:
This is a improvement.

Hello, I ended up cloning your project and made a minimal implementation so that the project construction already automatically provides the files in both SASS and SCSS, so you do not have to worry about doing them and I guarantee it will be good for the framework itself!

The modifications are only the inclusion of the `sass-convert` package as a development dependency and also a new execution script that already performs the conversion of all SASS files to SCSS. This modification does not change the rest of the framework structure. In addition, only adjustments to the package.json file were made so that they also created the SCSS files.


feute:
I think it is somehow redundant to have both sass and scss files in the same repository, because you have to synchronise them by either doing it manually (modifying the scss files according to the sass equivalents) or by executing the script; this would add an extra layer of work for future commits (you have to synchronise them every time you modify sass files).

It would be better to:
- Provide a script to convert sass to scss in `package.json` file.
- Place the converted files to a directory called `scss/`.
- Place `scss/` directory into `.gitignore`.

This way you build the scss files if you need them, or somehow trigger the scss building at installation.

leandroluk:
My fork causes you to not have to mess with anything in the SCSS files, let alone re-write the SASS, instead, it does the direct conversion (via script) in the package.json as soon as the build script is used.

So you do not have to change the original package itself if you add the dependencies I've added and the additional scripts. On my fork, I only have the scss folder because if you deploy the project to the npm package, you do not need to compile it, so it gets more functional.

oleksiipotiekhin:
My opinion is that you should ignore [package-lock.json](https://docs.npmjs.com/files/package-lock.json). Also you shouldn't include `css` files in PR.

jgthms:
I'm not convinced about the need for this PR. 🤔 

You're installing `sass-convert` **globally** on each install??

```
"preinstall": "npm i -g sass-convert replace-in-file rimraf"
```


What I fear is that this can lead to a lot of confusion to users. I can see the questions coming:

* "Why are there both scss and sass files?"
* "Why is the change I made in a scss file not picked up?"
* "Why is the repo double in size now?"
* "I wanted to see the source code for the columns, but I'm not sure where to look?"
* "Why does it take 20 seconds now to install Bulma?"

This is the first time in **2 years** someone specifically requests the files to be in `.scss`, although you can simply import `.sass` files since the Sass engine doesn't care and treats both equally.

Why not create a SCSS fork that watches the Bulma repo and creates a `.scss` version automatically, kind of like the script you implemented here? Because this conversion is now a burden to about 80,000 Bulma users just because you want this change and I don't want to inflict that on them, or at least, not like this.

leandroluk:
First of all thank you very much for the attention. 

Regarding the installation of `sass-convert` globally, I apologize for this, I was doing a test on my computer because I developed it at home and at work and put the reinstallation of the package because I was presenting an error at home. Then I remembered that sass-convert is already installed when I install the sass through ruby (`gem install sass`) and my home computer had been formatted in a short time. I already removed this installation from the `preinstall` script. 

On the other dependencies, they are important because in some cases of versions of Windows 7 I already had problems of global reference but, if it does not find interesting we can remove them.

About the questions that you have presented, can I justify them:

- "Why are there both scss and sass files?"
> Even if they are the same language (SASS or SCSS), there are several npm packages that do not fully recognize the 2 formats (eg https://github.com/webpack-contrib/sass-loader) and because of this it would facilitate the existence of two types, even if the main one is still the SASS files as in your original project_.
 - "Why is the change I made in a scss file not picked up?"
> My solution is not about the SCSS files but a new type of output for the SASS files. so the SCSS files would be just the result of the final build, just like the CSS files. Any change would continue as per the original design.
 - "Why is the repo double in size now?"
> The size of the repository does not directly interfere over the actual size of the solution since they are only script files. One proof of this is that if you download the original project master file and my project containing the SCSS files, the two have the same size (11.7MB).
 - "I wanted to see the source code for the columns, but I'm not sure where to look?"
> As the SCSS directory is a mirror of SASS, in both directories it is possible to have the same structures but in the 2 language formats. That is, the project would be accepted by the entire community working with SASS regardless of language format (SASS / SCSS).
 - "Why does it take 20 seconds now to install Bulma?"
> As with the double repository size issue, whether or not SCSS files would not affect large-scale project installation performance because they are just script files. The only thing that can affect is in the final compilation of the project because the creation of them only occurs when executing the build script or directly calling the build-scss script. For the end user, there are no apparent changes.

I know the project is not mine but I have the same as a great development project. I still think of creating my own framework as you did but unfortunately i do not have time for that at the moment, so i try to pass on my best ideas to the other frameworks i like, as i already did with the bootstrap (when i suggested using `jquery-slim` and to create a file containing also the dependencies like the popper.js) and with the materializecss (when I began an effort of removing of the jquery and completing the functions in pure javascript).

leandroluk:
@feute i'm sorry if I changed the indentation from 2 to 4 spaces, I'm used to always working with 4 and my editor is already configured for it. In my modifications I will then be leaving in 2 spaces ok?

grofit:
This is related to #1502 and really the solution seems to be just convert all sass to scss as a one time thing and replace the sass folders and files with scss files. The sass site even lists scss as its main examples.

`In version 3 of Sass, the SCSS (Sassy CSS) syntax was introduced as "the new main syntax" for Sass `

So can we not just remove this auto generator and instead just remove sass files and replace them with the newer scss syntax ones?

PR #1510, opened 2017-12-07
Implement colours for dropdown component

feute:
This is a **new feature/improvement**.

### Proposed solution

This PR adds the colours implementation to the dropdown component in a simple way: just adding the desired colour to the `dropdown` element as if it were a button -like `is-primary`-, and it will add the colour to both the dropdown button and the active item inside the dropdown menu.

### Tradeoffs

Maybe something with the usage of `@extend` for the button, but for now I think there are no tradeoffs.

### Testing Done

Built with `npm run build` and `npm run docs`, and looked at the local documentation (with `jekyll`) to see the results.

2017-12-07-160044_1262x499_scrot

### DIFF
--- a/sass/components/dropdown.sass
+++ b/sass/components/dropdown.sass
@@ -31,6 +31,18 @@ $dropdown-divider-background-color: $border !default
       padding-bottom: $dropdown-content-offset
       padding-top: unset
       top: auto
+  // Colors
+  @each $name, $pair in $colors
+    $color: nth($pair, 1)
+    $color-invert: nth($pair, 2)
+    &.is-#{$name}
+      .dropdown-trigger
+        .button
+          @extend .is-#{$name}
+      a.dropdown-item
+        &.is-active
+          background-color: $color
+          color: $color-invert
 
 .dropdown-menu
   display: none
--- a/docs/documentation/components/dropdown.html
+++ b/docs/documentation/components/dropdown.html
@@ -38,6 +38,152 @@ doc-subtab: dropdown
 </div>
 {% endcapture %}
 
+{% capture dropdown_colors_example %}
+<div class="dropdown is-active is-primary">
+  <div class="dropdown-trigger">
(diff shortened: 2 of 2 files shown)


sass/components/dropdown.sass
```diff
@@ -31,6 +31,18 @@ $dropdown-divider-background-color: $border !default
...
       padding-top: unset
       top: auto
+  // Colors
+  @each $name, $pair in $colors
+    $color: nth($pair, 1)
+    $color-invert: nth($pair, 2)
+    &.is-#{$name}
+      .dropdown-trigger
+        .button
+          @extend .is-#{$name}
```
jgthms: What is this extending??

feute: It extends every color class (e.g. `is-primary`), but I just saw that it actually extends those classes of *every* element/component, so it leads to unclear CSS output. It's tricky to implement this part. It needs work, I'll be thinking in a better approach to this.

feute:
Currently, the use of `@extend` works, but it adds so much unnecessary code because it extends every element/component that has the `.is-#{$name}` colour class.

It works better with `@extend .button.is-#{$name}`, but it throws a [deprecation warning](https://github.com/sass/sass/issues/1599) so we shouldn't rely on it.

There could be two options, if there are no other way to make this in a simple manner (without having to deal with the `.button` class):
- Use the previous extend with the compound class selector and ignore the deprecation warning.
- Let the colour class to change only the active dropdown item (`a.dropdown-item.is-active`) and let the user change the colour/style of the button inside `dropdown-trigger`. 

I believe the second one could work better because it would let the user customise the trigger button and the dropdown menu separately, and they *are* actually separate elements, so this makes it easy if the user wants to have, for instance, an inversed and outlined button with a specific colour for the active menu item.

Having this said, I can add commits to reflect the changes that you consider better for this topic, @jgthms.

PR #1517, opened 2017-12-08
Fixes #1501 (.box .is-small .is-rounded)

### LINKED ISSUE #1501: .box .is-small .is-rounded
Is it a feature.

### Overview of the problem

This is about the Bulma **CSS framework**.
I'm using Bulma **version** [0.6.1].
My **browser** is: Google Chrome.
I am sure this issue is **not a duplicate**.

### Description
I've faced the situation that there is no possibility to create boxes like you can see on the example  _(**ABOUT**, **SKILLS**)_ . I was trying to achieve it with disabled rounded button, but it didn't work. By adding this code there will be such possibility. @jgthms should I create a PR or it's not worth it?

Example

### Code
Code

### Result
Box

oleksiipotiekhin:
This is a **improvement**.

### Tradeoffs
Can't think of any. Documentation should be updated.

### Testing Done
Used those classes in a project.

### Demo
Box


sass/elements/box.sass
```diff
@@ -15,7 +16,11 @@ $box-link-active-shadow: inset 0 1px 2px rgba($black, 0.2), 0 0 0 1px $link !def
   color: $box-color
   display: block
   padding: $box-padding
-
+  &.is-small
+    padding: $box-padding-small
+  &.is-rounded
+    border-radius: $radius-rounded
+
```
feute: You should remove the trailing whitespaces here so the project is more consistent following the *Bulma Sass styleguide*.

PR #1539, opened 2017-12-18
Update navbar.html

dyldawg:
Fixed the hamburger showing up incorrectly because it was using a <button> and not using the "burger" class

This is an improvement.

### DIFF
--- a/docs/documentation/components/navbar.html
+++ b/docs/documentation/components/navbar.html
@@ -20,11 +20,11 @@ doc-subtab: navbar
 {% endcapture %}
 
 {% capture navbar_burger_example %}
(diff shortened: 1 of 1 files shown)


feute:
The `burger` class is irrelevant in the example, it is not necessary. If you try it, it will work perfectly using a `div` with the `navbar-burger` class only:
```html
<div class="navbar-burger">
   <span></span>
   <span></span>
   <span></span>
</div>
```

This is because there is no `burger` class in bulma.css:
```
$ git grep "\.burger" css/bulma.css
(no output)
```

farens:
This is the same as https://github.com/jgthms/bulma/pull/1432

PR #1581, opened 2018-01-01
Fixes #1580: Remove border radius from 'navbar-burger'

### LINKED ISSUE #1580: Buttons with 'navbar-burger' have border radius.
This is about **Bulma**.

### Overview of the problem
Button elements with the class combination of 'button' and 'navbar-burger' have border radius.

### Description
This was discussed in #1515, but their workaround was to remove the 'button' class and use a &lt;div&gt; tag. This workaround is not sufficient for all cases, because you lose access to modifiers like 'is-primary'.

A solution is to remove the border radius from 'navbar-burger', so it can be safely used in combination with 'button'.

### Expected behavior
image1

### Actual behavior
image2

ek1506:
This is a **bugfix** for #1580.

### Proposed solution
Currently, button elements with the class combination of 'button' and 'navbar-burger' have border radius. A solution is to remove the border-radius from 'navbar-burger', so it can be safely used in combination with 'button'.

### DIFF
--- a/sass/components/navbar.sass
+++ b/sass/components/navbar.sass
@@ -131,6 +131,7 @@ html.has-navbar-fixed-bottom
 .navbar-burger
   +hamburger($navbar-height)
   margin-left: auto
+  border-radius: 0px
 
 .navbar-menu
   display: none


ek1506:
Closed the PR because there was an error with the Travis CI, I believe unrelated to my repo:

wdefwdefwdeffdfddf

Once I see the Travis CI is working again for others, I will probably resubmit the PR.

sass/components/navbar.sass
```diff
@@ -131,6 +131,7 @@ html.has-navbar-fixed-bottom
 .navbar-burger
   +hamburger($navbar-height)
   margin-left: auto
+  border-radius: 0px
```
robbinfellow: x

robbinfellow: I would change `this` to:

`that`

PR #1592, opened 2018-01-08
# fixes issue 1583-button look like text link

nitin992503:
added new class behave_like_text_link:-
changed background color of default HTML button, changed border style to none, and text color to blue

This is a new feature

### Testing Done
yes ,it removed the default groove style, changed the default background color to white and the text color to default color of a link i.e blue.

### DIFF
--- a/css/bulma.css
+++ b/css/bulma.css
@@ -1930,11 +1930,7 @@ a.box:active {
   border-color: transparent;
   color: #fff;
 }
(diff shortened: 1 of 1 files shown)


jgthms:
You have to update the `.sass` files, not the `.css` one.

Also, it'd be great to have some documentation for it as well. And I think what you actually want is not a white background but a transparent one.

PR #1607, opened 2018-01-17
Update dropdown example

sbromley07:

### Proposed solution
Update the class for the icon to be fa instead of fas


jgthms:
The website is using Font Awesome 5 which requiems fas.

PR #1617, opened 2018-01-20
Add Gutter Class Modifiers (Margins & Paddings)

daveberning:
This is a **new feature**.

This new feature adds the `.has-(margin | padding)-left-right`, `.has-(margin | padding)-top-bottom`, `.has-(margin | padding)-sm` classes. CHANGELOG.md is updated.

### Proposed solution
This solution offers an easy way to space any element or add gaps (gutters) to a row or column (Ex: `.container` | `.container-fluid`). The `$gap` variable is used for consistent spacing.

### Tradeoffs
May be considered a long class name. Users could put `.has-magin-right-left` _instead_ of `.has-margin-left-right`.

### Testing Done
I have tested these classes on a local project with Bulma as a dependency.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,11 +3,10 @@
 ## 0.6.4
 
 ### New features
-* Add margin and padding (gutter) modifiers.
+* Add margin and padding (gap) modifiers.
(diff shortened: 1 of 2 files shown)


yusufbicer:
i agree with Jeremy, avoiding to bloat Bulma must be the priority.

PR #1619, opened 2018-01-21
add variables for background color of card header, content and footer

doc75:
This is an **improvement**.

I added 3 new variables to allow customization of the card background for:

- card-header
- card-content
- card-footer

### Proposed solution

This PR allow to set the background differently for each part of a card:

- **card-header** with `card-header-background-color`
- **card-content** with `card-content-background-color`
- **card-footer** with `card-footer-background-color`

### Tradeoffs

None, the default behavior is unchanged (all 3 color are set to `none` by default)

### Testing Done

I have tested it on a personal website.
I checked all 3 cases with custom colors


oppenheimer-:
Don't want to crash the party, but "background-color: none" ist not valid.
Please change those variables to _transparent_ or _unset_ (css 3), @doc75

doc75:
@oppenheimer- good remark.
PR #1838 submitted to fix this.

Issue #1376, opened 2017-10-27
Make card-footer-item a stand-alone element


nedjo:
This is about Bulma.
This is a feature request.

### Overview of the problem

This is about the Bulma **CSS framework
I'm using Bulma **version** 0.6.0
I am sure this issue is **not a duplicate**

### Description
A `card` can have a `card-footer` with one or more `card-footer-item`s. The documentation for this element:

> * `card-footer`: a horizontal list of controls
>     * `card-footer-item`: a repeatable list item

This is indeed a useful element to be able to put in the footer of a `card`. However:

* A horizontal list of controls may be useful elsewhere.
* Other elements may be appropriate in a `card` footer.

### Proposed resolution

* Pull `card-footer-item` and its wrapper into a new element. Proposed name: `controls` (wrapper), with one or more `control` elements. This would be parallel to `tags` and `tag`.
* As needed, generecize code to allow various elements to be placed in the `card-footer` region.

Issue #948, opened 2017-07-29
Suggestion: .card-footer-item modifier


bjasonr:
Is it about Bulma or about the Docs?
Bulma
Is it a bug/feature/question or do you need help?
feature

### Overview of the problem

This is about the Bulma **CSS framework**

I'm using Bulma **version** [0.4.4]

### Description
`.is-primary` and friends are not applicable to the `.card-footer-item`.
using `.button.is-primary` as a child element does not give full width and height to the footer item.
![image](https://user-images.githubusercontent.com/20331749/28741633-cefbd4d2-7444-11e7-8882-cf9ac151d162.png)
(left: `.button.is-primary` inside `.card-footer-item`, right: what I want to achieve without additional styling)

Suggestion:
Styling the `.card-footer-item` using the existing modifiers would be nice.
`<a class="card-footer-item is-primary">View</a>`

posva:
I was wondering what was the reason to make `is-*` classes only work with specific elements. Is it to make sure people don't use it in wrong places where it could look horrible? 😆

miklb:
> I was wondering what was the reason to make is-* classes only work with specific elements

I was just looking to see if those could be applied  the footer. 

`<footer class="footer is-dark">` seems would make perfect sense.

miklb:
Looks like #849 extends the `is-*` to the `navbar` might look at extending in a separate PR.

vintprox:
oof, have to use button right now

Issue #673, opened 2017-04-13
Button in card footer


KKSzymanowski:
### Overview of the problem

This is about the Bulma **CSS framework**
I'm using Bulma **version** 0.3.2
My **browser** is: Google Chrome 57.0.2987.133
I am **pretty** sure this issue is **not a duplicate**.

### Description

I want to create a login form in a card component and everything is fine if my footer is like this:
```html
<footer class="card-footer">
    <a class="card-footer-item" @click="login">Zaloguj</a>
</footer>
```
![image](https://cloud.githubusercontent.com/assets/13980973/25000524/ed4b6ada-2041-11e7-868f-803d608c0f4b.png)
However users could not login by hitting `enter` in any of the text inputs, so I thought I'll wrap the form in a `<form @submit.prevent="login">` tag(I'm using VueJs). This changed nothing because the form did not have a `submit` button.
If I change the link in the footer to a submit button it has the default browser look:
```html
<footer class="card-footer">
    <button type="submit" class="card-footer-item">Zaloguj</button>
</footer>
```
![image](https://cloud.githubusercontent.com/assets/13980973/25000624/5230c67a-2042-11e7-9c4d-4c7a0047640a.png)

### Steps to Reproduce

[See JSFiddle](https://jsfiddle.net/7rxyw6nj/)

### Expected behavior

Button still looks like in the first screenshot.

### Actual behavior

Button doesn't look like in the first screenshot at all.

### Possible fix

If I reset the button styling:
```scss
button.card-footer-item {
  background-color: transparent;
  cursor: pointer;
  color: $link;
  border: 0;
  font-size: 1rem;
  line-height: 1.5;
}
```
it looks exactly like in the first screenshot.

jgthms:
You can include the `appearance: none` styles: https://github.com/jgthms/bulma/blob/master/sass/utilities/controls.sass#L8

KKSzymanowski:
I'm sorry, but I'm not sure I understand.
If I add `style="-webkit-appearance: none"` to the button, it looks like this:
![image](https://cloud.githubusercontent.com/assets/13980973/25001693/f891a260-2046-11e7-8a28-7e1231dd2d8a.png)
If I include the `control` mixin
```scss
button.card-footer-item {
  @include control();
}
```
it looks like this:
![image](https://cloud.githubusercontent.com/assets/13980973/25001783/5e483d58-2047-11e7-9bf4-4c74bff1e929.png)

Could you please elaborate on how should I proceed?

[...]

PR #1637, opened 2018-01-30
WIP: Add footer buttons to .card

timacdonald:
This is a **new feature**. Here's what we are doing:

screen shot 2018-01-30 at 7 21 01 pm

Currently the `.card` footer only allows `.card-footer-item`. I always want to be able to use `.button` elements in the footer instead of just plain links. This PR is a **WIP** and is more of just a suggestions of a possibility. I'm not a CSS ninja - so there could be things wrong here.

Suggestions and feedback welcomed.

### Proposed solution

Here is what I've done to achieve the current SASS...

Duplicated `.button` and then...

- Removed `+control`.
- Removed `+unselectable` (although perhaps this should be retained).
- Removed `border:` stuff.
- Removed `box-shadow:` stuff.
- Removed `.is-outlined` stuff.
- Removed **sizes**.
- Removed `.is-fullwidth`.
- Removed `.is-rounded`.
- Adjusted to match `.card-footer-item`.

### Tradeoffs

This would create a kind of `.card-footer-*` pattern which is perhaps not the best solution. Perhaps a more element agnostic solution would be better for the `.card-footer`. Going forward with this pattern would mean a (possibly substantial) increase in CSS with each new element that is added to the pattern.

@nedjo suggested this here: https://github.com/jgthms/bulma/issues/1376

### Testing Done

I've created [this codepen](https://codepen.io/timacdonald/pen/bLGXxd) with the current built css with a few examples to look at.

## Need input from Jeremy

- [ ] Can we use the button variables so we don't double up on existing rules?
- [ ] The `.card-footer-item` right border when mixing. Check out the blue, white, yellow example. I feel the white buttons right border looks weird?
- [ ] Codepen created with all possible variations to check for possible issues.
- [ ] Individual `card-footer-*` items or a more element agnostic approach.
- [ ] Ensure all existing button variations work, such as loading icon, disabled etc.

### Related

https://github.com/jgthms/bulma/issues/948
https://github.com/jgthms/bulma/issues/673

### DIFF
--- a/sass/components/card.sass
+++ b/sass/components/card.sass
@@ -65,3 +65,105 @@ $card-footer-border-top: 1px solid $border !default
 .card
   .media:not(:last-child)
     margin-bottom: 0.75rem
+
+.card-footer-button
+  align-items: center
+  background-color: $button-background-color
+  color: $button-color
+  cursor: pointer
+  display: flex
+  flex-basis: 0
+  flex-grow: 1
+  flex-shrink: 0
+  justify-content: center
+  padding: 0.75rem
+  strong
+    color: inherit
+  .icon
+    &,
+    &.is-small,
+    &.is-medium,
+    &.is-large
+      height: 1.5em
+      width: 1.5em
+    &:first-child:not(:last-child)
+      margin-left: calc(-0.375em - 1px)
+      margin-right: 0.1875em
+    &:last-child:not(:first-child)
+      margin-left: 0.1875em
+      margin-right: calc(-0.375em - 1px)
+    &:first-child:last-child
+      margin-left: calc(-0.375em - 1px)
+      margin-right: calc(-0.375em - 1px)
+  // States
+  &:hover,
+  &.is-hovered
+    color: $button-hover-color
+  &:focus,
+  &.is-focused
+    color: $button-focus-color
+  &:active,
+  &.is-active
+    color: $button-active-color
+  // Colors
+  &.is-text
+    background-color: transparent
+    color: $button-text-color
+    text-decoration: underline
+    &:hover,
+    &.is-hovered,
+    &:focus,
+    &.is-focused
+      background-color: $button-text-hover-background-color
+      color: $button-text-hover-color
+    &:active,
+    &.is-active
+      background-color: darken($button-text-hover-background-color, 5%)
+      color: $button-text-hover-color
+    &[disabled]
+      background-color: transparent
+  @each $name, $pair in $colors
+    $color: nth($pair, 1)
+    $color-invert: nth($pair, 2)
+    &.is-#{$name}
+      background-color: $color
+      color: $color-invert
+      &:hover,
+      &.is-hovered
+        background-color: darken($color, 2.5%)
+        color: $color-invert
+      &:focus,
+      &.is-focused
+        color: $color-invert
+      &:active,
+      &.is-active
+        background-color: darken($color, 5%)
+        color: $color-invert
+      &[disabled]
+        background-color: $color
+      &.is-inverted
+        background-color: $color-invert
+        color: $color
+        &:hover
+          background-color: darken($color-invert, 5%)
+        &[disabled]
+          background-color: $color-invert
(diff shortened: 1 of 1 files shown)


HDVinnie:
Any update? Will this be merged?

timacdonald:
Nothing from yet - but I think he has been super busy with his new book and the website overhaul. Fingers crossed he might get some spare time soon to have a look at this and let us know his thoughts on this idea.

This will still need a bit of love once he checks it out - so if you wanted to implement it sooner you could checkout [the codepen](https://codepen.io/timacdonald/pen/bLGXxd) example I did up and use the CSS in there. Everything you need is in there to make this work in a current project as it is in this PR.

Or if you are already using a build pipeline for you CSS you could probably just add [the changes](https://github.com/jgthms/bulma/pull/1637/files) to a LESS file.

BenBao93:
This is a very good feature. Found this while trying to search for a solution to do exactly this. Hopefully we can get it merged soon!

jthorpe-marco:
I found one issue when using this fix, but I'm not sure if it's specific to my project.
I had to remove the left and top portions to get my loading icon to be within the button:
```
.card-footer-button.is-loading:after {
    position: absolute;
    /* left: calc(50% - (1em / 2)); */
    /* top: calc(50% - (1em / 2)); */
    position: absolute !important;
}
```

timacdonald:
Thanks for posting @jthorpe-marco This is not a fully complete version, posted here to get a discussion around the idea going. Will definitely need a bit of work and one of the todo items is to create a version of all the button variations already possible, which will include the loading icons. But I didn't wanna put in all that work if it isn't wanted in framework.

If there are any other issues please do post them as it will make it easier to ensure we cover all the bases if it does get merged.

If this idea is wanted I'll happily do more with it. Wait and see I guess :)

moonshine31:
> If this idea is wanted I'll happily do more with it.

I think it's a great idea! But it's been now almost one year since you introduced it.
Regarding if it should be in the framework or not: IMO, definitely - why should a card only have links in its footer. 

Is there a way to push that feature?

maximebloch:
This also applies when you want to disable the footer item. The disabled property is only available for the <button> element, but I want the proper styling of the <a> element in the card-footer-item. 
Where does this issue stand?

timacdonald:
Hey @maximebloch yea that is another thing to add for sure. I'd be happy to add that functionality to this PR. The PR is just waiting for review - I'm happy to push it forward once it has been reviewed.

I've added the disabled state to the original comment todo list.

LastxTemplar:
is this going to be pulled anytime soon?

jgthms:
Hey, thanks for creating this PR. It seems this feature has quite a bit of demand.

The only problem I see is how it's coded. Since you've copied `.button`, there's a _lot_ of duplication. Since these card buttons look and act like regular buttons, but with only a few minor differences, why not use `.button` directly?

Instead of copying the whole thing, just create a new container for these buttons, and apply the differences you want to see.

```html
<div class="card-buttons">
  <button class="button is-info">Cloud</button>
  <button class="button is-static">Or</button>
  <button class="button is-warning">Sun</button>
</div>
```

And then just apply custom styles in this **context**:

```sass
.card-buttons
  .button
    // Custom styles
```

There would be a dependency with `button.sass` but that's fine.

AhmedMKamal:
Just ended up here after some googling, so, if anyone else still struggling with this, here's how I use buttons in card footer.

```html
<footer className="card-footer">
  <p className="card-footer-item is-paddingless">
    <button className="button is-fullwidth">Save</button>
  </p>
  <p className="card-footer-item is-paddingless">
    <button className="button is-fullwidth">Reset</button>
  </p>
</footer>
```

it looks like this

Untitled

timacdonald:
Hey, I don't think I'll be able to get this one off the ground and running. If someone else wants to fork my branch, make the changes suggested by @jgthms  and resubmit it as a PR to benefit Bulma, I'm sure it would be appreciated.

PR #1641, opened 2018-02-04
Add word-break

ckcks12:
This is a **improvement**.

Text is being overflowed from a `.panel-box` if it is long enough. So this `word-break` style is needed and especially `break-word` is the best I think.

#### Before
before
#### Improved
improved

As you can see, long file name _(left panel box)_ is overflowed but now, in the improved version, it's pretty good and has no side effect on well-broken words _(right panel box)_.

### Tradeoffs

Not at all.

### Testing Done

I built it and applied to my private project. Of curse no error found.

### DIFF
--- a/sass/components/panel.sass
+++ b/sass/components/panel.sass
@@ -72,6 +72,7 @@ $panel-icon-color: $text-light !default
   display: flex
   justify-content: flex-start
   padding: 0.5em 0.75em
+  word-break: break-word
   input[type="checkbox"]
     margin-right: 0.75em
   & > .control


HDVinnie:
Any update? Will this be merged?

cdanis:
Thank you, this patch was helpful to me.

ckcks12:
@jgthms @thiamsantos 

Hi. 

I know It's been 3yrs already and I don't want to interrupt you with this tiny pull request but there are still people struggling with this. 

Hope you have time for checking this out. Im not saying about merge, it's okay to close but at least respond to this issue.

Thanks for your hard work and this awesome project 😁.

PR #1680, opened 2018-02-21
No browser's validity message for .file component

### LINKED ISSUE #1664: Browser's validity message is not visible for `file` input.
This is about **Bulma**.

### Overview of the problem

I'm using Bulma **version** [0.6.2]
My **browser** is: `Chromium Version 62.0.3202.94 (Official Build) Built on Ubuntu , running on Ubuntu 17.04 (64-bit)`

### Description

I'd expect that for file input we will be able to see browser validity error, similar to what we can see in plain `input` or `select` even if enriched with Bulma's `input` or/and `select` classes. However the validity message from the browser is not visible.

### Steps to Reproduce

```
<div className="file has-name is-small">
        <label className="file-label" htmlFor="file">
          <span className="file-cta">
            <span className="file-icon">
              <i className="fa fa-upload" />
            </span>
            <span className="file-label">Upload file</span>
          </span>
          <span className="file-name">Some file</span>
        </label>
        <input
          id="file"
          className="file-input"
          type="file"
          name="file"
          required
        />
</div>
```

### Expected behavior

The browser's validity message is visible near to the field

### Actual behavior

The browser's validity message is not visible at all

kornicameister:

Fixes #1664 

### Proposed solution
Add note about limitation


kornicameister:
@jgthms - would you mind checking this up ?

docs/documentation/form/file.html
```diff
@@ -527,6 +527,16 @@
 </div>
 {% endcapture %}
 
+{% capture required_input_file %}
+<form>
+  <input
+    type="file"
+    name="file"
+    required
+  />
```
jgthms: You don’t need the closing slash.

docs/documentation/form/file.html
```diff
@@ -527,6 +527,16 @@
 </div>
 {% endcapture %}
 
+{% capture required_input_file %}
+<form>
+  <input
+    type="file"
```
jgthms: You need the class as well I assume.

docs/documentation/form/file.html
```diff
@@ -739,6 +749,23 @@ <h2 class="subtitle">
       </p>
     </div>
 
+    {% include anchor.html name="Limitations" %}
+
+    <div class="message is-warning">
+      <div class="message-body">
+        <p>
+          Please be aware that browser's validity message, for example when value of input is <code>required</code>, will not be visible when using <code>.file</code>
```
jgthms: “the browser”
“when the input is required”
Move the “for example” clause at the end.

docs/documentation/form/file.html
```diff
@@ -739,6 +749,23 @@ <h2 class="subtitle">
...
+          field.
+        </p>
+        <p>
```
jgthms: Remove this second paragraph. The implementation details are not relevant to the user.

docs/documentation/form/file.html
kornicameister: Well, actually I wanted to just present the code with `required` over the `input type=file`. Didn't mean to use bulma specific code in here. It is available above anyway, so just wanted to back up the paragraph above with very short code snippet.

PR #1691, opened 2018-02-26
Update icons.json

diomed:
updated material and font-awesome icons


docs/_data/icons.json
```diff
@@ -1,7 +1,7 @@
 {
   "fontawesome4": "https://maxcdn.bootstrapcdn.com/font-awesome/4.7.0/css/font-awesome.min.css",
-  "fontawesome5": "https://use.fontawesome.com/releases/v5.0.0/js/all.js",
+  "fontawesome5": "https://use.fontawesome.com/releases/v5.0.6/css/all.css",
```
jgthms: You want js, not css.

diomed: why tho? isn't having css better than having js?
and since f-a 4 is in css...

jgthms: Cause it’s used in a script tag. And it’s the recommended way.

PR #1707, opened 2018-03-07
Disable table hover in .content by default [Fixes #1295]

### LINKED ISSUE #1295: Disable hover styling on tables within .content class by default

andrewsosa:
This is a duplicate of PR #1296, which is a **bugfix** to #1295. Same commits, only now I've rebased @admench's changes as requested. 

### Proposed solution

Uses @admench's changes, which removes the table's hover background as the default background on hover. Again, fixes #1295.

### Tradeoffs

The downside is needing to explicitly mark `.content` tables as `.is-hoverable`, but the previous behavior was undocumented and had no way to disable within the framework.

### Testing Done

Used in a project I'm working on, hover effects went away. 

### Docs Update
It looks like the docs changed to uses JSON content files since this change was originally suggested, let me know if I've made a mistake in updating the docs.

### DIFF
--- a/docs/_data/variables/elements/content.json
+++ b/docs/_data/variables/elements/content.json
@@ -61,11 +61,6 @@
       "name": "$content-table-cell-heading-color",
       "value": "$text-strong"
     },
-    "content-table-row-hover-background-color": {
-      "id": "content-table-row-hover-background-color",
-      "name": "$content-table-row-hover-background-color",
-      "value": "$background"
-    },
     "content-table-head-cell-border-width": {
       "id": "content-table-head-cell-border-width",
       "name": "$content-table-head-cell-border-width",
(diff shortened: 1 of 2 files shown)


andrewsosa:
Here's two GIFs comparing the changes:

### Current version
current_version

### Pull request version
pull_request_version

PR #1719, opened 2018-03-11
better visual layout

farshad-sadri:
I suggest use this layout as better shows nesting than the existing one. Hope it helps. :)


jgthms:
Do you have before and after screenshots?

jgthms:
I see. I agree the white columns are a bit confusing since columns in the documentation are always colored.

PR #1726, opened 2018-03-13
fix(documentation): wrong import derived variables

obedm503:

### Proposed solution

remove extension from derived variables import. with the extension it doesn't import correctly and subtle changes happen


jgthms:
Thanks!

PR #1739, opened 2018-03-19
Remove unused code

shirohana:
This is a **internal improvement**.

The code snippet I removed was declared but do not used in anywhere, not documented in the docs, and has a similar code above it,

so I think it should be removed.


shirohana:
Rebase `jgthms:master`

jgthms:
Thanks!

PR #1741, opened 2018-03-20
Fix margin typo for .button in .buttons

mubaidr:
This is a **bugfix**.

### Proposed solution
Fix the tab spacing issue `.button` inside `.buttons` for sass

### Tradeoffs
none

### Testing Done
Yes, `.button` inside `.buttons` container have expected margins now.


jgthms:
It’s not a typo. The margin is for buttons, not button.

PR #1744, opened 2018-03-21
Update table.sass, fix color hover on striped rows [Fix #1710]

### LINKED ISSUE #1710: Variable $table-striped-row-even-hover-background-color  is not accurate in source
This is about **The Docs**.

### Overview of the problem
This is about the Bulma **Docs*

### Description

The variable name for `$table-striped-row-even-hover-background-color` is confusing since it applies to all rows, not just the "even" rows. Looking through the source code, it is confirmed that the selector is working on all table rows and not just even rows.

While this is very low priority issue, it is confusing since there is `$table-row-hover-background-color` that does not trigger if the table is both "hoverable" and "striped". In that case you'd think there is a variable for the hover color, and indeed there is. But the variable name has the word "even" in it which for all other variables only applies for striped tables on even indexed rows.

This involves changing the source variable name, but I marked this issue as affecting the "docs" due to it being so simple and initially confusing in the documentation. It seems the variable should just remove the "even" in it and everything would make sense.

EEParker:
This is a bugfix for #1710.

### Proposed solution
This is a very simple change to fix hover color on striped tables.

### Tradeoffs
N/A

### Testing Done
Result in screenshot:
image
image

### DIFF
--- a/sass/elements/table.sass
+++ b/sass/elements/table.sass
@@ -98,7 +98,9 @@ $table-striped-row-even-hover-background-color: $white-ter !default
       tbody
         tr:not(.is-selected)
           &:hover
-            background-color: $table-striped-row-even-hover-background-color
+            background-color: $table-row-hover-background-color
+            &:nth-child(even)
+              background-color: $table-striped-row-even-hover-background-color
   &.is-narrow
     td,
     th


jgthms:
I'm not sure what this is fixing? 🤔 

Do you have a Codepen with before and after?

PR #1747, opened 2018-03-22
Makes adding colors easier

Come2Daddy:
This is a **improvement**.
As bulma colors are kind hard-coded into a map, which wrap colors and functions (findColorInvert), to extend them you have to import variables and functions then customize your style, then import derived variables and reassing colors to a merged map. Damn this is heavy. I've come to solution where you set you customizations to $custom-colors variable, the way bulma does, and then you import bulma, done.

### Proposed solution
Add a function in `utilities/functions.sass` that handle `$custom-colors` to merge it with bulma's `$colors` droping malformated elements. Assing it to `$colors` in `utilities/derived-variables.sass` so it can be overriden and is backward compatible with 'old' integrations.

### Tradeoffs
Can't see anyone... 1 call over whole project, with relatively small maps

### Testing Done
Customized through `$custom-colors` with malformated elements, then compiled and used in html page. All custom (proper) and base colors where accessible with `.is-#{name}` classes dependent of colors.

### DIFF
--- a/sass/utilities/derived-variables.sass
+++ b/sass/utilities/derived-variables.sass
@@ -75,8 +75,9 @@ $size-medium: $size-5 !default
 $size-large: $size-4 !default
 
 // Lists and maps
+$custom-colors: null !default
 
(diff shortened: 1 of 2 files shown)


jgthms:
Nice! I'll have a look.

jgthms:
Great stuff! I'm gonna add some documentation, and also add a `$custom-shades` variable.

PR #1749, opened 2018-03-23
Fix icons floating out of input area

ReeceM:
Fix the problem of the `control has-icon-*` icons floating out of the input area when a `.help` element `is-pulled-right` of a preceding field.

This is a **imporvement / bugfix**.

There are no similar problems as I don't think anyone has done the layout I used to get the layout error.

### Proposed solution

Add `clear: both` to the `.control` class to account for `help` elements being pulled to the right when there are no labels between individual inputs. This sort of layout may be preferred when a minimal form is used. But when the help text is made visible everything goes out of whack.

#### Examples of the problem

screen shot 2018-03-23 at 21 56 34

#### Fix Applied

Adding `clear: both`

screen shot 2018-03-23 at 21 56 59

### Tradeoffs

There are no draw backs from what I can tell. The function is supported since CSS 2.1.

### Testing Done

The feature is confirmed to be working across the browsers I have access too.
Latest Safari on Mac, Chrome Mobile, Firefox dev. and Firefox 
[Can work across all though...](https://caniuse.com/#search=clear%3A%20both)

I used the default examples from Bulma documentation to do tests on for creating the error and fixing it. Has been built on the latest release and tested in a development app.

### DIFF
--- a/sass/elements/form.sass
+++ b/sass/elements/form.sass
@@ -538,6 +538,7 @@ $help-size: $size-small !default
   font-size: $size-normal
   position: relative
   text-align: left
+  clear: both //fixes the icon floating out of the input when help text is floated right
   // Modifiers
   // DEPRECATED
   &.has-icon


Come2Daddy:
Reproduced with doc example, fixed with @ReeceM bugfix.

ReeceM:
@jgthms Thanks for merging the PR 👍. Only looked today

PR #1758, opened 2018-03-27
Enable varied tile breakpoint, Fixes #1757

### LINKED ISSUE #1757: Question: How to adjust Tile breakpoint?
This is about **Bulma**.

### Overview of the problem

I'm using Bulma **version** 0.6.2
My **browser** is: Chome

### Description
Is there a way to adjust the breakpoint at which the tiles collapse? I'm using a simple three column tile layout (to add containers with equal verticle heights). My issue is that they don't go responsive vertically until mobile and I would prefer that they do this at the tablet breakpoint (possibly desktop).

Is there a modifier for this?

Jared-Dev:
… tiles to be more customizable in terms of breakpoints.

This is an **improvement**.

This adds the ability to use the `.is-mobile`, `.is-tablet`, `.is-desktop`, etc on the `.tile` class to modify the breakpoint at which tiles are responsive.  If no class is added it maintains it's current default breakpoint.

### Proposed solution

I moved the current code that made tiles into a local mixin to prevent repeating any code. I then just used the existing mixins when a `.is-[breakpoint]` class was added alongside a .tile class. Finally, I used the `:not()` to apply the current defaults (this is alongside the `.is-tablet` code).

Fixed #1757 

### Tradeoffs

Slightly larger compiled css size?

### Testing Done

I used the newly generated CSS to confirm tiles (at least when added to the root tile) began to break at desktop.

### DIFF
--- a/docs/documentation/modifiers/typography-helpers.html
+++ b/docs/documentation/modifiers/typography-helpers.html
@@ -113,9 +113,9 @@ breadcrumb:
           <td><code>is-size-1-tablet</code></td>
           {{ unchanged }}
           {{ size1 }}
-          {{ unchanged }}
-          {{ unchanged }}
-          {{ unchanged }}
+          {{ size1 }}
+          {{ size1 }}
+          {{ size1 }}
         </tr>
         <tr>
           <td><code>is-size-1-touch</code></td>


Come2Daddy:
Haven't tested it yet, moving mixin to `utilities/mixins.sass` with specific name like `=tile-responsive` would be nice please.

Man, I had a hard time understanding what you wanted to achieve here, mislead by 'breakpoints' keyword. You add granurality in responsiveness, which is great.

InquisitiveDevelopment:
Sorry about that, it was late, I was tired. Just needed a feature :)

gonzofish:
Is there any update on this? Having the responsive tiles would be a great feature

Dinuz:
@jgthms is this going to be included and merged in the framework? The big issue is the tile size, when set it up, the size sticks until you hit mobile. Some layout clearly will break on tablet especially in portrait mode.
Be able to control the tiles size and layout as it is possible with columns would be excellent.

Looking forward to hear your opinion Jeremy.

PR #1761, opened 2018-03-27
A11y: Feature/enhance docs

Muhnad:

### DIFF
--- a/docs/documentation/components/modal.html
+++ b/docs/documentation/components/modal.html
@@ -80,7 +80,7 @@ doc-subtab: modal
         </li>
       </ul>
       <p>
-        <a class="button is-primary is-large modal-button" data-target="modal">Launch example modal</a>
+        <button class="button is-primary is-large modal-button" data-target="modal" aria-haspopup="true">Launch example modal</button>
       </p>
     </div>
 
@@ -119,7 +119,7 @@ doc-subtab: modal
     <div class="content">
       <p>If you want a more classic modal, with a <strong>head</strong>, a <strong>body</strong> and a <strong>foot</strong>, use the <code>modal-card</code>.</p>
       <p>
-        <a class="button is-primary is-large modal-button" data-target="modal-ter">Launch modal card</a>
+        <button class="button is-primary is-large modal-button" data-target="modal-bis" aria-haspopup="true">Launch image modal</button>
       </p>
     </div>
 
@@ -152,14 +152,15 @@ doc-subtab: modal
           </div>
           <nav class="level is-mobile">
             <div class="level-left">
-              <a class="level-item">
-                <span class="icon is-small"><i class="fas fa-reply"></i></span>
-              </a>
-              <a class="level-item">
-                <span class="icon is-small"><i class="fas fa-retweet"></i></span>
-              </a>
-              <a class="level-item">
-                <span class="icon is-small"><i class="fas fa-heart"></i></span>
+              <a class="level-item" aria-label="retweet">
+                  <span class="icon is-small">
+                    <i class="fas fa-retweet" aria-hidden="true"></i>
+                  </span>
+                </a>
+                <a class="level-item" aria-label="like">
+                  <span class="icon is-small">
+                    <i class="fas fa-heart" aria-hidden="true"></i>
+                </span>
               </a>
             </div>
           </nav>
@@ -167,17 +168,17 @@ doc-subtab: modal
       </article>
     </div>
   </div>
-  <button class="modal-close is-large"></button>
+  <button class="modal-close is-large" aria-label="close"></button>
 </div>
 
 <div id="modal-bis" class="modal">
   <div class="modal-background"></div>
   <div class="modal-content">
     <p class="image is-4by3">
-      <img src="{{site.url}}/images/placeholders/1280x960.png">
+      <img src="{{site.url}}/images/placeholders/1280x960.png" alt="">
     </p>
   </div>
-  <button class="modal-close is-large"></button>
+  <button class="modal-close is-large" aria-label="close"></button>
 </div>
 
 <div id="modal-ter" class="modal">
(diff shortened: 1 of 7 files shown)


Muhnad:
- Enhance some parts of the component to be accessible.


**TODO**
- we have a lot of parts need to refactor because we have a misunderstanding between `<a>` and `<button>` that make our code is not accessible.

docs/documentation/components/modal.html
```diff
@@ -152,15 +183,21 @@ <h3 class="title">Modal card</h3>
           </div>
           <nav class="level is-mobile">
             <div class="level-left">
-              <a class="level-item">
-                <span class="icon is-small"><i class="fas fa-reply"></i></span>
+              <a class="level-item" aria-label="reply">
```
Muhnad: we should replace the <a> with <button> but if we did that now the style will be ugly, so I kept it.

I think maybe we rewrite some of the parts of the docs and this need time and effort or we can just write a comment above inaccessible component and say we should use x instead of y but we used y for styling purpose.

docs/documentation/components/modal.html
```diff
@@ -80,11 +80,19 @@ <h2 class="subtitle">A classic <strong>modal</strong> overlay, in which you can
         </li>
       </ul>
       <p>
-        <a class="button is-primary is-large modal-button" data-target="modal">Launch example modal</a>
+        <button class="button is-primary is-large modal-button" data-target="modal" aria-haspopup="true">Launch example modal</button>
       </p>
     </div>
 
-    {% highlight html %}{{ modal }}{% endhighlight %}
+{% highlight html %}
```
jgthms: You need to keep `{% highlight html %}{{ modal }}{% endhighlight %}` here.

docs/documentation/components/modal.html
```diff
@@ -106,11 +114,21 @@ <h3 class="title">Image modal</h3>
     <div class="content">
       <p>Because a modal can contain <strong>anything you want</strong>, you can very simply use it to build an image gallery for example:</p>
       <p>
-        <a class="button is-primary is-large modal-button" data-target="modal-bis">Launch image modal</a>
+        <button class="button is-primary is-large modal-button" data-target="modal-bis" aria-haspopup="true">Launch image modal</button>
       </p>
     </div>
 
-    {% highlight html %}{{ image_modal }}{% endhighlight %}
+{% highlight html %}
```
jgthms: `{% highlight html %}{{ image_modal }}{% endhighlight %}`

docs/documentation/components/modal.html
```diff
@@ -119,15 +137,28 @@ <h3 class="title">Modal card</h3>
...
+        <button class="button is-primary is-large modal-button" data-target="modal-ter" aria-haspopup="true">Launch modal card</button>
       </p>
     </div>
 
-    <div class="highlight-full">
-      {% highlight html %}{{ modal_card }}{% endhighlight %}
-    </div>
-
-    {% include variables.html type='component' %}
+{% highlight html %}
```
jgthms: Edit the captured snippet, not the whole block here. And leave the variables partial

docs/documentation/components/nav.html
```diff
@@ -53,20 +53,20 @@ <h2 class="subtitle">
 {% capture nav_example %}
 <nav class="nav">
   <div class="nav-left">
-    <a class="nav-item">
-      <img src="{{ site.url }}/images/bulma-logo.png" alt="Bulma logo">
+    <a class="nav-item" href="#">
+      <img src="{{ site.url }}/images/bulma-logo.png" alt="Bulma">
```
jgthms: Keep `alt="Bulma logo"`

docs/documentation/components/nav.html
```diff
@@ -145,9 +145,9 @@ <h3 class="title">Modifiers</h3>
   <div class="container">
     <div class="nav-left">
       <a class="nav-item">
-        <img src="{{ site.url }}/images/bulma-logo.png" alt="Bulma logo">
+        <img src="{{ site.url }}/images/bulma-logo.png" alt="Bulma">
```
jgthms: `alt="Bulma logo"`

docs/documentation/components/nav.html
Muhnad: Why do we need to keep 'logo'?

jgthms: The `alt` attribute is supposed to describe what the image is, not what it says.

Muhnad: and in our case, this is a link so the user needs to hear "Bulma link" not "Bulma logo link"the second one means if the user clicks on it it will open the bulma logo not redirect to the main page.

Come2Daddy: You are mixing `alt` and `title` (I pass on aria attributes), `alt` stands for alertative content that should describe what is shown on the missing picture/media. The anchor should carry information on his path, img is described through his `alt` attribute.

jgthms: Actually the whole nav is deprecated so it doesn't matter.

Also, `fas fa-github` won't work since it needs `fab`.

docs/documentation/components/modal.html
```diff
@@ -119,7 +119,7 @@ <h3 class="title">Modal card</h3>
     <div class="content">
       <p>If you want a more classic modal, with a <strong>head</strong>, a <strong>body</strong> and a <strong>foot</strong>, use the <code>modal-card</code>.</p>
       <p>
-        <a class="button is-primary is-large modal-button" data-target="modal-ter">Launch modal card</a>
+        <button class="button is-primary is-large modal-button" data-target="modal-bis" aria-haspopup="true">Launch image modal</button>
```
jgthms: This is the wrong text and wrong target.

jgthms:
Ok there were still some issue but I merged it and fixed them. Thanks.

PR #1777, opened 2018-04-05
always leads to latest available download

diomed:


jgthms:
Ah there's already `{{ site.data.meta.download }}`. I just forgot those instances.

PR #1785, opened 2018-04-12
Changed "scroll" to "auto" in html overflow-y

humarh-dharnarh:
This fixes the shown scrollbar by hiding it in a page with just "is-fullHeight" hero or 100vh and 100vw.


jgthms:
It's intentional to prevent horizontal jumps when switching pages.

It could be turned into a variable though.

PR #1786, opened 2018-04-13
Add is-small section modifier

aseredenko:
.is-small for section works for me. Why it wasn't listed here?


jgthms:
It's not in the code: https://github.com/jgthms/bulma/blob/master/sass/layout/section.sass

aseredenko:
oh sorry, my bad

frederikhors:
@jgthms why not `is-small` to `section`?

PR #1819, opened 2018-04-23
fix error message for required file fix #1137

### LINKED ISSUE #1137: [Bug] Error message for a required file upload is not shown correctly
### Overview of the problem

This is about the Bulma **CSS framework**, more specifically about the **file upload** input.

I'm using Bulma **version** [0.5.1]
My **browser** is: Chrome / Edge

### Description

Trying to submit an empty required file input will not show the default error message correctly.
fileinputbug

### Steps to Reproduce

1. Add a form with a required file input
```html
<form>
  <div class="file">
    <label class="file-label">
        <input class="file-input" type="file" name="file" required>
        <span class="file-cta">
            <span class="file-icon">
                <i class="fa fa-upload"></i>
            </span>
            <span class="file-label">
                Choose a file…
            </span>
        </span>
    </label>
  </div>
  <button type="submit" class="button is-primary is-outlined">Submit</button>
</form>
```
2. Try submitting the form without uploading a file

### Expected behavior

The browser's default error message is shown

### Actual behavior

Chrome: The browser's default error message is only shown for a very short period
Edge: No error message is shown

kossa:
This is a bugfix

If a file input is required we dont get browser error, for that I change the with and height 

image

### DIFF
--- a/sass/elements/form.sass
+++ b/sass/elements/form.sass
@@ -352,12 +352,12 @@ $help-size: $size-small !default
       border-color: darken($file-name-border-color, 5%)
 
 .file-input
-  height: 0.01em
+  height: 100%
   left: 0
   outline: none
   position: absolute
   top: 0
-  width: 0.01em
+  width: 100%
 
 .file-cta,
 .file-name


jgthms:
It seems to work in Chrome on Mac. How does it behave on Windows?

kossa:
Dont have windows, but it should works ;)

jgthms:
"Should work" is not good enough to be merged 😉 

We have to be careful to not break existing installations. I'll try to have a look on Windows.

kossa:
LOL, alright

I'll check if I can get windows laptop

turbopixel:
I can check it later on Windows 7

PR #1820, opened 2018-04-24
Add an extra prerequisite - ruby2.x-dev in docs/README.md

kudapara:

### Proposed solution
Tell users to have ```ruby2.x-dev``` installed before attempting to build the docs. Installing Jekyll on Linux (Ubuntu 16.04) machine failed because ```ruby 2.x``` was installed and ```ruby2.x-dev``` was not.

### Testing Done
Yes it works :smiley:.


jgthms:
This instruction is only valid for Ubuntu. Can you make a note instead?

PR #1824, opened 2018-04-25
Fixes #1366 - $breadcrumb-item-separator-color not inverted

### LINKED ISSUE #1366: Hero with breadcrumbs: $breadcrumb-item-separator-color not inverted
This is about **Bulma**.
This is about **bug**.
This is not a **browser bug**.

I'm using Bulma **version** [6.0.0]

### Description

The breadcrumb item separator color is not inverted when breadcrumbs are inside a dark hero section.

mercyaj:
This is a bugfix.

Fixes #1366.

### Proposed solution
Fix issue that prevents breadcrumb separator from being inverted based on the background (i.e. breadcrumb that is a child of hero is-dark). 

### Tradeoffs
Removes default breadcrumb separator color of #b5b5b5.

### Testing Done
Tested the breadcrumb component with different parents with classes such as is-primary and is-dark.

### DIFF
--- a/sass/components/breadcrumb.sass
+++ b/sass/components/breadcrumb.sass
@@ -5,8 +5,6 @@ $breadcrumb-item-active-color: $text-strong !default
 $breadcrumb-item-padding-vertical: 0 !default
 $breadcrumb-item-padding-horizontal: 0.75em !default
 
-$breadcrumb-item-separator-color: $grey-light !default
-
 .breadcrumb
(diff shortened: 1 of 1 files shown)


jgthms:
This is a bit of a breaking change since people use `$breadcrumb-item-separator-color` to customize the separator. 

A better approach would be to change the separator color only when used in a hero.

PR #1831, opened 2018-04-27
Improved Card Box-Shadow

ayushsharma82:
Card shadow in bulma is kind of old type. I have updated it with much cleaner shadow.
The improvement creates a beautiful blending shadow which is the current design standard in the industry.


jgthms:
There is some redesign in the pipelines. 😉

Issue #1902, opened 2018-05-30
Modal doesn't center in IE11


eduardoarandah:
Use IE11.48

Reproduce:
https://bulma.io/documentation/components/modal/

Click "launch example modal", it goes to the right

I know Browser Support says IE partially supported, however IE11 is used more than safari (7%)

(Works perfect in Edge)

![bulma](https://user-images.githubusercontent.com/4065733/40740614-e21653b6-640e-11e8-9cc1-e14769ae51aa.png)

patrickabkarian:
https://github.com/jgthms/bulma/pull/1905

PR #1905, opened 2018-05-31
bugfix: issues/1902 fix for ie11

patrickabkarian:
https://github.com/jgthms/bulma/issues/1902

This is a bugfix

### Proposed solution
Fixes modal not being centered on IE11

### Tradeoffs
Drawbacks: none

### Testing Done
Yes. Works in IE11, works in Chrome.

### DIFF
--- a/sass/components/modal.sass
+++ b/sass/components/modal.sass
@@ -32,6 +32,7 @@ $modal-card-body-padding: 20px !default
   +overlay
   align-items: center
   display: none
+  flex-direction: column
   justify-content: center
   overflow: hidden
   position: fixed


jgthms:
How was this tested?

patrickabkarian:
I've tested this in the current project I'm working on by simply just adding the CSS rule. I've also made a few vanilla tests (including on the Bulma homepage)

image

It doesn't really change anything honestly, since this flex container just always contains one relative element according to the Bulma description.

image

scottwernervt:
@jgthms Also verified the fix on a [free Windows 7 IE11 VM](https://developer.microsoft.com/en-us/microsoft-edge/tools/vms/):

ie11vmmodalflexdirectioncolumnfix

PR #1909, opened 2018-06-01
Modal card broken on IE11 fix

patrickabkarian:
This is a bugfix

How it should be:

image

How it is:

image

Adding -ms-overflow-y: visible will only affect IE11.

Tested in IE11 in a real project and the Bulma homepage.


patrickabkarian:
@jgthms any reason this & https://github.com/jgthms/bulma/pull/1908 don't get merged, they are just IE11 bugs with simple fixes

PR #1942, opened 2018-06-14
Make navbar SASS more flexible

lunfel:
This is a ** improvement **.

I am proposing a little loosened up SASS definition for navbar, that will allow Buefy (or any other sub-project) to customize the navbar.

### Proposed solution

Please see the issue on the Buefy project to understand the underlying problem  https://github.com/buefy/buefy/issues/614

The current SASS has rules that applies to direct children class. This makes it impossible for project like Buefy to work because the component aspect of Vue.js will add a <div> in the hierarchy and thus breaking the rule. 

By instead applying the rule to any descendent child of the class, this makes it possible for Buefy to be styled appropriatedly.

### Tradeoffs

The original SASS was the right way to go, but it didn't allow sub-project to adhere to the rules.

### Testing Done

I have edited the SASS and checked that the style is the same when I insert an extra DIV in the DOM hierarchy.

### DIFF
--- a/sass/components/navbar.sass
+++ b/sass/components/navbar.sass
@@ -62,10 +62,10 @@ $navbar-bottom-box-shadow-size: 0 -2px 0 0 !default
       background-color: $color
       color: $color-invert
       .navbar-brand
-        & > .navbar-item,
+        & .navbar-item,
         .navbar-link
           color: $color-invert
-        & > a.navbar-item,
+        & a.navbar-item,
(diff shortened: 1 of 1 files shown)


jgthms:
This will break setups because the direct children selector is meant to avoid targeting navbar items that are in the dropdown.

By the way, `& .navbar-item` doesn't do anything. It's similar to `.navbar-item`.

Issue #1875, opened 2018-05-16
Feature request – "are-medium" for div classes "tags" and "buttons"


lukepighetti:
Feature request.

Instead of this

```html
<div class="tags">
  <span class="tag is-medium"> One </span>
  <span class="tag is-medium"> Two </span>
  <span class="tag is-medium"> Three </span>
  <span class="tag is-medium"> Four </span>
  <span class="tag is-medium"> Five </span>
  <span class="tag is-medium"> Six </span>
</div>
```

Can we have

```html
<div class="tags are-medium">
  <span class="tag"> One </span>
  <span class="tag"> Two </span>
  <span class="tag"> Three </span>
  <span class="tag"> Four </span>
  <span class="tag"> Five </span>
  <span class="tag"> Six </span>
</div>
```

jgthms:
Yes, that's a good idea.

iamtimsmith:
@jgthms I will work on this.

lukepighetti:
yeeeaaaaah!

PR #1953, opened 2018-06-20
Change color of loading bar on inverted button

psingl:
Inverted and loading button i.e. ```<button class="button is-inverted is-loading">Loading</button>``` looks something like 
capture1
But this will change the same button to
image


jgthms:
I think you added a second commit that has nothing to do with this PR.

psingl:
Yeah! That's just by mistake Actually I was trying to resolve the Issue #1875 The second commit partially resolves the Issue #1875 Sorry for the mistake.

PR #1959, opened 2018-06-21
Fixes Issue #1875

iamtimsmith:
Set navbar hamburger color to match items and allow manual setting
---
This is a new feature.

### Proposed solution
This patch is a new feature to accommodate the request in #1875. The issue has a request for sizing classes on the the `.buttons` and `.tags` elements. This PR adds that feature by adding `are-medium` and `are-large` classes to the tag.sass file and `are-small`, `are-medium`, and `are-large` classes to the button.sass file.

### Tradeoffs
Using the sizing classes on the `.tags` or `.buttons` elements removes the ability to size the children individually. For instance, if you used `.are-medium` on the parent element and then tried to give one of the children `.is-small`, it wouldn't apply the `.is-small` styling.

If the user wishes to have differently sized buttons, they will have to add the desired sizing class to each `.button` element. This would be no different than how they currently have to size buttons differently.

### Testing Done
I have tested this feature locally with a basic html document which uses Bulma.


jgthms:
This looks good. Can you remove the generated CSS though?

diomed:
I'm thinking of hijacking this if @iamtimsmith doesn't resolve conflicts and change anything by the end of the weekend.

jgthms:
No problem @diomed

jgthms:
It doesn't seem to have worked 🤔

jgthms:
Closed by #2201

PR #1974, opened 2018-07-03
add .is-spaced for .has-navbar-fixed-top/bottom in navbar

Matsukiyo:
This is a **improvement**.

From the design concept of Bulma, has-navbar-fixed-top/bottom classes would be better to have "is-spaced" class.
Users cannot notice has-spaced-navbar-fixed-top/bottom classes,  because they are not written in docs.

### DIFF
--- a/sass/components/navbar.sass
+++ b/sass/components/navbar.sass
@@ -403,6 +403,10 @@ a.navbar-item,
       padding-top: $navbar-height + ($navbar-padding-vertical * 2)
     &.has-spaced-navbar-fixed-bottom
       padding-bottom: $navbar-height + ($navbar-padding-vertical * 2)
+    &.has-navbar-fixed-top.is-spaced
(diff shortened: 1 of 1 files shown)


sass/components/navbar.sass
```diff
@@ -403,6 +403,10 @@ a.navbar-item,
       padding-top: $navbar-height + ($navbar-padding-vertical * 2)
     &.has-spaced-navbar-fixed-bottom
       padding-bottom: $navbar-height + ($navbar-padding-vertical * 2)
+    &.has-navbar-fixed-top.is-spaced
```
jgthms: What is `is-spaced` supposed to do here?

Matsukiyo: ```is-spaced``` extends padding on the desktop. It prevents screen collapse on the desktop when ```navbar``` has ```navbar-fixed-top / bottom``` and ```is-spaced```.

Issue #1969, opened 2018-06-27
file input displays unwanted "choose file" text with is-right in Chrome


girroir:
This is about **Bulma**.

This is a  bug only present in Chrome

### Overview of the problem
This is about the Bulma **CSS framework**
I'm using Bulma **version** [0.7.1]
My **browser** is: Chrome  67.0.3396.99 (Official Build) (64-bit) for Windows

### Description
Modifier example #2 on https://bulma.io/documentation/form/file/ displays unwanted text "Choose file" in the top left corner of the input box when viewed in Chrome v67. This is not present in Edge or Firefox
### Steps to Reproduce

Load https://bulma.io/documentation/form/file/

### Expected behavior
"Choose file" button is not displayed. If is-right is removed as in Modifier example 1, the text and button are not present.

### Actual behavior
"Choose file" button is displayed.
![image](https://user-images.githubusercontent.com/8406639/41992531-93a5f768-79fd-11e8-9fdd-1728313f5fb4.png)

Issue #1904, opened 2018-05-31
File upload - Original button visible in Chrome latest version


mm98:
This is about **Bulma**.

it about Bulma
Is it a bug
If it's a bug, is it a browser bug? Yes

This is about the Bulma **CSS framework**
I'm using Bulma  **0.7.1**
My **browser** is: Google Chrome v67.0.3396.62 (win10)
I am sure this issue is **not a duplicate**? Yes (checked)

### Steps to Reproduce
It seams like a bug related to the latest Google Chrome. If you look at https://bulma.io/documentation/form/file/ you'll see that the original upload file button is visible ([sass](https://github.com/jgthms/bulma/blob/56d58280caa83689c3ccb8c549be2b36739bda71/sass/elements/form.sass#L354)). 

![Screenshot from latest Google Chrome](http://dev.roob.dk/bulma-file-upload.png)

### Expected behavior
Original form file/upload control should be hidden/invisible.

### Actual behavior
Original form file/upload control is visible.

jgthms:
Ugh it was the bug fix I merged yesterday. I knew something would break.

mm98:
Don't think so .. I haven't updated my local version and this shows the same problem. It works ok in FF and IE.

jgthms:
Are you on Linux? I can't reproduce:

![screen shot 2018-05-31 at 17 38 53](https://user-images.githubusercontent.com/1254808/40795298-88070816-64f9-11e8-955c-474d4e56a725.png)

mm98:
No, this is very odd. This started after i updated to the latest official Google Chrome **v67.0.3396.62** (64bit) without any extensions (released a couple of days ago) on Windows 10. I'll try to do some more debugging.

mm98:
This must be a specific Google Chrome on Windows problem. I just installed Chrome on another machine (clean machine), and the same happens:

<img width="421" alt="screenshot_1" src="https://user-images.githubusercontent.com/992714/40825088-d6b7edcc-6576-11e8-8a03-f7089f448fd1.png">

eugestarr:
Can confirm, using **Chrome** (67.0.3396.87) on **Mac OS X Sierra** (10.12.6):

![screen shot 2018-06-18 at 12 18 26](https://user-images.githubusercontent.com/13213650/41513507-c58ec466-72f1-11e8-98bd-85cf6bdf4bcd.png)

Why not just add `display: none` to `.file-input`?

mm98:
[...]

PR #1976, opened 2018-07-05
Native upload button visible in latest Google Chrome

mm98:
This is a **bugfix**.

### Proposed solution
The recent Google Chrome on Windows and Mac seams to make the native file upload button visible, even though it should be hidden (which it is in Firefox and IE). I've managed to narrow it down to css `height` and `width` has no effect on the button itself, but only the text-field on the right. This solves #1969 and #1904.

### Tradeoffs
The `.file-input` can't be set to `display: none` or `visibility: hidden`, since third-party plugins (like jQuery Validate) can't focus on the element. Twitter Bootstrap uses `opacity: 0`, and this works very well with third-party plugins.

### Testing Done
Has tested on Windows 10 with Google Chrome, Firefox, Internet Explorer, Microsoft Edge and Safari (iOS).

### DIFF
--- a/sass/elements/form.sass
+++ b/sass/elements/form.sass
@@ -352,12 +352,8 @@ $help-size: $size-small !default
       border-color: darken($file-name-border-color, 5%)
 
 .file-input
-  height: 0.01em
-  left: 0
-  outline: none
+  opacity: 0
   position: absolute
-  top: 0
-  width: 0.01em
 
 .file-cta,
 .file-name


kornicameister:
Yeah, that struck my project as well.
@jgthms do you think it'd be doable to pull in this fix as part of `0.7.2`, if that's not out already ?

kornicameister:
FYI: This only occurs in `Chrome <= 69`. I am also running unstable Chrome `>=71` and that issue is not present there.

Issue #1981, opened 2018-07-07
Navbar burger icon does not use the appropriate color when color modifier is used in navbar


corneliusludmann:
This is about **Bulma**. It's a bug in navbar.sass.

### Overview of the problem

This is about the Bulma **CSS framework**

I'm using Bulma **version** 0.7.1 and have tested the latest commit in master (1083f01).

My **browser** is: Google Chrome Version 67.0.3396.99

### Description

When using a color modifier for a navbar (e.g., is-dark) the burger icon does not use the appropriate color. This leads to a burger icon and close icon which are not visible when clicked.

### Steps to Reproduce

You can see the wrong behavior in the following minimal code sample (see also: https://jsfiddle.net/pm26zwer/6/). The navbar burger is blue and should be whitesmoke (inverse). When hover the burger button nearly disappears.

```
<!DOCTYPE html>
<html>

  <head>
    <meta charset="utf-8">
    <meta name="viewport" content="width=device-width, initial-scale=1">
    <title>Hello Bulma!</title>
    <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/bulma/0.7.1/css/bulma.min.css">
    <script defer src="https://use.fontawesome.com/releases/v5.0.7/js/all.js"></script>
  </head>

  <body>
    <nav class="navbar is-dark" role="navigation">
      <div class="container">
        <div class="navbar-brand">
          <a href="/" class="navbar-item">Company</a>

          <a id="navbar-burger-id" role="button" class="navbar-burger is-hoverable">
          <span></span>
          <span></span>
          <span></span>
        </a>
        </div>
      </div>
    </nav>
    <section class="section">
      <div class="container">
        <h1 class="title">
          Hello World
        </h1>
        <p class="subtitle">
          My first website with <strong>Bulma</strong>!
        </p>
      </div>
    </section>
  </body>

</html>
```

### Expected behavior

The navbar burger icon color should be handled as the navbar link color. If you use is-dark for the navbar the navbar burger icon color should be whitesmoke. You can uncomment the CSS in https://jsfiddle.net/pm26zwer/6/ to see how it should look like.

### Actual behavior

The navbar burger color has not the appropriate color when using is-dark or other color modifiers for the navbar.

corneliusludmann:
See pull request #1982 for a proposed solution.

emanuelmutschlechner:
[...]

PR #1982, opened 2018-07-07
Fix navbar-burger color when color modifier is used

corneliusludmann:
When using a color modifier for a navbar (e.g., is-dark) the burger icon does not change the color. This commit adds the color modification for the navbar burger icon.

This is a **bugfix**.
Issue #1981.

### Proposed solution

This solution adds color-invert for the .navbar-burger for the color modifiers the same way it is done for .navbar-link.

### Tradeoffs

No tradeoffs.

### Testing Done

I checkout out master and changed the lines. The color of the burger icon is as desired. You can see a minimal example I used there: https://jsfiddle.net/pm26zwer/6/

See also issue #1981.


sass/components/navbar.sass
```diff
@@ -74,6 +74,8 @@ $navbar-bottom-box-shadow-size: 0 -2px 0 0 !default
         .navbar-link
           &::after
             border-color: $color-invert
+        & > .navbar-burger
```
jgthms: You don't really need the nesting here. Just use `.navbar-burger`.

corneliusludmann: OK, I changed it. Is it that what you meant?

jgthms: Not exactly. Put `.navbar-burger` as a sibling of `.navbar-brand`, not as a child of `.navbar-brand`.

corneliusludmann: Ah, I see: `.navbar-burger` does not necessarily need to be a child of `.navbar-brand`. Makes sense. So, here my next try. Is that correct? Thank's for your patience.

corneliusludmann:
Just want to ask whether it is ready to merge after the latest correction to include it in the next release!? Is there anything I can do to help? Thanks for considering …

PR #1983, opened 2018-07-09
Update to the latest version of Font Awesome

Mortiferr:
Font Awesome recently updated from 5.0.7 to 5.1. I've updated this to reflect that change.


Mortiferr:
I know this is closed, but does anyone know of any official CDN link where it links to the latest? This was available for FA 4.x, but since the bump to 5.x it seems to have been dropped, as they've moved to  a more self-hosted (use.fontawesome.com) solution--probably to better support Pro.

PR #1985, opened 2018-07-10
Fix percentage

elpopisencio:
Not a major fix but the wrong percentage bothered me :)

This is a documentation fix.


jgthms:
Haha nice catch.

PR #2000, opened 2018-07-18
Remove unnecessary use of @extend inside mixins

Log1x:
This is a **improvement/bugfix**. In the current implementation of `mixins.sass`, it is relying heavily on the use of `@extend` which are not being used elsewhere throughout Bulma causing otherwise useful mixins to fail upon compilation when attempting to use them inside of media queries.

### Proposed solution
Move the CSS of these various selectors inside of the mixins themselves. 

### Tradeoffs
None that I can find. Everything compiles as intended and none of these inheritance selectors are being used anywhere else within' Bulma or the docs.

### Testing Done
Everything compiles and builds as intended. I am now able to make use of mixins such as `+arrow()` from within' media queries and the responsive mixin helpers shipped with Bulma.

The only mixins that had to be slightly modified were `+arrow()` with setting `$color` to default to `transparent` and removing the need for `border-color` since we can just set it on our initial `border` rule and `+delete()` with changing its `@extend` of `%unselectable` to use the `+unselectable` mixin instead.

### DIFF
--- a/sass/utilities/mixins.sass
+++ b/sass/utilities/mixins.sass
@@ -130,12 +130,15 @@
 
 // Placeholders
 
-=unselectable
+%unselectable
   -webkit-touch-callout: none
   -webkit-user-select: none
   -moz-user-select: none
   -ms-user-select: none
   user-select: none
+  
+=unselectable
+  @extend %unselectable
 
 =arrow($color: transparent)
   border: 3px solid $color
@@ -157,8 +160,8 @@
   &:not(:last-child)
     margin-bottom: 1.5rem
 
-=delete
-  +unselectable
+%delete
+  @extend %unselectable
   -moz-appearance: none
   -webkit-appearance: none
   background-color: rgba($black, 0.2)
@@ -222,8 +225,11 @@
     min-height: 32px
     min-width: 32px
     width: 32px
+    
+=delete
+  @extend %delete
 
-=loader
+%loader
   animation: spinAround 500ms infinite linear
   border: 2px solid $grey-lighter
   border-radius: $radius-rounded
@@ -234,6 +240,9 @@
   height: 1em
   position: relative
   width: 1em
+
+=loader
+  @extend %loader
 
 =overlay($offset: 0)
   bottom: 0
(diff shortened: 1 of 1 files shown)


jgthms:
The reason behind the `@extends` is to reduce file size and have a more optimized CSS because it combines selectors that share the same set of rules.

Now I understand the idea of using them inside media queries. I can see `+overlay` `+block` `+arrow` used as such. Not sure about `+unselectable` `+delete` and `+loader`.

If you intend to remove the overlay mixin, you need to replace all occurrences of `+overlay` with `@extend %overlay` because it's already being used in many places, and I want to keep the selector grouping.

Log1x:
I kept the `+overlay` mixin, the commit just looks a little weird at a glance with it moving everything around.

I understand your reasoning behind the other uses of `@extend` and will push an update shortly.

Log1x:
I reverted the changes to `+unselectable`, `+delete`, and `+loader`. `+overlay` is intact just to clarify on the comment above.

Everything else should be good and compiles as intended, including inside of media queries.

Log1x:
@jgthms any further observations / requests before I can get this potentially merged? It's causing a slight annoyance in my current project.

jgthms:
Thanks @Log1x. So to keep parity with the current setup, it'd be great to keep the extends when possible.

A clean swap would be better. Instead of having the mixin using the extend, we need the extend to use the mixin:

```sass
// Before
=arrow($color)
  @extend %arrow
  border-color: $color

// After
@extend %arrow
  +arrow
```

For example, there's 2 instances of `+arrow`. We should replace them with `@extend %arrow`. Same for `+block` `+loader` and `+overlay`.

That way, the mixins are still available for anyone to use, but we use `@extend` when possible.

Log1x:
Alright. Swapped out the styling logic from the `@extend` selector to the mixin, and then call the mixin inside of the `@extend` instead. For `+delete`, I kept the usage of `@extend %unselectable`.

Hopefully, I got it right this time.

jgthms:
Yeah that'll work. You forgot to replace the occurrences of each mixin in the codebase but I can handle it. I'll merge it like this. Thanks for your work!

PR #2004, opened 2018-07-18
Fixed docs when Vimeo fails to load

lightswitch05:
All javascript breaks if Vimeo fails to load or is blocked. Checking for `Vimeo` before accessing it should fix it. This can easily be reproduced using [uMatrix](https://github.com/gorhill/uMatrix).

This is a **improvement**.
All JavaScript breaks on the docs if Vimeo fails to load - such as 'add column'. Checking for Vimeo before trying to use it should fix it

### Proposed solution

Check for Vimeo before using it.

### Tradeoffs

Code doesn't look as pretty

### Testing Done

Block Vimeo using [uMatrix](https://github.com/gorhill/uMatrix) browser plugin, but I have not tested the fix. I figured this visual would still be more useful then a bug ticket though.

### DIFF
--- a/docs/_javascript/index.js
+++ b/docs/_javascript/index.js
@@ -4,12 +4,14 @@ document.addEventListener('DOMContentLoaded', () => {
 
   const introVideo = document.getElementById('introVideo');
   const introIframe = document.getElementById('introIframe');
-  const introPlayer = new Vimeo.Player(introIframe);
   const npmClipboard = new Clipboard('#npmCopy');
 
-  introPlayer.ready().then(function() {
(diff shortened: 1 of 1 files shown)


jgthms:
Thanks that's a good fix.

jgthms:
Oh wait, you forgot to compile it. It's alright, I'll do it.

lightswitch05:
@jgthms I'm sorry, I didn't spend too much time looking at the internals. I just learned of the project today, and found this issue immediately

jgthms:
No worries. Thanks for helping. I just need to run `npm run build`.

PR #2039, opened 2018-08-02
Fixed responsive breakpoints

corford:
Documentation suggests breakpoints for desktop, widescreen and fullhd are from: 1024px, 1216px and 1408px respectively. Comments in sass/utilities/initial-variables.sass corroborate this (container base width + 4 rem i.e. 64px). Also, real life agrees too (e.g. desktop width traditionally starts at 1024px). Prior to this commit; desktop, widescreen and fullhd breakpoints were wrongly: 1088px, 1280px and 1472px (because $gap was multiplied twice - a hang over from when $gap was 32px rather than 64px?).

### DIFF
--- a/sass/utilities/initial-variables.sass
+++ b/sass/utilities/initial-variables.sass
@@ -50,12 +50,12 @@ $gap: 64px !default
 // 960, 1152, and 1344 have been chosen because they are divisible by both 12 and 16
 $tablet: 769px !default
 // 960px container + 4rem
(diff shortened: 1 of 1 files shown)


corford:
Hasn't been merged yet (hoping it will at some point).

kaligari:
Other solution is to set **$desktop** variable:
`$desktop: 896px + (2 * $gap)`

PR #2059, opened 2018-08-19
Remove unnecessary right margin from last level-item (level.is-mobile)

ciscorn:
This is a bugfix.

In a mobile level (`level.is-mobile`), the last level-item has unnecessary right-margin:

### Proposed solution

Just wrap with `&:not(:last-child)`

### Tradeoffs

None

### Testing Done

I tested with my webapp. It looks good.


ciscorn:
@jgthms Sorry, I've updated my PR.

PR #2075, opened 2018-08-29
Remove "display: block" property definition

PinkiNice:
Any reasons to have this property in a stylesheet?
This reduces flexibility, as it becomes not that easy to override "display" property.
For example ".tile.flex--row" won't work if ".flex--row" is defined before ".tile" which makes it not that easy to create row tiles

### DIFF
--- a/sass/grid/tiles.sass
+++ b/sass/grid/tiles.sass
@@ -1,6 +1,5 @@
 .tile
   align-items: stretch
-  display: block
   flex-basis: 0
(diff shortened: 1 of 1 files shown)


jgthms:
I think it's used in case the `.tile` is an `<a>` element.

PinkiNice:
Looks like this case is could be covered by adding 'is-block' class if necessary. Adding `display: block` is little overkill for such purposes, what do you think?

PR #2123, opened 2018-09-24
Add content ol types - fixes #1940

### LINKED ISSUE #1940: Support different ordered list (ol) styles
This is about **Bulma**.

Change request

### Overview of the problem

I'm using Bulma **version** 0.7.1
My **browser** is: Firefox 60, Chrome 66

### Description

With Bulma, the following HTML code is displayed with decimals instead of roman numerals:

```html
  <div class="content">
    <ol type="I">
      <li>First</li>
      <li>Second</li>
      <li>Third</li>
    </ol>
  </div>
```

This is because Bulma defines the following rule:
```css
.content ol {
  list-style: decimal outside;
}
```

So with Bulma, it's currently impossible to style an ordered list without writing additional CSS rules. I propose that either

- the `list-style` CSS rule is dropped so that a list can be customized using the HTML `type` attribute, or
- Bulma defines classes like `has-roman-enumeration` that can be used to define a different ordered list style.

### Steps to Reproduce

1. Save the following code as an HTML file:
```html
<!DOCTYPE html>
<html>
<head>
    <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/bulma/0.7.1/css/bulma.min.css">
    <title>Bulma List Test</title>
</head>
<body>
  <div class="content">
    <ol type="I">
      <li>First</li>
      <li>Second</li>
      <li>Third</li>
    </ol>
  </div>
</body>
</html>
```
2. Open the file in a browser.

### Expected behavior

Using Bulma, it should be possible to define different ordered list styles without writing CSS.

### Actual behavior

Bulma forces the ordered list style `decimal` inside `.content`.

slawekrewaj:
Add list style modifiers emulating type attribute values different then decimal to ol element in content class - fixes #1940

This is an improvement.
Currently bulma content doesn't support any way of adding semantic ol types in HTML. ol type attribute added in HTML is ignored.

### Proposed solution
Add the following modifiers to content ol element:
- is-lower-roman
- is-upper-roman
- is-lower-alpha
- is-upper-alpha

### Tradeoffs
This solutions is easy to use and not coupled to any other element.
It increases build time and complexity minimally.
The performance penalty should also be negligible.

### Testing Done
I have created a simple content div with embedded ol element. I tried all classes and the display was correct. Example below.

```
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8">
    <meta name="viewport" content="width=device-width, initial-scale=1">
    <title>Hello Bulma!</title>
    <link rel="stylesheet" href="css/bulma.min.css">
  </head>
  <body>
  <section class="section">
    <div class="content">
      <ol class="is-upper-alpha">
        <li>ala</li>
        <li>ela</li>
      </ol>
    </div>
  </section>
  </body>
</html>
```

### DIFF
--- a/sass/elements/content.sass
+++ b/sass/elements/content.sass
@@ -73,6 +73,15 @@ $content-table-foot-cell-color: $text-strong !default
     list-style: decimal outside
     margin-left: 2em
     margin-top: 1em
+    // list style modifiers emulating ol[type] attribute values different then decimal
+    &.is-lower-roman
+      list-style-type: lower-roman
+    &.is-upper-roman
+      list-style-type: upper-roman
+    &.is-lower-alpha
+      list-style-type: lower-alpha
+    &.is-upper-alpha
+      list-style-type: upper-alpha
   ul
     list-style: disc outside
     margin-left: 2em


r0the:
Thanks, that is exactly what I need!

PR #2141, opened 2018-10-04
Remove $desktop navbar dropdown hover, Fixes #2138

### LINKED ISSUE #2138: The colors of a navbar-item that is inside a navbar-dropdown, doesn't apply properly
This is about **Bulma**.
 Is it a bug.

### Overview of the problem

My **browser** is: Chrome

### Description

when hovering over a navbar-item that is inside a navbar-dropdown in a nav element, there is a color property on hover inside a media query that overwrites the default color of "a.navbar-item:hover" that has a color of "#3273dc" that is being changed for "#0a0a0a" and the properties inside the media query at the same time overwrites the default background-color on hover state from a color of "#fafafa" to "whitesmoke".

### Steps to Reproduce

inspect in a browser your own example of a "Basic Navbar" https://bulma.io/documentation/components/navbar/#basic-navbar
(because the same happens here)
- mark the parent navbar-element with :hover (the one with the dropdown obviously) for it to stay down.
- mark the 'a' tag that says "About" with :hover and look in the inspector the color being overwritten.

bulmaissue

### Expected behavior

Media queries should not overwrite the default color of any element.

### Actual behavior

The class for elements "a.navbar-item:hover" with a default color is being overwrite by a media query
@media screen and (min-width: 1088px)

Krisseck:
Simple solution to #2138, just remove the hover actions on $desktop media-query.

### DIFF
--- a/sass/components/navbar.sass
+++ b/sass/components/navbar.sass
@@ -34,8 +34,6 @@ $navbar-dropdown-z: 20 !default
 $navbar-dropdown-boxed-radius: $radius-large !default
 $navbar-dropdown-boxed-shadow: 0 8px 8px rgba($black, 0.1), 0 0 0 1px rgba($black, 0.1) !default
 
-$navbar-dropdown-item-hover-color: $black !default
-$navbar-dropdown-item-hover-background-color: $background !default
 $navbar-dropdown-item-active-color: $link !default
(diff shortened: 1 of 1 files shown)


jgthms:
Thanks but as mentioned in #2138, it's not actually a problem.

PR #2142, opened 2018-10-04
Change dropdown icon to use pure CSS, Fixes #2118

### LINKED ISSUE #2118: Dropdown chevron without font-awesome
This is about **Bulma**.

### Overview of the problem

### Description
Right now the `select` element has a (default blue) chevron made purely with CSS. The `dropdown`, instead, uses font-awesome for the chevron/arrow.

It would be good to add the capability of adding a pure CSS implementation of the arrow also for the dropdown element.

Krisseck:
As in the issue #2118, Dropdown component uses fontawesome icons to display arrow. I changed so that it uses the pure CSS arrow used in for example `select` elements.

Also updated the docs to match this new markup, ´<i>´ is no longer needed in Dropdowns.

### DIFF
--- a/sass/components/dropdown.sass
+++ b/sass/components/dropdown.sass
@@ -1,5 +1,5 @@
 $dropdown-content-background-color: $white !default
-$dropdown-content-arrow: $link !default
+$dropdown-content-arrow: $black !default
 $dropdown-content-offset: 4px !default
 $dropdown-content-radius: $radius !default
 $dropdown-content-shadow: 0 2px 3px rgba($black, 0.1), 0 0 0 1px rgba($black, 0.1) !default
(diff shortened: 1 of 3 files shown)


sass/components/dropdown.sass
```diff
@@ -17,6 +17,11 @@ $dropdown-divider-background-color: $border !default
   display: inline-flex
   position: relative
   vertical-align: top
+  &::after
```
jgthms: This shouldn't be under `.dropdown` but under a new `.dropdown-button`.

sass/components/dropdown.sass
```diff
@@ -72,3 +80,7 @@ a.dropdown-item
   display: block
   height: 1px
   margin: 0.5rem 0
+
+.dropdown-trigger
+  button
+    padding-right: 2.3em;
```
jgthms: This value doesn't seem right.

PR #2147, opened 2018-10-08
bulma-start better 'getting started'

hmaesta:
This is a **documentation fix.**

Instructions are confused because `bulma-start` it's not an extension or component – it's a template. This way, using `npm` and copying files from `node_modules` is NOT intuitive.

`git clone` gives a better understanding of how to get started, since it's obvious that you are copying files and not adding components to your project (as usual with `npm`).

### Proposed solution
Using `git clone` instead of `npm` or `yarn`.

### Tradeoffs
User should have `git` installed – but I don't think this may be a problem, since `npm` is not native either.

_Ps: sorry for the typo on "best"_

### DIFF
--- a/docs/bulma-start.html
+++ b/docs/bulma-start.html
@@ -30,24 +30,16 @@ breadcrumb:
 
 {% include elements/anchor.html name="Install" %}
 
-{% highlight bash %}
-cd your-projects-folder  
-git clone https://github.com/jgthms/bulma-start
-mv bulma-start my-new-bulma-site
-rm -rf my-new-bulma-site/.git       # remove original git settings
-{% endhighlight %}
-
-<p class="block"><em>or</em> <a href="https://github.com/jgthms/bulma-start/archive/master.zip" rel="nofollow noopener">download .zip</a></p>
-
-Now that you prepared the groundwork for your project, set up Bulma and run the watchers:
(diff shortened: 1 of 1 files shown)


jgthms:
I agree with the fact that using `node_modules` is not intuitive. I thought about changing this part of the docs, but not like this actually.

I believe most people would rather download the folder rather than clone it. So the workflow would be:

* download and extract `bulma-start-master.zip`
* `cd` into that folder
* run `npm install` or `yarn install`

hmaesta:
Your suggestion is way simpler (and better).
What do you think?

docs/bulma-start.html
```diff
@@ -30,11 +30,16 @@
 
 {% include elements/anchor.html name="Install" %}
 
-{% highlight bash %}npm install bulma-start{% endhighlight %}
+<a href="https://github.com/jgthms/bulma-start/archive/master.zip" rel="nofollow noopener">Download bulma-start (.zip)</a>, rename to your project's name (i.e. `my-new-bulma-site`) and:
 
-<p class="block"><em>or</em></p>
+{% highlight bash %}
+cd my-new-bulma-site
+rm -rf .git        # remove original git settings
```
jgthms: I don't think this part is necessary anymore. The `.zip` doesn't include any `.git` folder.

PR #2157, opened 2018-10-14
Allow border radius if only one control in field, fixes #1656

### LINKED ISSUE #1656: change/feature Request: Border-radiuses on .field.has-addons
This is about **Bulma**.

### Overview of the problem

I'm using Bulma **version** 0.6.2
My **browser** is: Opera

### Description

the Field with addons is made in a way that when changing from a multiple elements in the field to a single one requires removing this addon-class as well, otherwise the border-radiuses would be killed, this is because of the following styles:

buttons have by default rounded corners. by inserting them into a field.has-addons the following things are applied:
~~~
if it is not the first and not the last child, all radiuses become zero

if it is the first child, the right border-radiuses become zero

if it is the last child, the left border-radiuses become zero.
~~~

wouldn't it make more sense to only say:
~~~
not first child -> left border-radiuses become zero

not last-child -> right border-radiuses become zero.

therefore:
only-child -> no borders changed
middle-child (not first and not last) -> all border-radiuses become zero
~~~

this would not only remove some redundant CSS (not a lot, but whatever) but also make scripting logic a lot easier, because one doesnt have to care, whether the field has just one or multiple children.

### Steps to Reproduce

create field.has-addons with only one 

### Expected behavior

border-radiuses are applied (like with button-groups in bootstrap)

### Actual behavior

the border-radiuses are zero

jcs224:
This is an **improvement**.

If there is only one child inside a field with addons, the current first and last child selectors remove any radius. Since a single control would be both a first and last child at the same time, it gets canceled out. Fixes #1656 

### Proposed solution

This solution leverages the :only-child property to not apply this rule if there is only one control within a given field with addons.

### Tradeoffs

None that I an think of, other than it may suddenly add radius to a .control only used once under a .field.has-addons.

### Testing Done
I've tested this feature. Here is a field group:
screen shot 2018-10-14 at 1 28 11 am
And here is what this same field group with only one control looks like:
screen shot 2018-10-14 at 1 28 30 am
Notice the missing radius? This is what this fix does, before and after with one, two, three buttons:
screen shot 2018-10-14 at 1 30 47 am
screen shot 2018-10-14 at 1 31 05 am
screen shot 2018-10-14 at 1 31 25 am

With the addition of the :not(:only-child) selector, the :first-child and :last-child rules are ignored when there's only one child.

### DIFF
--- a/sass/elements/form.sass
+++ b/sass/elements/form.sass
@@ -437,13 +437,13 @@ $help-size: $size-small !default
         .input,
         .select select
           border-radius: 0
-      &:first-child
+      &:first-child:not(:only-child)
         .button,
         .input,
         .select select
           border-bottom-right-radius: 0
           border-top-right-radius: 0
-      &:last-child
+      &:last-child:not(:only-child)
         .button,
         .input,
         .select select


My1:
awesome. thanks.

PR #2170, opened 2018-10-20
Fixes #2165

### LINKED ISSUE #2165: Missing docs for column.is-1 class
This is about **Bulma** - a missing feature documentation?

### Overview of the problem

This is about the Bulma **documentation** 
I'm using Bulma **version** [0.7.2]

### Description

I try to use Bulma for a project where I need the first column as being very narrow: 1/12. But there is no `is-1` class for columns - was the [documentation](https://bulma.io/documentation/columns/sizes/) left out by intention? There is an `is-2` class documented; and there an `is-11` class which I *could* use for the second column, but that's only a workaround.
I then saw that there actually **IS** an `is-1` class that works, it's just not documented.

### Steps to Reproduce

```
<div class="columns">
  <div class="column is-1">Test<div>
  <div class="column is-11">Test<div>
</div>
```
### Expected behavior

one fixed narrow column, 11 cols rest.

### Actual behavior

Works. Docs are missing.

StereotypicalCat:
Added missing documentation for the is-1 class.

### Proposed solution
Fixes issue #2165 by adding is-1 to the docs.

### Tradeoffs
Small load time impact, maybe some confusion as to why you would want is-1

### Testing Done
Please see the picture, from where i built the docs with Jekyll on my local machine.
image

### DIFF
--- a/docs/documentation/columns/sizes.html
+++ b/docs/documentation/columns/sizes.html
@@ -288,6 +288,7 @@ breadcrumb:
 <div class="content">
   <p>As the grid can be divided into <strong>12</strong> columns, there are size classes for each division:</p>
   <ul>
+    <li><code>is-1</code></li>
     <li><code>is-2</code></li>
     <li><code>is-3</code></li>
     <li><code>is-4</code></li>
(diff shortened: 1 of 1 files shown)


docs/documentation/columns/sizes.html
```diff
@@ -306,7 +307,44 @@
...
+    <p class="bd-notification is-wajofij has-text-centered">1</p>
+  </div>
+  <div class="column">
+    <p class="bd-notification is-fojeisj has-text-centered">1</p>
+  </div>
+  <div class="column">
+    <p class="bd-notification is-wajofij has-text-centered">1</p>
+  </div>
+  <div class="column">
+    <p class="bd-notification is-fojeisj has-text-centered">1</p>
```
jgthms: Can you remove all `is-fojeisj` and `is-wajofij` classes?

StereotypicalCat: Arh yeah of course! Should be fixed now. Is that jekyll making those?

jgthms:
I didn't realise I put them in the first place. They don't anything. I need to clean that up.

StereotypicalCat:
If you want, i could clean it?

jgthms:
That’s alright but thanks.

Issue #2172, opened 2018-10-21
The first tab element is displaced vertically


micobarac:
This is about **Bulma**.

### Overview of the problem

This is about the Bulma **CSS framework**.

I am using the latest Bulma version.

### Description

The first tab element is displaced vertically a few pixels, forming a small gap from the bottom border. This problem is almost 2 years old, but wasn't fixed along the way.

https://github.com/jgthms/bulma/issues/474

### Steps to Reproduce

I am facing the this issue on Windows 10 x64 with the latest Chrome version.

![bug](https://user-images.githubusercontent.com/16001837/47269287-fa20e480-d55b-11e8-95b4-816937a5784b.png)

```html
<section class="section content">
  <div class="container is-fluid">
    <h1 class="title">Dashboard</h1>
    <div class="tabs">
      <ul>
        <li class="is-active"><a>Pictures</a></li>
        <li><a>Music</a></li>
        <li><a>Videos</a></li>
        <li><a>Documents</a></li>
      </ul>
    </div>
  </div>
</section>
```

 Aligning items to the baseline did the trick.

```sass
.tabs ul { align-items: baseline; ... }
```

![ok](https://user-images.githubusercontent.com/16001837/47269324-600d6c00-d55c-11e8-91cd-88dde55cc4a7.png)

### Expected behavior

All tabs should be aligned vetically.

### Actual behavior

The first tab is displaced.

mobayen:
same here
this style causes 
```css
.content li + li {
    margin-top: 0.25em;
}
```
either removeing it or lose the "+ li" part will solve the issue

MikelisB:
If it helps resolving it - same to me and the fix by OP did the trick. Thanks! :)

logological:
The issue appears to have been fixed in Bulma 0.7.5.  (I can reproduce it with Bulma 0.7.4, but not Bulma 0.7.5.)  I think this issue can be closed now.

Issue #474, opened 2017-01-17
Tabs Border not working right in Chromium


alexanderwe:
### Overview of the problem

This is about the Bulma **CSS framework**

I'm using Bulma **version** [0.3.1] 
My **browser** is:  Chromium 53.0.2785.143 (Electron)

### Description
I am using Bulma with Electron and React. When I want to use the simple horizontal tabs, a small issue is happening when I set `is-active` on the first tab. The bottom border seems to be 1px higher than the base border. When `is-active` is set on the second tab, everything is working fine.

<img width="404" alt="bildschirmfoto 2017-01-17 um 20 31 40" src="https://cloud.githubusercontent.com/assets/12934015/22037060/13ae1d92-dcf6-11e6-9642-c80b6a470c62.png">

```
<div class="tabs">
    <ul>
        <li class ="is-active"><a>Pictures</a></li>
        <li><a>Music</a></li>
        <li><a>Videos</a></li>
        <li><a>Documents</a></li>
    </ul>
</div>
```

This problem also applies to all other tab implementations. 

### Steps to Reproduce

1. Use Chromium
2. Insert example HTML from http://bulma.io/documentation/components/tabs/

### Solution

I changed the CSS for the tabs class from

`.tabs ul {
    align-items: center;
    ...
}`

to 

`.tabs ul {
    align-items: baseline;
    ...
}`

and everything is working as expected. 

Is this the expected behavior ? If this is not the case, is it possible to change the CSS according to my solution ? 

Thanks in advance. And also thanks for this great CSS framework !

ilicmarko:
This can't be reproduced, has it been fixed?
Tested in:

- Firefox 50.1.0
- Chrome 55.0.2883.87 m
- Chrome 57.0.2984.0 canary
- Edge 38.14393.0.0
- Opera 42

alexanderwe:
Maybe the Problem is related specific to Electron or within my Setup.

Yes, it was solved with changing the CSS to

 `.tabs ul { align-items: baseline; ... }`.

alexanderwe:
I think because it can not be reproduced I can close this issue for now.

Twan-Blum:
Had the same problem. Issue seems to be when you add tabs inside container with the "content" class:

bulma.css 1690:
`.content li + li {
    margin-top: 0.25em;
}`

henriquemattos:
[...]

PR #2175, opened 2018-10-23
proper aligning for tabs

diomed:
# This is a  bugfix.

### Proposed solution
[x] fix for #2172

replaces center with baseline, so all tabs are properly aligned.

### DIFF
--- a/sass/components/tabs.sass
+++ b/sass/components/tabs.sass
@@ -58,7 +58,7 @@ $tabs-toggle-link-active-color: $link-invert !default
         border-bottom-color: $tabs-link-active-border-bottom-color
         color: $tabs-link-active-color
   ul
-    align-items: center
+    align-items: baseline
     border-bottom-color: $tabs-border-bottom-color
     border-bottom-style: $tabs-border-bottom-style
     border-bottom-width: $tabs-border-bottom-width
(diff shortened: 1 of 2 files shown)


jgthms:
Do you have an example where this is an issue?

diomed:
example is shown in issue which I'm closing. 
problem is explained [here](https://github.com/jgthms/bulma/issues/474#issuecomment-304547082)

jgthms:
The solution for that particular problem is to change the following rule in `content.sass`:

```sass
li + li
  margin-top: 0.25em
```

And make sure it's not applied for tabs.

diomed:
so do I change to 0 and modify PR?
is this is?

jgthms:
Yeah change this PR so this selector `li + li` doesn't target `li` that are within a `.tabs`.

diomed:
I've tried to make this shorter but I was getting errors. 
I hope this is all right?

jgthms:
No that’s the wrong output. The “not tabs” part should be for the ul, not the li.

diomed:
@jgthms so - like this then?

if not, can you please update it properly so this one could be fixed already.

jgthms:
Just add before `// Sizes`:

```sass
.tabs
  li + li
    margin-top: 0
```

jgthms:
You've indented that one level too far. And you can remove the rest.

diomed:
like this?

Issue #349, opened 2016-10-15
Support for secondary font-family


neovive:
### Description

Add the ability to specify a secondary font-family, making it easier to create font combinations for headers and body text.

This will likely require additional Sass variables for `$family-secondary` and possibly `$family-heading` and `$family-body`.

jimblue:
+1

bigardone:
+1

jeanleonino:
Updates on that?

christopherdupont:
+1

jeanleonino:
@jgthms you are self-assigned to this bug, would you like some help on this?

sjclark:
This would be fantastic - I think primarily just having the ability to define a heading font family would be fantastic.

Even just: `$title-font-family` ? And/or `$subtitle-font-family` for consistency!

jeanlucaslima:
ping :-)

jeturcotte:
cloooosed?  but but but but but... hurting for this, presently.

emcog:
I have a bit of hack which works for me

- First I divided my two fonts into $family-serif and $family-sans-serif. 
- $family-primary is declared as serif of sans – what ever you want. In my case $family-serif
- .title class is defined as a different font. In my case this is $family-sans-serif

Dividing into $family-serif and $family-sans-serif isn't necessary but I like, as it follows a convention I can easily understand.

emkajeej:
@emcog seems to me as the Bulma way of customising. This is how I would do it and there's nothing stopping anyone from using this approach either, I think,

mlncn:
It seems Bulma should come with a `$family-serif` by default (it has a `$family-sans-serif` so many will assume there's a serif family to match) and then a `$family-heading`.  An example could show how to set `$family-heading` to `$family-serif`.

daggepagge:
Support for a secondary font-family for titles in Bulma would change my life basically. Big fat +1

emcog:
```
$family-serif: "adobe-garamond-pro", serif
$family-sans-serif:  "futura-pt", sans-serif
$family-primary: $family-serif
.title
	font-family: $family-sans-serif

```

cyrilchapon:
Hey,

We're facing this, and fighting against with a single `.has-alternative-font` for now.. (ugly but working)

Regarding multi-family support, IMHO and with experience of our website built with bulma :
- There should be way more than just a `heading-family` variable
  - `button-family` could make sense
  - `input-family` same
  - `heading-family` indeed
[...]

PR #2178, opened 2018-10-24
Add $family-secondary, $family-tertiary, and 6 font family helpers

lesleslie:
This is a **improvement**.

This adds a 2 new font family variables and 6 font family helpers.

I believe these variables and helpers will solve a lot of the talk that was going around in #349.

The 2 new variables $family-secondary and $family-tertiary will give people a few more semantic options for setting custom fonts and are set by default to $family-sans-serif just like $family-primary.

The 6 new helpers make it very easy to customize the text of any element:

.has-font-primary
.has-font-secondary
.has-font-tertiary
.has-font-sans-serif
.has-font-monospace
.has-font-code

These improvements work as expected,  are straightforward, and as far as I can tell should require little to practically no maintenance. The footprint of these improvements is also very light as far as size is concerned.

I personally use these improvements with Bulma, find them extremely useful and think most, if not all,  Bulma users would appreciate and use them as well (#349). Doc updates included in this PR.

Thank you for making such an amazing framework!!!

### DIFF
--- a/sass/base/helpers.sass
+++ b/sass/base/helpers.sass
@@ -119,18 +119,20 @@ $alignments: ('centered': 'center', 'justified': 'justify', 'left': 'left', 'rig
   font-weight: $weight-semibold !important
 .has-text-weight-bold
   font-weight: $weight-bold !important
-  
-.has-font-primary
+
+.is-family-primary
   font-family: $family-primary !important
-.has-font-secondary
+
+.is-family-secondary
   font-family: $family-secondary !important
-.has-font-tertiary
-  font-family: $family-tertiary !important
-.has-font-sans-serif
+
+.is-family-sans-serif
   font-family: $family-sans-serif !important
-.has-font-monospace
+
+.is-family-monospace
   font-family: $family-monospace !important
-.has-font-code
+
+.is-family-code
   font-family: $family-code !important
 
 // Visibility
--- a/sass/base/helpers.sass
+++ b/sass/base/helpers.sass
@@ -119,6 +119,19 @@ $alignments: ('centered': 'center', 'justified': 'justify', 'left': 'left', 'rig
   font-weight: $weight-semibold !important
 .has-text-weight-bold
   font-weight: $weight-bold !important
+  
+.has-font-primary
+  font-family: $family-primary !important
+.has-font-secondary
+  font-family: $family-secondary !important
+.has-font-tertiary
+  font-family: $family-tertiary !important
+.has-font-sans-serif
+  font-family: $family-sans-serif !important
+.has-font-monospace
+  font-family: $family-monospace !important
+.has-font-code
+  font-family: $family-code !important
 
 // Visibility
 
--- a/docs/documentation/modifiers/typography-helpers.html
+++ b/docs/documentation/modifiers/typography-helpers.html
@@ -417,7 +417,7 @@ breadcrumb:
 
 <div class="content">
   <p>
-    You can change the font family with the use of one of <strong>5 font family helpers</strong>:
+    You can change the font family with the use of one of <strong>6 font family helpers</strong>:
   </p>
 </div>
 
--- a/docs/documentation/modifiers/typography-helpers.html
+++ b/docs/documentation/modifiers/typography-helpers.html
@@ -396,7 +396,7 @@ breadcrumb:
   <tbody>
   <tr>
     <td><code>has-text-weight-light</code></td>
-    <td>Transforms  text weight to <strong>light</strong></td>
+    <td>Transforms text weight to <strong>light</strong></td>
   </tr>
   <tr>
     <td><code>has-text-weight-normal</code></td>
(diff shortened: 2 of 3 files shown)


cyrilchapon:
As per my comment here https://github.com/jgthms/bulma/issues/349#issuecomment-431360253, I think those additional fonts should come with :
- not only helpers but configurability on each component / element of the framework
- a set of size for each font (every font don't have the same visual size)

lesleslie:
@cyrilchapon I think you are making this too confusing which is one reason I don't think font family helpers have been implemented yet.

- To add add a helper for each component/element would increase the code size footprint considerably.

- Are you really using more than three fonts that you need to have a different one for each component? 
If you are loading custom fonts (ie Google Fonts) you shouldn't prob be using more than two anyway to keep your page load times down.

- This option covers multiple possibilities for use. You can say $family-primary is for your body text, $family-secondary is for you titles, and $family-tertiary for your subtitles and then use the helpers accordingly.

As far as helpers for configurability of size and what not of each component / element goes:

1. Again this would increase the footprint of the code base.

2. If your looking to change font size, color, weight, etc on each component / element, these helpers already exist:

- has-test-weight-semibold etc.
- is-size-4 for text or is-4 for titles etc.
- has-text-color-white etc.

If you need more configurability you can always do that in your stylesheet or if you look at my pull request see how easy it too create new variables / helpers for your own use. I know I do.

Just my thoughts / approach on this.

cyrilchapon:
Hey @lesleslie

I think you misunderstood my point about configurability, I'm personaly using only 2 fonts and am facing 2 troubles :

I made a variable just like yourself (alternativeFamily, actualy), but think we just would lack the ability to **assign** those \*ary fonts. I'm just thinking about, for example, $buttons-family wich would default to $primary-family, and one could configure to be $secondary-family. Not more footprint than a couple of font-size actualy.

As for font-size, I agree, the edge cases are hard to handle and would lead to much headache and footprint

jgthms:
I like the ideas discussed in this PR. I think `$family-secondary` makes sense. I'm not convinced by `$family-tertiary`: most websites will only have 2 families (one for headings and/or buttons, one for everything else).

For websites that will need 3 families, we can add what @cyrilchapon talked about: Sass variables for some elements, like buttons or titles. This would introduce new variables like `$button-family` and `$title-family` which could default to `$family-secondary`.

In the case where a user wants to use a 3rd family, they can simply create their own `$family-tertiary` and override the default value like `$button-family: $family-tertiary`.

As for the font sizes, we could have `$family-primary-size` and `$family-secondary-size`, and even `$family-code-size` actually.

sass/base/helpers.sass
```diff
@@ -119,6 +119,19 @@ $alignments: ('centered': 'center', 'justified': 'justify', 'left': 'left', 'rig
   font-weight: $weight-semibold !important
 .has-text-weight-bold
   font-weight: $weight-bold !important
+  
+.has-font-primary
```
jgthms: I would call them `.is-family-primary`.

lesleslie:
Ok @jgthms, I agree $family-tertiary is not necessary and have removed it. I've also changed '.has-font' to '.is-family' as suggested for the helpers and updated the docs accordingly.

As for $button-family, $title-family, etc I usually build buttons and such from template macros so adding $is-size-4 or$ is-family-primary to the macro is just as easy as adding an extra variable for such things - but, as I would consider my use cases fairly simple, I can definitely see how these would be useful. Is there a consensus between you and @cyrilchapon as to which all elements should be included. The suggestion from #349 so far are:
$button-family
$title-family
$subtitle-family?
$header-family
$input-family
$body-family

If we're going to go ahead and add these then I think it would be appropriate, at this point, to go ahead and add the $*-family-size for each of these as well as for $family-primary-size, $family-secondary-size, and $family-code-size (or should we say $primary-family-size to make it consistent with $button-family-size?).

I'm still getting up to speed with SASS and am willing to work on it if you'd like.

Also while I have your attention @jgthms, I have found adding responsive helpers for $is-marginless and $is-paddingless very helpful. ie:
$is-marginless-touch
$is-paddingless-desktop
If you feel these would be useful in including in bulma I will add them in another PR.

cyrilchapon:
Hey @lesleslie @jgthms thanks for your work and your attention on it. Indeed some great discussions in here.

So sum up, there are 4 topics here :

- 1) Font-family variables creation, and a couple of standardized helpers. This PR.
- 2) Font-family derived-variables, atomic to assign variable to elements.
- 3) Font-size relative to those families.
- 4) Responsive padding and margin helpers.

So

- 1) is ok, done

- 2) and 3) should come in the same PR IMHO

> which all elements should be included. The suggestion from #349 so far are:
> - $button-family
> - $title-family
> - $subtitle-family?
> - $header-family
> - $input-family
> - $body-family

I just digged into bulma's doc to consider every single thing we can tweak the family on.

- ~Columns~ : nope
- Layout :
  - ~Container~ : nope
  - ~Level~ : nope
  - Media : *would deserve discussion*
  - ~Hero~ : nope
  - ~Section~ : nope
  - Footer : *would deserve discussion*
  - ~Tiles~ : nope
- Form : *would deserve discussion : would whold form deserve a variable ?*
  - Control : Yes
  - Input : Yes (inherit Control ?)
  - Textarea : Yes (inherit Control ?)
  - Select : Yes (inherit Control ?)
  - Checkbox : Yes (inherit Control ?)
  - Radio : Yes (inherit Control ?)
  - File : Yes (inherit Control ?)
- Elements
  - Box : *would deserve discussion*
  - Button : Yes (inherit Control ?)
  - Content : I think no... ?
  - ~Delete~ : nope
  - ~Icon~ : nope
  - ~Image~ : nope
  - Notification : Yes, definately
  - ~Progress bars~ : Nope for now, careful if we add percentage one day
  - Table : Yes, and maybe split up th and the rest
  - Tag : Yes, definately
  - Title : Yes, definately
- Components
  - Breadcrumb : Yes
  - Card : Yes, probably multiples
  - Dropdown : Yes (inherit Control ?)
  - Menu : Yes
  - Message : Yes, probably multiples
  - Modal : Yes, probably multiples
  - Navbar : Yes
  - Pagination : Yes
  - Panel : Yes
  - Tabs : Yes

--------

So I sum up

# The simple ones

- Button
- Notification
- Tag
- Title
- Breadcrumb
- Menu
- Pagination
- Tabs

# The "inherit" ones

- Control ?
- Input
- Textarea
- Select
- Checkbox
- Radio
- File
- Dropdown

# The harsh ones

*These ones can be hard to anticipate, because they can include some other components / elements which can hold family. Should the parent family take-over the children one ? Or because they could be split into sub families (like message-title and message-content)*

- Media
- Footer
- Form
- Box
- Table
- Card
- Menu
- Message
- Modal
- Navbar
- Panel

codev0:
Hey, guys, don't forget update npm version, please

recio-sjogren:
Hi all,
I can't find $family-secondary on the bulma start npm package. Did you manage to update it?

PR #2181, opened 2018-10-25
`Type` removed from textarea

hmaesta:

`Type` is not a valid attribute for <textarea> tag. [(W3Schools)](https://www.w3schools.com/tags/tag_textarea.asp)


jgthms:
Nice, thanks!

PR #2182, opened 2018-10-26
Fixed responsiveness for logo in README.md

scriptcoded:

### Proposed solution
The Bulma logo was not responsive, and stretched oddly on smaller devices. By adding a 600x315 version of the logo, and removing `width="600" height="315"` we manage to keep it responsive.

### Tradeoffs
Virtually none :)

### Testing Done
The solution only adds an extra version of the logo, as well as removes the width and height attributes from the documentation. There should be no problems.

**Before:**
chrome_2018-10-26_07-48-41
**After:**


jgthms:
The solution is actually to just add `height: auto;` to the image.

scriptcoded:
@jgthms Perhaps I was doing something wrong, but I couldn't get GitHub to accept my inline style. And auto isn't a valid value for the width attribute.

jgthms:
It's for the height, not the width.

scriptcoded:
Is it valid for height then?

PR #2187, opened 2018-10-26
Add classes for reverse flex row/column

kbarralon:
This is a new feature.

### Proposed solution
Add a `is-column-reversed` and a `is-row-reversed` classes in order to manage the flex container direction.


cyrilchapon:
Careful, I'm using that in production, and this sort of isn't compatible with the `is-offset-` approach.
That depends of what you want actually, but one should keep in mind the offset is not inverted.

```
[   ]  [   5 offset 1   ]  [                6                ]
```

Inverted

```
[                6                ] [  ] [   5 offset 1   ]
```

(the offset is still on the **left**)

Issue #2091, opened 2018-09-06
.tags.has-addons.is-right introduces margins at wrong place


bertBruynooghe:
This is about **Bulma**.

### Overview of the problem

This is about the Bulma **CSS framework**

I'm using Bulma **version** [0.7.1]

### Description

When using `class="tags has-addons is-right", a margin is introduced between the tag parts.

### Steps to Reproduce

    <div class="tags has-addons is-right">
        <span class="tag">Package</span>
        <span class="tag is-primary">Bulma</span>
    </div>

### Expected behavior

'Package' and 'Bulma' should be nicely glued together, aligned to the right.

### Actual behavior

There is a gap between 'Package' and 'Bulma'

msangui:
I think the right fix is to add a quick "exception" in elements/tag.sass:


     &.is-centered, &.is-right
        .tag
          margin-left: 0
          margin-right: 0


in line 28 right after .has-addons definition

bertBruynooghe:
FYI: I patched locally using:   

    .tags.is-right.has-addons .tag:not(:first-child) {
        margin-left: 0;
    }

PR #2188, opened 2018-10-27
#2091 remove css rule which causes has-addons to not work correctly

gvdp:
This is a **bugfix**.

### Proposed solution
Remove css line which causes extra margin mentioned in #2091 

### Tradeoffs
Removed css rule will not be used anywhere else so removing it will not cause issues.

### Testing Done

Icons now look like this:

image

With following html:

```
<div class="tags has-addons is-right">
  <span class="tag">Package</span>
  <span class="tag is-primary">Bulma</span>
</div>
```


sass/elements/tag.sass
```diff
@@ -34,7 +34,7 @@ $tag-delete-margin: 1px !default
     justify-content: flex-end
     .tag
       &:not(:first-child)
-        margin-left: 0.5rem
+        margin-left: 0
```
jgthms: You're targeting all tags, not the addons one only. You need to put the the code under `&.has-addons`.

gvdp:
Yep, you're right. I made the changes in the css and then didn't copy it in the right place in the .sass, my bad. Tested some more combinations now with the actual compiled css, looks like this so I think all still works:

image

jgthms:
You need to remove the compiled CSS from your PR.

gvdp:
> You need to remove the compiled CSS from your PR.

Done

PR #2194, opened 2018-10-28
Makes is-full-height-with-navbar samples match

isaaczafuta:

The code-sample in the Hero documentation didn't quite match the example. The `is-full-height-with-navbar` class had been omitted.

### Proposed solution

Add the missing class name. Also, the code-sample used `<h1>` while the example used `<p>`, so I made that match up too.
### Tradeoffs

None?

### Testing Done
Ran jekyll locally-- looks good.

The code-sample now includes the `is-full-height-with-navbar` class, like so:
2018-10-28-005909_467x183_scrot

### DIFF
--- a/docs/documentation/layout/hero.html
+++ b/docs/documentation/layout/hero.html
@@ -469,12 +469,12 @@ meta:
   </div>
 
 {% highlight html %}
-<section class="hero is-link is-fullheight">
+<section class="hero is-link is-fullheight-with-navbar">
   <div class="hero-body">
     <div class="container">
-      <h1 class="title">
+      <p class="title">
         Fullheight hero with navbar
-      </h1>
+      </p>
     </div>
   </div>
 </section>


docs/documentation/layout/hero.html
```diff
@@ -469,12 +469,12 @@ <h2 class="subtitle">
   </div>
 
 {% highlight html %}
-<section class="hero is-link is-fullheight">
+<section class="hero is-link is-fullheight-with-navbar">
```
jgthms: This class is only used when the navbar is fixed.

isaaczafuta: Sorry-- perhaps I'm not understanding something.

I believe this line is the code-snippet corresponding to the example right above it. Then shouldn't this line match with line 460 above?

jgthms: Ah yes of course! I missed that. Sorry, you're right.

PR #2195, opened 2018-10-28
improvement - Bring code-example as readable HTML

delphianer:
In all other documentation -> column documents there is the source code clearly written, so I missed it in the nested documentation and brought it in. Please verify and thanks for merge :+1:

This is a **improvement**.

I missed the HTML to be clearly readable (without going to "view source")

### Proposed solution
There was no issue. Life would go on without it, but it's easier to read, so why not?

### Tradeoffs

There are alternatives like Right-Click -> View Source.

### Testing Done
Mhm, ok, I believe it works, because I copied the commands for highlight html from gap-docs and really when you look at that tiny bits, I don't think I need to install via npm, or? I hope not, because it is too late this evening :-)

### DIFF
--- a/docs/documentation/columns/nesting.html
+++ b/docs/documentation/columns/nesting.html
@@ -69,6 +69,8 @@ breadcrumb:
 </div>
 {% endcapture %}
 
+{{ columns_nested_example }}
+
 <div class="content">
   <p>
     Multiline columns will also have a <strong>gap</strong> between each <strong>line</strong>.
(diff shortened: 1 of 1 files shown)


jgthms:
Yeah but `capture` doesn't output the snippet. So you need to display it again. Because right now, you removed the actual output, and you just show the code.

delphianer:
Thanks for your quick comment :smiley: :+1: 
I think it's fixed now :dancer:

jgthms:
Nice one.

Issue #2198, opened 2018-10-29
Contributing is a pain


cyrilchapon:
# What I was expecting
`git clone [url] && cd bulma && npm install && npm run dev`

# What I got instead

## Installation

- cloned the repo
- `npm install`
- `cd docs && npm install`
- `brew install ruby`
- `gem install jekyll`
- Error message, ruby sucks with SSL blablabla
- Don't have time for this
- `gem sources --remove https://etc.. && gem sources -a http://etc..`
- `gem install jekyll`
- Error message, gem sucks with urls
- Don't have time for this
- `gem sources --remove http://etc.. && gem sources -a http://etc../` (with a trailing slash at the end)
- `gem install jekyll`
- Error message, ruby sucks with user permissions blablabla
- Don't have time for this
- `sudo gem install jekyll`

## Configuration

- `cp _config.yml _config.local.yml`
- Edit the url to http://localhost:4000

## Running the stuff

Shell session 1)
- `npm start`

Shell session 2)
- `cd docs && npm start`

Shell session 3)
- `cd docs && jekyll serve --incremental --config _config.local.yml`

# Testing the stuff

- Open browser to http://localhost:4000
- Error message, ruby sucks with network interfaces
- Don't have time for this
- Open browser to http://127.0.0.1:4000
- Page shows but I get 404 for every images / vendor / css / lib

As a result, I have
- 4 **shell opened shell sessions**
- **nothing is working**
- **now way to test my contribution**

Come on, it's almost in 2019 and I just lost 1 hour in installing stuff like people did in 1998 with PHP. That's insane, I was full of motivation to address an issue and that simple fact almost disgusted me to do so. I'm not very brave at it, I confess, but come on, this is a 30k stars repo, can do better.

EDIT: And the funniest part is that bulma in itself installs and compiles greatly, fastly, with 2 node.js commands. That time-and-effort-killing dependency is introduced only to build a static website. Bazooka to kill a fly welcome, except the bazooka basically doesn't work.

jgthms:
There have been 235 contributors so far. I'm pretty sure contributing is easier than think 😉 

I'm developing for Bulma on different machines and have rarely had any issue. It seems it's mostly related to ruby itself, not Bulma.

In order to contribute to the **framework**, you don't necessarily need to run the **documentation website**. In which case, you don't even need ruby. And fixing a typo in the docs only requires to edit some Markdown or HTML file, without running the website.

[...]

PR #2199, opened 2018-10-29
Add family configuration for simple elements

cyrilchapon:
This is a **new feature**.

(I'll edit CHANGELOG and Docs when I'm completely done, and when I'll be able to build the docs)

### Proposed solution
Partialy adresses https://github.com/jgthms/bulma/issues/349, following on https://github.com/jgthms/bulma/pull/2178.

In here I:
- add variables to configure elements font family
- add the actual `font-family` code

### Tradeoffs
A little of code size, a couple of `font-family`

### Testing Done
Nothing, I didn't manage to run the docs. See https://github.com/jgthms/bulma/issues/2198

### DIFF
--- a/sass/components/breadcrumb.sass
+++ b/sass/components/breadcrumb.sass
@@ -2,6 +2,8 @@ $breadcrumb-item-color: $link !default
 $breadcrumb-item-hover-color: $link-hover !default
 $breadcrumb-item-active-color: $text-strong !default
 
+$breadcrumb-family: $family-secondary !default
+
 $breadcrumb-item-padding-vertical: 0 !default
 $breadcrumb-item-padding-horizontal: 0.75em !default
 
@@ -10,6 +12,7 @@ $breadcrumb-item-separator-color: $grey-light !default
 .breadcrumb
   @extend %block
   @extend %unselectable
+  font-family: $breadcrumb-family
   font-size: $size-normal
   white-space: nowrap
   a
--- a/sass/components/menu.sass
+++ b/sass/components/menu.sass
@@ -1,4 +1,5 @@
 $menu-item-color: $text !default
+$menu-item-family: $family-primary !default
 $menu-item-radius: $radius-small !default
 $menu-item-hover-color: $text-strong !default
 $menu-item-hover-background-color: $background !default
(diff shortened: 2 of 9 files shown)


cyrilchapon:
This is WIP :

## The simple ones

- [x] Button
- [x] Notification
- [x] Tag
- [x] Title
- [x] Breadcrumb
- [x] Menu (splitted `item` and `label`)
- [x] Pagination
- [x] Tabs

## The "inherit" ones

- [ ] Control ?
- [ ] Input
- [ ] Textarea
- [ ] Select
- [ ] Checkbox
- [ ] Radio
- [ ] File
- [ ] Dropdown

## The harsh ones
These ones can be hard to anticipate, because they can include some other components / elements which can hold family. Should the parent family take-over the children one ? Or because they could be split into sub families (like message-title and message-content)

- [ ] Media
- [ ] Footer
- [ ] Form
- [ ] Box
- [ ] Table
- [ ] Card
- [ ] Menu
- [ ] Message
- [ ] Modal
- [ ] Navbar
- [ ] Panel

jgthms:
I was actually working on this too. Thanks for helping out.

I want to introduce these family variables but they shouldn't have that default value. Otherwise, you're simply duplicating the font-family value everywhere, although it's the same.

There should be a condition here, and only set it when there's a different value.

cyrilchapon:
hey @jgthms,

> There should be a condition here, and only set it **when there's a different value.**

Different than what ?

jgthms:
Different than `$family-primary`. If there's only one font family across the whole website, there's no need to redefine it, especially when the list of families is quite long.

cyrilchapon:
Oh I got your point. True :)

> If there's only one font family across the whole website, there's no need to redefine it, especially when the list of families is quite long.

Except in the case, of say, a `modal`-family-tweaked containing a `button`-family-tweaked. You see what I mean ?

```
<div class="modal">
  <button class="button">
  </button>
</div>
```

```
$button-family: $family-primary;
$modal-family: $family-secondary;
```

jgthms:
Yeah but is should be optional. People can always enforce the font of the button if they want.

But we should have the following:

```sass
$button-family: null !default

.button
  @if $button-family
    font-family: $button-family
```

So people can set it with `$button-family: $family-secondary` if needed.

cyrilchapon:
Okay right, I'll give it a shot tomorow I think

svdHero:
Any update on this?

PR #2200, opened 2018-10-29
.is-parallax

diomed:
This is a **new feature**.
This PR adds parallax

### Testing Done


jgthms:
I like the idea for an extension I'm not sure this is necessary within Bulma core?

kisamoto:
My two cents - as such a small addition it could be a useful feature as I would class it as a current web design "element". 

I stumbled across this PR looking to see how to integrate parallax effects with Bulma and if there was any "official" support. It would be great to have a simple `.is-parallax` option to quickly change the behaviour of a `div/section`

PR #2211, opened 2018-10-31
Update docs table background vars

motionsuggests:

### Proposed solution
Updated the docs for previous PR https://github.com/jgthms/bulma/pull/2206. 

### Tradeoffs
I'm not quite sure what `type`, `computed_type`,`computed_value` so I took a guess. I'm not aware of any other use of `transparent` currently in the docs. 

### Testing Done
I ran the docs via jekyll an confirmed new variables were in on the table docs page.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## 0.7.3
 
 ### New features
-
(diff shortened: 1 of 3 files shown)


jgthms:
Ah don't worry, I've got a script to do that, but thanks!

Issue #2196, opened 2018-10-29
Anchor elements inside hero loose the text colour


weyert:
I need help but potentially could be a bug

### Overview of the problem

When I am using an anchor element inside a hero block the default text colour of the anchor element gets overridden due to the css rule:  

```
.hero.is-light a:not(.button):not(.dropdown-item):not(.tag), .hero.is-light strong {
    color: inherit;
}
```

I am not sure if it's by design to override the text colour of anchor elements? The HTML that I am using within the anchor element is:

```
<div class="field"><div class="control has-text-centered"><a href="http://127.0.0.1:3000/reset_password">Forgot password?</a></div></div>
```

### Expected behavior

The text colour of a anchor element should not be overriden when it's inside a `hero`-element

### Actual behavior

The text colour of the anchor element gets overridden

PR #2212, opened 2018-11-01
Distinguish link color in hero container and in .is-link elements fixes #1646

### LINKED ISSUE #1646: Links in `.hero.is-light` are not discernable from regular text
This is about **Bulma**.

### Overview of the problem

This is about the Bulma **CSS framework**.

I'm using Bulma **version** 0.6.1
My **browser** is: Firefox 58, Chromium 64 and Epiphany 3.18.11.

### Description

With a layout like this:

```
<section class="hero is-light">
    <p class="content">
        This is content with <a href="#" title="Some title">a link[/a> in it.
    </p>
</section>
```

The `<a>` is not discernible from the rest of the content, whereas in a plain `.hero`, for example, it at least has a different colour.

### Steps to Reproduce

Use the code above.

### Expected behavior

Links are discernible from the rest of the text (preferably in more than one modality, but at least through a different colour).

### Actual behavior

The link is styled the same as the rest of the text.

gvdp:
This is a **improvement | bugfix** for issue #1646 .

### Proposed solution
In the hero container there was an exception for links not to change colors, I removed this so the links are distinguishable even in hero containers.

### Tradeoffs
When you add the `is-link` class to your  `.hero` container the links in it will have the same color as the background so the link will not be visible. This is the same for all elements for which you add the `.is-link` class. To remedy this I added an extra color to the generic sass for the `<a>` tag so now the link is just black. This will effect links contained in a `is-link` element everywhere.

### Testing Done

Examples of some hero containers with the distinguishable links:
image

Examples of tags with links, no change here except for the last `is-link` tag:

image

### DIFF
--- a/sass/base/generic.sass
+++ b/sass/base/generic.sass
@@ -65,9 +65,10 @@ a
   text-decoration: none
   strong
     color: currentColor
+  .is-link &
+    color: $black
   &:hover
     color: $link-hover
-
 code
   background-color: $code-background
   color: $code
--- a/sass/layout/hero.sass
+++ b/sass/layout/hero.sass
@@ -17,7 +17,6 @@
     &.is-#{$name}
       background-color: $color
       color: $color-invert
-      a:not(.button):not(.dropdown-item):not(.tag),
       strong
         color: inherit
       .title


weyert:
Sounds like this fixes my #2196 ticket too

gvdp:
> Sounds like this fixes my #2196 ticket too

Correct, I think it does.

diomed:
@gvdp hello

could you maybe release a fix for [this](https://codepen.io/diomed/pen/oQvaGM) as well?
[button issue, of course]

sass/base/generic.sass
```diff
@@ -65,9 +65,10 @@ a
   text-decoration: none
   strong
     color: currentColor
+  .is-link &
```
jgthms: This is only a fix for one color. And what if `$link: $black`?

sass/layout/hero.sass
```diff
@@ -17,7 +17,6 @@
     &.is-#{$name}
       background-color: $color
       color: $color-invert
-      a:not(.button):not(.dropdown-item):not(.tag),
```
jgthms: This is gonna break a _lot_ of current installs.

PR #2213, opened 2018-11-02
Update _config.yml

diomed:


jgthms:
Thanks but that line is not really used anymore.

diomed:
it it's not used you should delete it

diomed:
if it's not used you should delete it

PR #2232, opened 2018-11-12
Is-borderless helper

damanwiththeplan:
Improvement.

Is useful in certain situations when custom styling is being applied. Should exist considering there's also is-radiusless & is-shadowless.


kornicameister:
@jgthms would be quite nice to have this, could we have it merged for 0.7.3?

PR #2233, opened 2018-11-13
Add style to disabled file input field

bodograumann:
Just as with the button element we add an opacity rule when the file input is disabled.

This is an improvement.
Previously a disabled file input field was not styled differently from an enabled one.

### Tradeoffs
This relies on the .file-cta element being a sibling of the .file-input element.
The `~` selector still gives the user some freedom with respect to the structure of the html.
Alternatively one could use the `+` selector, which trades this flexibility for better efficiency.

### Testing Done
I have tested this feature on my existing forms and it works without any issues.

### DIFF
--- a/sass/elements/form.sass
+++ b/sass/elements/form.sass
@@ -381,6 +381,9 @@ $help-size: $size-small !default
   background-color: $file-cta-background-color
   color: $file-cta-color
 
+.file-input[disabled] ~ .file-cta
+  opacity: 0.5
+
 .file-name
   border-color: $file-name-border-color
   border-style: $file-name-border-style


rogerfar:
@bodograumann any chance you could fix the merge conflict and merge it? :)

bodograumann:
@rogerfar No chance, because I am not a maintainer of this project.
Honestly this has been sitting for so long already without any attention from the maintainer, I don't think it is worth any effort from my side.

Feel free to create a new PR though. The change is not that complex. Then I will close this PR.

danilo-augusto:
Maybe we could add `cursor: not-allowed;` too? ;)

The snippet I use is this one:
```scss
.file-input[disabled] ~ .file-cta {
  opacity: 0.5;
  cursor: not-allowed;
  .file-label {
    // Without this part the cursor appears differently over the text:
    // it would be "not-allowed" only around the text and over the icon.
    cursor: not-allowed;
  }
}
```

PR #2246, opened 2018-11-22
Fix rounded button paddings using a variable

markotom:
This is a **bugfix**.

### Proposed solution
Just added `$button-padding-horizontal` on rounded buttons

### Tradeoffs
We can change rounded paddings on buttons using a custom value


jgthms:
It’s not the same value though. The padding for rounded buttons has to be bigger than for non-rounded ones.

Preen:
This is still relevant though?

Issue #2252, opened 2018-11-26
Gap initial variable is incorrect


wiamsuri:
This is about Bulma.

### Overview of the problem

This is a **Sass** issue: I'm using version 0.7.2.
I think the intended gap size is 32px as initial variables says that
```
// 960px container + 4rem
$desktop: 960px + (2 * $gap) !default
```
But the gap variable is set to `$gap: 64px !default`, which makes `(2 * $gap)` = `8rem`.

### Expected behavior

Gap variable should be set to `$gap: 32px !default` to make `(2 * $gap)` = `4rem`

jgthms:
Thanks but I’m pretty sure it’s. A duplicate. Will close this one.

PR #2253, opened 2018-11-26
Fix gap variable value

wiamsuri:
This is a bugfix.

Fix gap variable value
issue https://github.com/jgthms/bulma/issues/2252

### Proposed solution
The gap variable value should be `32px`.


noymer:
Hi.
I think  $gap is `64px` is intended.

See  `Variable changes->Updated default values` in the below link.
https://github.com/jgthms/bulma/releases/tag/0.7.0

PR #2303, opened 2019-01-03
fix(Tag): fix delete button in tag

revathskumar:
This is a **bugfix**.

When tag is imported alone the delete button inside the tag is not proper
CodeSandbox : https://codesandbox.io/s/43vpkjj35x
screenshot from 2019-01-03 21-10-44

Couldn't find any issue reported for this. 

### Proposed solution

Extend delete mixin
screenshot from 2019-01-03 21-14-19

### Testing Done

https://github.com/revathskumar/bulma-tag-fix 

([codesandbox seems to not work with github dependency](https://codesandbox.io/s/github/revathskumar/bulma-tag-fix)).

### DIFF
--- a/sass/elements/tag.sass
+++ b/sass/elements/tag.sass
@@ -69,6 +69,7 @@ $tag-delete-margin: 1px !default
   padding-right: 0.75em
   white-space: nowrap
   .delete
+    @extend %delete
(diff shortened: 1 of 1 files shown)


revathskumar:
@jgthms ping.

PR #2309, opened 2019-01-08
Hide burger menu after tap on phones and tablets

munrocket:
This is a **documentation improvement**.

### Proposed solution

Current js example in documentation doesn't hide target menu after tap on touch devices by default. Also 95% of bulma users have only one burger menu on web page and current solution is overcomlicated IMHO.

### Tradeoffs
If you have many hamburgers you need to wrap this solution in cycle by yourself.

### Testing Done
Tested on pc and mobile device in online.

### DIFF
--- a/docs/documentation/components/navbar.html
+++ b/docs/documentation/components/navbar.html
@@ -562,27 +562,26 @@ meta:
 {% capture navbar_js_code %}
 document.addEventListener('DOMContentLoaded', () => {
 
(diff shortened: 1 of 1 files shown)


jgthms:
AFAIK, this only works for the first `.navbar-burger` you'll find, not all of them?

munrocket:
K.I.S.S. in action :D

PR #2314, opened 2019-01-13
Add Django Simple Bulma to Related Projects.

heavysaturn:

### Intention
Add [Django Simple Bulma](https://github.com/python-discord/django-simple-bulma) to the Related Projects table in `README.md`. Provide some kaizen for the other lines in that table to make all the pipes line up and be more easily readable in raw form. Removing a handful of trailing periods so style  is more consistent across all descriptions. 

I'd be happy to revert the kaizen commit if you prefer to just add the project and do nothing else (I understand it might mess up git blame) - I've kept the commit separate so this will be very easily done.

image

### Testing Done
Viewed in PyCharms Markdown renderer to ensure it renders correctly. Tested URL.

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,40 +75,40 @@ Browse the [online documentation here.](https://bulma.io/documentation/overview/
 
 ## Related projects
 
-| Project                                                                            | Description                                                        |
-|------------------------------------------------------------------------------------|--------------------------------------------------------------------|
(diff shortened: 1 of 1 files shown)


jgthms:
Ok thanks for explaining the cleanup. I’ll have a look when I’m on my laptop.

Issue #53, opened 2016-02-08
Accessibility Issues


viion:
Hello,

There are some major accessibility issues in this project, I am wondering if this would ever be solved, some could be implemented quite easily:

- Roles missing, eg: Notifications require `role="alert"`
- Aria actions missing, eg: labels on close buttons
- Colors too washed out, default fail contrast accessibility (Including `is-dark` tags)
- HTML syntax is invalid, eg: `<a class="button">Button</a>` - Links are not buttons

Some would probably fall onto the user (eg the html syntax, as this is a CSS framework?)

Any plans to address some of these, at least the colours, providing an extra tag, or would it be down to the user to modify the `variables.sass` file?

jgthms:
Yes, accessibility has not been tackled at all for now.

Most accessibility features are tied to the HTML markup, not the CSS. So it goes beyond the scope of Bulma, which is a _CSS_ framework after all.

I like the idea of the high contrast flag that could be added in the variables. We could have functions to calculate those colors (like the `findColorInvert` function that's already being used).

If you want to update the HTML markup of the docs, feel free to submit a PR on the relevant repo: https://github.com/jgthms/bulma-website

I'll leave this issue open for discussion.

jwdallas:
Hey, a number of these controls do not have focus styles. You seem to have disabled the default focus styles and forgot to add back in your own. This is definitely an accessibility issue within the scope of a CSS framework.

Also super easy to fix. :)

johncaseysmith:
I would argue against a tag specifically for high contrast, since the actual problem is that the colors don't even meet the minimum requirements for contrast as they are. So really, the "high contrast" should be the _standard_. Regular text **must** have a contrast ratio of 4.5:1 to comply with the [WCAG 2.0 standards](https://www.w3.org/TR/WCAG20/#visual-audio-contrast-contrast).

And I second the focus styles being important and easy. [Take care with outlines](http://www.outlinenone.com/) and[ take care with that tab panel](http://heydonworks.com/practical_aria_examples/#tab-interface). You have a huge opportunity here to improve the accessibility of every website using this framework, I hope it isn't squandered! A great feature would be [CSS markup based on role attributes](http://www.webmonkey.com/2011/01/styling-webpages-with-arias-landmark-roles/).

chmanie:
> Most accessibility features are tied to the HTML markup, not the CSS. So it goes beyond the scope of Bulma, which is a CSS framework after all.

[...]

PR #2321, opened 2019-01-18
make focus behave like hover for the navigation.

ahus1:
Using :focus-within for dropdown menu.

This is a **improvement**.

The navigation bar behaves with the keyboard (tab+focus) the same like when using the mouse. Before the change the dropdown didn't work with the keyboard, now it works.

### Proposed solution

A dropdown menu will expand when using the keyboard (didn't expand before).
The highlighting on focus will work like on hovering

### Tradeoffs

None that I can think of. No additional classes and coupling. No JavaScript necessary.

### Testing Done

I've changed the sass file and tested it locally with the Bulma documentation. I can now tab trough the navigation. In Chrome there is a blue standard highlighting for the active menu item when I tab trough (standard browser behavior). The Menu items focused now appear with light grey background - same as hovering. When I reach the "More" menu item, is expands. I can also use shift-tab to move back through the menu. 

Screenshot of test attached.
bulma_menu

I've used the same changes on my personal homepage.

### DIFF
--- a/sass/components/navbar.sass
+++ b/sass/components/navbar.sass
@@ -69,6 +69,7 @@ $navbar-breakpoint: $desktop !default
           color: $color-invert
         & > a.navbar-item,
         .navbar-link
+          &:focus,
           &:hover,
           &.is-active
             background-color: darken($color, 5%)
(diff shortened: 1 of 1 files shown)


ahus1:
This might remedy some accessibility issues regarding the menu - #53

jgthms:
Nice one! Thanks.

Issue #1325, opened 2017-10-17
Search for Bulma docs


ioan-cristea:
Please add a search functionality for Bulma docs.

dguo:
Algolia's free [DocSearch](https://community.algolia.com/docsearch/) is an option.

gusmd:
I second this. I'm just starting out with Bulma (and web dev, for that matter), and looking through examples. Trying to find the specific classes and modifiers on the documentation right now is a pain. There are several options for adding searching to Jekyll-based docs.

midzer:
As Jekyll is used for build tool, we could use https://github.com/christian-fei/Simple-Jekyll-Search to implement internal search functionality easily ourselves.

patrickdaze:
I’ve done some testing with both [DocSearch](https://community.algolia.com/docsearch/) and [jekyll-algolia](https://community.algolia.com/jekyll-algolia/). I’m leaning towards DocSearch since it provide out-of-the-box functionality for scraping, indexing, and understanding the documentation’s hierarchy.

For example the following searches are able to display the section, and subsection of the documentation (examples are for testing purposes — please ignore the design aspect for now).

![dosearch-example](https://user-images.githubusercontent.com/250287/35160069-1b09df06-fd0a-11e7-836a-fef29366e7ee.png)

If we go forwards with DocSearch this is what I recommend:

1. [ ] Create the initial DocSearch configuration for testing (almost done, will update soon with link)
2. [ ] Create the initial frontend implementation for testing (almost done, will update soon with link)
3. [ ] Discuss, and adjust frontend design
4. [ ] Implement search versioning (possibly with http://versions.bulma.io/)
5. [ ] Update repo's documentation with search topic (if required)
6. [ ] Need site owners (Bulma team / @jgthms?) to [initialize conversation with Algolia](https://community.algolia.com/docsearch/)

Anyone agree, or disagree?

patrickdaze:
The latest frontend code can be found on [my fork](https://github.com/patrickdaze/bulma/tree/feature-1325-search). For testing locally, [append this to your Jekyll configuration](https://github.com/patrickdaze/bulma/commit/b5363482ef7b71018bde2b0684a10cb3cf229afb#comments). Feel free to discuss, or do a PR on my repo if you have any design ideas.

![2018-01-23 22_04_47](https://user-images.githubusercontent.com/250287/35313495-a9a3d794-008e-11e8-8b2b-5d9727780faa.gif)

[...]

PR #2329, opened 2019-01-23
Add print modifier and utility classes (fixes #721)

### LINKED ISSUE #721: Add print utility classes
### Overview of the problem

I'm using Bulma **version** [0.4.1]

### Description

It would be nice to
 * have response helpers for the print media in order to hide/show elements
 * be able to determine the sizing of elements with the `.is-([0-9]+)-print` classes

It might also be reasonable to hide some elements in print styles like buttons and navs by default but that may be to invasive, unintuitive and in my experience a few `.is-hidden-print` classes are easily added.

I’m not familiar with sass, but in case you like the idea of having print utility classes i might be able to put together a pull request.

o8e:
This is a **new feature**.

### Proposed solution
There is currently no way to modify the page layout based on the media type (specifically, print). This essentially adds modifiers to columns, and provides a helper class to hide when the print type is in use.

### Tradeoffs
N/A

### Testing Done

✅ Build project
✅ Test the added classes in a real scenario

bulma-print

### DIFF
--- a/sass/base/helpers.sass
+++ b/sass/base/helpers.sass
@@ -258,6 +258,10 @@ $displays: 'block' 'flex' 'inline' 'inline-block' 'inline-flex'
   .is-invisible-fullhd
     visibility: hidden !important
 
++print
+  .is-hidden-print
+    display: none !important
+
 // Other
 
 .is-marginless
--- a/sass/grid/columns.sass
+++ b/sass/grid/columns.sass
@@ -426,6 +426,63 @@ $column-gap: 0.75rem !default
         width: percentage($i / 12)
       &.is-offset-#{$i}-fullhd
         margin-left: percentage($i / 12)
+  +print
+    &.is-narrow-print
+      flex: none
+    &.is-full-print
+      flex: none
+      width: 100%
+    &.is-three-quarters-print
+      flex: none
+      width: 75%
+    &.is-two-thirds-print
+      flex: none
+      width: 66.6666%
+    &.is-half-print
+      flex: none
+      width: 50%
+    &.is-one-third-print
+      flex: none
+      width: 33.3333%
+    &.is-one-quarter-print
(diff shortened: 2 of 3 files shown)


jgthms:
Nice one! I’ll have a look at it.

jgthms:
I like the `is-hidden-print` helper and the `=print`. We should get that.

I'm not sure about all the print column classes. I think > 90% will never use it.
What we can do however is have a class to disable Flexbox for some columns:

```sass
.columns
  +print
    &.no-print
      display:block
```

This would disable columns on print, kind of like in your gif, and keep everything vertical.

mpereira:
@o8e thanks for submitting this PR.

@jgthms fwiw, I'd use the print column classes.

Maxhodges:
@jgthms attention please  #1325

mlncn:
https://github.com/jgthms/bulma/pull/2329#issuecomment-459391961 is what i expected Bulma to do already, disabling columns on print (it seems columns are really problematic for pagination, causing everything to be on separate pages).

I think a sensible default so that Bulma sites look good printed for the 95% of people who won't even touch stylesheets is good, but i think that the print column classes should be available and override, because for the few of us who do want to actively adjust how things are printed, we'll need them, or something more than is-hidden-print, most of the time!

smisaacs:
we'd use columns as well. +1

suterma:
Late, and surprised that we still do not have this yet. Seems like a simple and unobtrusive solution. If this is too much, I really suggest doing at least something like "is-hidden-print". This would really help in many cases. 

I will do this in my own addition now, in the meantime.

suterma:
By now I have created this small NPM package:
https://www.npmjs.com/package/bulma-print
to use as a stand-in until printing support gets into Bulma proper. Suggestions, Comments and contributions are welcome!

Issue #2331, opened 2019-01-24
Bulma Modal DOCS don't make it clear that you need to place `is-clipped` on body


DropsOfSerenity:
This is about **Bulma | the Docs**.

### Overview of the problem

This is about the Bulma **Docs**

### Description

The Docs for bulma modal should make it clear that you need to add `is-clipped` class on a containing element in order to prevent scrolling.

james-wasson:
This has been fixed. Directly from the docs:

> To activate the modal, just add the `is-active` modifier on the `.modal` container. You may also want to add `is-clipped` modifier to a containing element (usually html) to stop scroll overflow.

PR #2332, opened 2019-01-24
Clarify modal documentation.

DropsOfSerenity:

### Proposed solution
Modal documentation did not clarify that `is-clipped` should be placed on containing elements, even though modal examples on the page place `is-clipped` on the `html` tag.


DropsOfSerenity:
Closes #2331

DropsOfSerenity:
@jgthms changed to html

PR #2341, opened 2019-01-30
Add -ms-flexbox vendor prifix for better IE10 support

yanglin5689446:
This is an **improvement**.
Add `-ms-flexbox` vendor prefix to `is-flex` related class 

### Proposed solution
According to [mozilla developer docs](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Flexible_Box_Layout/Mixins) and [caniuse.com](https://caniuse.com/#search=flex), IE10 supports `flex` partially with `-ms-flexbox` vendor prefix, and this is how [bootstrap 4](https://github.com/twbs/bootstrap/blob/v4-dev/dist/css/bootstrap-grid.css#L1012) do as a fallback with their `d-flex` helper class, so I suppose it's not a bad idea to add it to `is-flex` classes?

### DIFF
--- a/css/bulma.css
+++ b/css/bulma.css
@@ -1129,59 +1129,69 @@ a.has-text-danger:hover, a.has-text-danger:focus {
 }
 
 .is-flex {
+  display: -ms-flexbox !important;
   display: flex !important;
 }
 
 @media screen and (max-width: 768px) {
   .is-flex-mobile {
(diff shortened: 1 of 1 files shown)


yanglin5689446:
nvm... didn't notice that autoprefixer already did it

PR #2363, opened 2019-02-14
fix/update responsive breakpoints in docs

lesleslie:
This is a bugfix.

The responsive breakpoints in Bulma have changed but the docs have not been updated to reflect those changes.

From initial_variables.sass:

// Responsiveness

// The container horizontal gap, which acts as the offset for breakpoints
$gap: 64px !default
// 960, 1152, and 1344 have been chosen because they are divisible by both 12 and 16
$tablet: 769px !default
// 960px container + 4rem
$desktop: 960px + (2 * $gap) !default
// 1152px container + 4rem
$widescreen: 1152px + (2 * $gap) !default
$widescreen-enabled: true !default
// 1344px container + 4rem
$fullhd: 1344px + (2 * $gap) !default
$fullhd-enabled: true !default

So in the docs:

1024 -> 1088
1216 -> 1280
1408 -> 1472

docs/_data/breakpoints.json has been updated accordingly.

@jgthms a question about the comments in the responsiveness section of initial_variables.sass:

The commented parts say for example
 // 960px container + 4rem
which produces the old value of 1024. This line is not valid sass so it's not the old code. Is it just a reference to the old code or should the 4rem actually be 8rem to match the (2 * $gap) - thus producing the current 1088?

Maybe these comments should just removed because they are confusing and unnecessary.

I would love to include updates to those comments in this PR as well depending on what the recommended course of action is for either updating or removing them.

-- les

### DIFF
--- a/docs/_data/breakpoints.json
+++ b/docs/_data/breakpoints.json
@@ -9,23 +9,23 @@
     "id": "tablet",
     "name": "Tablet",
     "from": 769,
-    "to": 1023
+    "to": 1087
   },
   "desktop": {
     "id": "desktop",
     "name": "Desktop",
-    "from": 1024,
-    "to": 1215
+    "from": 1088,
+    "to": 1279
   },
   "widescreen": {
     "id": "widescreen",
     "name": "Widescreen",
-    "from": 1216,
-    "to": 1407
+    "from": 1280,
+    "to": 1471
   },
   "fullhd": {
     "id": "fullhd",
     "name": "FullHD",
-    "from": 1408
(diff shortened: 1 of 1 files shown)


JonOGriffin:
+1, $gap is [defined in the documentation](https://bulma.io/documentation/overview/responsiveness/#variables) as 32px, but as shown above is now set to default at 64px. Definitely caused some confusion until I saw that.

lesleslie:
I'm pretty sure the issue I was experiencing which changed the breakpoints has been addressed in 0.76

PR #2381, opened 2019-02-24
Make centered buttons have equal margin

james-wasson:
This is a **bugfix**.

#### Issue
When you have long buttons within **.buttons.is-centered** and a button moves to a lower line the alignment of each button is off.
Same issue with **.buttons.is-right**
#### Solution
Add equal margin to both sides of the button.

### Tradeoffs
We are using button margin value of **0.5rem** (**0.25rem** each side for **.is-centered**) that value could change so it might be beneficial to use a variable here instead.

### Testing Done
I have some js-fiddles with my tests:
* Fiddle before fix: https://jsfiddle.net/james_wasson/a9nmyzcg/
* Fiddle after fix: https://jsfiddle.net/james_wasson/9kfnz3b0/

shrink the screen to see the broken buttons under
* Test Buttons Group -> Text Case 1 -> Centered
* Test Buttons Is-Right -> Any test case

#### Before:
bulmabeforefix
#### After:
bulmaafterfix

### DIFF
--- a/sass/elements/button.sass
+++ b/sass/elements/button.sass
@@ -277,3 +277,7 @@ $button-static-border-color: $grey-lighter !default
         margin-right: 0.25rem
   &.is-right
     justify-content: flex-end
+    &:not(.has-addons) 
+      .button:not(.is-fullwidth)
+        margin-left: 0.25rem
+        margin-right: 0.25rem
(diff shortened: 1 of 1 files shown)


james-wasson:
Update pull request with new information about .is-right and .is-centered.has-addons

seb3s:
I've upgraded from previous version 0.6.0. This update unfortunately breaks buttons right alignement with forms for normal size buttons (which is the common case).

PR #2388, opened 2019-03-03
Add example on how to delete notifications with JS

ro31337:

### Proposed solution

This change adds example to delete notifications. It's automatically loaded on dom-ready, and sets `click` handler for all `.delete` elements of `.notification`s.

### Tradeoffs

It's example in JavaScript, there is no tradeoffs, because it's optional.

### Testing Done

Yes, it works. I tried it in my own application, and it handler works perfectly.


jgthms:
Thanks!

PR #2396, opened 2019-03-07
Update docs with webpack 4 example

ashleymoogle:
Hello! Extract-text-plugin is deprecated in webpack 4 and onward in favor of Mini-css-extract: [https://github.com/webpack-contrib/extract-text-webpack-plugin](https://github.com/webpack-contrib/extract-text-webpack-plugin)
So I edited your doc to include a webpack 4 example and indicated that the previous example is for webpack 3 and less

### DIFF
--- a/docs/documentation/customize/with-webpack.html
+++ b/docs/documentation/customize/with-webpack.html
@@ -228,7 +228,7 @@ Wrote CSS to /path/to/mybulma/css/mystyles.css
 <hr>
 
 {% include components/step.html
-  title="3. Create a webpack config (Webpack < 3)"
+  title="3. Create a webpack config (Webpack <= 3)"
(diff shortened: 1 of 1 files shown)


docs/documentation/customize/with-webpack.html
```diff
@@ -174,10 +228,17 @@
 <hr>
 
 {% include components/step.html
-  title="3. Create a webpack config"
+  title="3. Create a webpack config (Webpack < 3)"
```
paulovieira: Small type here. Should be `(Webpack <= 3)`

ashleymoogle: Fixed

jgthms:
Nice one!

PR #2403, opened 2019-03-12
Added support for buttons where 'a' tags were explicitly targetted

brendanmckenzie:
This is an **improvement**.

### Proposed solution

There are some components that explicitly target `a` tags where semantically they could be `button` tags.  This PR addresses these issues.

### Tradeoffs

This overrides the base `button` style.  I am not sure if that's desired or not.

### Testing Done

I have tested this in a couple of projects I work on and have experienced no ill side effects.


jgthms:
I see what you mean but removing the background and border of _all_ buttons is an accessibility issue. People using Bulma should still be able to have a native button.

PR #2425, opened 2019-03-26
WIP: Initial start of bulma dark theme

### LINKED ISSUE #2342: [Feature] Dark mode support
It'd be awesome if Bulma supported the `@media (prefers-color-scheme: dark)` media-query out of the box!

I'm aware this would require two whole colour pallet/themes so understand if this is more suited for a theme/port of Bulma!

jloh:
This is a **new feature**.

### Proposed solution

This adds a dark theme to Bulma via the `@media (prefers-color-scheme: dark)`. Its obviously far from done but I'm opening this as a WIP to see whether its something that could be merged in the future or whether it should be kept outside of Bulma.

### Tradeoffs

I don't believe this should add a huge amount to the buildtime or anything but it does mean a dark/light version of Bulma must be maintained.

### Testing Done

TODO

Closes #2342

### DIFF
--- a/sass/base/generic.sass
+++ b/sass/base/generic.sass
@@ -1,8 +1,10 @@
 $body-background-color: $white !default
+$body-background-color-dark: #17181c !default
 $body-size: 16px !default
 $body-rendering: optimizeLegibility !default
 $body-family: $family-primary !default
 $body-color: $text !default
+$body-color-dark: $text-dark !default
 $body-weight: $weight-normal !default
 $body-line-height: 1.5 !default
 $body-overflow-x: hidden !default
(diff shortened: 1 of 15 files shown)


jgthms:
Thanks for helping out. I think most websites won’t or don’t need a dark theme however. Like you said, this would need to be maintained as well.

I imagine a case where someone uses Bulma to build their business website. They customise the colors of the “light” theme and don’t realise that there is also an underlying “dark” theme. So they customise the website and one day a user complains that the website looks terrible. The reason would be that the customisation was done without the dark theme in mind, so it was never tested. And that user happens to be on MacOS with Safari with dark mode enabled. 

Anyway, I like the idea of this PR but it should exist as an extension.

joshtriplett:
I would love to see this available as an extension, both for use with `prefers-color-scheme: dark`, and as an option for sites to enable if they want to *always* use a dark theme (in which case it'd be nice to be able to omit the CSS for the non-dark theme).

septatrix:
I think it should be on by default but one should be able to force light and dark scheme by applying a `is-dark` or `is-light` class to the body (or another element which would apply the setting to all contained elements).

This was it would yield a great out of the box experience and developers wanting to only use one scheme could set the class on the body. Additionally this would enable developers to let users choose the theme and set the corresponding classes with javascript (or remove them if the user wants it to be dynamic).

PR #2427, opened 2019-03-28
Update tiles documentation

ldub:

It is unclear what is meant by "parent tile" (the containing HTML element or the `is-parent` tile), so this clarifies the documentation to say "containing tile" instead of "parent tile".


jgthms:
Usually when talking about the nesting of HTML elements, we talk about “parent” and “child” relationship for immediate elements, and “ancestor” and “descendant” for elements further apart in the nesting hierarchy.

Issue #2426, opened 2019-03-27
is-spaced documentation missing?


selketjah:
This is about **the Docs**.
This is a question/I need help.

### Overview of the problem

This is about the Bulma **Docs**

I am sure this issue is **not a duplicate**? 
I searched for `is-spaced` in the issues, but did not find anything mentioning it.

### Description

There is a class called `is-spaced` in the sass file of the component of navbar (https://github.com/jgthms/bulma/blob/master/sass/components/navbar.sass#L278)
and it is being used for the top navbar of the documentation too, but it is not documented (https://bulma.io/documentation/components/navbar/)
It adds extra padding to the navbar.

Is this missing from the documentation?

### Steps to Reproduce
Look at the source code of the top navbar of Bulma documentation:

<img width="977" alt="Screenshot 2019-03-27 at 14 23 22" src="https://user-images.githubusercontent.com/2250570/55079181-f578d300-509b-11e9-8cbb-76fa2be37da3.png">

### Expected behavior
Documented

### Actual behavior
Not documented

Related issue: https://github.com/Fulma/Fulma/issues/180

mi9dev:
Issue Fix merged as part of #2436. This can be closed.

PR #2430, opened 2019-03-30
Add is-spaced documentation #2426

mi9dev:

### Proposed solution

Added is-spaced documentation under documentation/modifiers/helpers/

### Tradeoffs

N/A

### Testing Done

Validated changes. Attached screenshot 
Bulma-#2426

### DIFF
--- a/docs/documentation/components/navbar.html
+++ b/docs/documentation/components/navbar.html
@@ -1280,5 +1280,24 @@ $(document).ready(function() {
   {% include examples/navbar-color.html color="white" light=true %}
 </div>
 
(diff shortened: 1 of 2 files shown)


tomhrtly:
`is-spaced` is a modifier for the navbar component **only** and not a global utility to be used for adjusting the padding on any element.

The two commits should be reverted and the modifier should be added to the [navbar.html](https://github.com/jgthms/bulma/blob/master/docs/documentation/components/navbar.html) docs file.

mi9dev:
Sure. Will do it

Issue #2402, opened 2019-03-11
List Component Missing from Bulma Docs List


dyllandry:
### Overview of the problem

This is about the Bulma **Docs**

### Description
The [docs page for list components](https://bulma.io/documentation/components/list/) is not present in the component list sidebar or tab menu.

mi9dev:
Per @jgthms, This component is not actually ready, and might be deprecated soon.

Jeremey, Can you please close this issue as well?

mi9dev:
#2434

mi9dev:
Per #2434, this issue can be closed.

PR #2434, opened 2019-04-01
Display list component in side menu and Tab menu  #2402 

mi9dev:
To present the component-list in sidebar or tab menu

### Proposed solution

Component-List data is added to show up link URL in side menu or  tab menu

### Tradeoffs

N/A

### Testing Done

Testing done. Now user can navigate to list component using side menu or tab menu

-mi9dev

### DIFF
--- a/docs/_data/links.json
+++ b/docs/_data/links.json
@@ -430,6 +430,12 @@
       "icon": "angle-down",
       "path": "/documentation/components/dropdown"
     },
+    "components-list": {
+      "name": "List",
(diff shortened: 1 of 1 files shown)


jgthms:
That component is not actually ready, and might be deprecated soon.

mi9dev:
#2402

kczx3:
> That component is not actually ready, and might be deprecated soon.

Interesting.  Seems like a pretty basic component that would be frequently used.  Is there a suggested alternative or is that in the works?

PR #2479, opened 2019-05-18
Added foreground (e.g., text) color to the footer

VishwaJay:
While setting up my own site, I was surprised to find that this wasn't an option!

This is an improvement to give better control over the coloration of the text in the footer, which often needs to be de-emphasized from the regular text. However, if I want a footer that stands out (which is the case on a few web sites I maintain), the option to alter the text color in SASS is something I would really like.

### Testing Done

I tested a recompile. It works as expected.

### Changelog updated?

No, though I'm not sure if it's significant enough, since there isn't much documentation on some SASS options.

### DIFF
--- a/sass/layout/footer.sass
+++ b/sass/layout/footer.sass
@@ -1,6 +1,8 @@
 $footer-background-color: $white-bis !default
+$footer-color: $text !default
 $footer-padding: 3rem 1.5rem 6rem !default
 
 .footer
   background-color: $footer-background-color
+  color: $footer-color
   padding: $footer-padding
--- a/sass/layout/footer.sass
+++ b/sass/layout/footer.sass
@@ -1,4 +1,5 @@
 $footer-background-color: $white-bis !default
+$footer-color: false !default
 $footer-padding: 3rem 1.5rem 6rem !default
 
 .footer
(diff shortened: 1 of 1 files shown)


jgthms:
Not a bad idea, although the footer already inherits the color of its ancestors, which is preferred in case you change the color of the `body` and want the footer to use that as well.

I think you could set `$footer-color` to `false` by default, and use this like [https://github.com/jgthms/bulma/blob/master/sass/elements/title.sass#L36](https://github.com/jgthms/bulma/blob/master/sass/elements/title.sass#L36):

```sass
  @if $footer-color
    color: $footer-color
```

VishwaJay:
I like your proposed solution!

However, you say it's "preferred" (which isn't false), yet the case on a large number of websites is to have a different foreground color in the footer, if they have one.

The fact remains, I was really, really surprised that it wasn't an option. Should I close this and make the change, or should I just close it and let you make it on the main?

jgthms:
Yeah that works. You just need to initialize the variable above with

```sass
$footer-color: false !default
```

VishwaJay:
Doing that. :)

jgthms:
You just need to indent the last 2 lines.

PR #2486, opened 2019-05-21
Remove duplicate position in button is-loading.

apolokaklab:
Remove duplicate position in button is-loading.

### DIFF
--- a/sass/elements/button.sass
+++ b/sass/elements/button.sass
@@ -230,7 +230,7 @@ $button-static-border-color: $grey-lighter !default
     pointer-events: none
     &::after
       @extend %loader
-      +center(1em)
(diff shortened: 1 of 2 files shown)


jgthms:
It loses the `!important` flag though.

jgthms:
This might break some setups though. I don't really mind the duplicate property since they're exactly similar (the second one has priority).

apolokaklab:
yes right, but I only see the center function used for the button, and of course in the future it will be used, so I add a $flag option.
- for add !important:
+center($weight,$height,true)
example for button: +center(1em, 0, true)

Note: I like bulma with a small size to any time, and I will look for and fix other duplicates :)

PR #2502, opened 2019-05-28
invert button spinner colour when focussed

glenndevenish:
When you click a button with `.is-loading` class, it has the `:focus` property. I am using a button with the following classes, which means the background colour is the same as the spinner.
`button is-black is-outlined is-loading`

This edit inverts the button colour when it is focussed.

This is an **improvement**.

### Testing Done

None.

### Changelog updated?

No.


jgthms:
It's already there: https://github.com/jgthms/bulma/blob/master/sass/elements/button.sass#L176

PR #2516, opened 2019-06-03
Add "Vue CLI" to "Customize"

burkeholland:

### Proposed solution

Adds instructions on how to include Bulma when using the Vue CLI.

### Tradeoffs

Assumes that the user already has a Vue CLI project and doesn't instruct them on how to set that part of it up.

### Testing Done

Locally

### Changelog updated?

No.

### DIFF
--- a/docs/_data/links.json
+++ b/docs/_data/links.json
@@ -25,7 +25,7 @@
     },
     "overview-customize": {
       "name": "Customize",
-      "subtitle": "Create your <strong>own theme</strong> with a simple set of <strong>variables</strong>",
+      "subtitle": "Create your <strong>own themes</strong> with a simple set of <strong>variables</strong>",
       "color": "purple",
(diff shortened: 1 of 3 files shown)


jgthms:
Fantastic!

johnwebbcole:
Is there an easy way to view these?  I'm interested in the content :-)

johnwebbcole:
Could you add a section on creating a vue-cli based component that uses Bulma.  Pointers like https://twitter.com/jgthms/status/1133415219823300608?s=20 would be very helpful!

burkeholland:
Hey @johnwebbcole!

Can you elaborate a tad more there? The tweet from Jeremy is about the doubling up of styles in components, but that shouldn't happen if we import into `main.ts`. If we show an example of a component, it would just be any Vue component using Bulma styles. Is that correct or am I missing something here?

Issue #2298, opened 2018-12-31
Be able to use <button> instead of <a> for navbar hamburger


hansott:
This is about **Bulma**.

### Overview of the problem

This is about the Bulma **CSS framework**.
I'm using Bulma **version** 0.7.2.
I am sure this issue is **not a duplicate**.

### Description

Be able to use `<button>` instead of `<a>` for navbar hamburger. Why? https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/anchor-is-valid.md

### Steps to Reproduce

Use `<button>` as navbar hamburger instead of `<a>`.

### Expected behavior

The `<button>` is styled like `<a>`.

### Actual behavior

<img width="957" alt="screenshot 2018-12-31 16 47 28" src="https://user-images.githubusercontent.com/3886384/50563400-b9dca000-0d1c-11e9-9b54-3db8bc21d7b3.png">

<img width="957" alt="screenshot 2018-12-31 16 47 42" src="https://user-images.githubusercontent.com/3886384/50563397-b6e1af80-0d1c-11e9-8b32-94d12e16593a.png">

mobayen:
you can do so
just add `button is-white` css classes to the button

gvdp:
@hansott Next time, you should include which browser you are using, this doesn't seem to happen in Chrome. Create a PR with a proposal for a solution once I remembered other browsers exist :-).

hansott:
@gvdp You're right! Thought it was a global issue... Thanks for fixing.

kekkler:
I'm seeing this issue.

MefhigosetH:
Hi there.

I recently faced the same jsx-a11y/anchor-is-valid error. My fix was:

```
<button className="navbar-burger" aria-label="menu" aria-expanded="false" data-target="primaryNavbar" href="#">
    <span aria-hidden="true"></span>
    <span aria-hidden="true"></span>
    <span aria-hidden="true"></span>
</button>
```

I remplaced the `<a>` element by `<button>`, remove the `role='button'` statement, and put `href='#'` at end. This changes pass the rule and keep functional the burger menu.

PR #2519, opened 2019-06-04
#2298 Enable using button in navbar-burger

gvdp:
This is a **improvement | bugfix ** for issue #2298 .

### Proposed solution

The `<button>` tag gets some default styles from the 'user agent stylesheet' from the browser, these make it appear different from the `<a>` tags used in styling the navbar-burger.

I add some rules which override these default styles when a `<button>` tag is present in a navbar-burger.

### Testing Done

Tested in 

Chrome

image

Safari

image

Firefox

image

### Changelog updated?

No.

### DIFF
--- a/sass/components/navbar.sass
+++ b/sass/components/navbar.sass
@@ -152,6 +152,8 @@ body
   color: $navbar-burger-color
   +hamburger($navbar-height)
   margin-left: auto
+  @at-root button#{&}:hover
+    border: none
 
 .navbar-menu
   display: none
--- a/sass/components/navbar.sass
+++ b/sass/components/navbar.sass
@@ -152,8 +152,16 @@ body
   color: $navbar-burger-color
   +hamburger($navbar-height)
   margin-left: auto
-  @at-root button#{&}:hover
+  @at-root button#{&}
+    -webkit-appearance: none
+    -moz-appearance: none
     border: none
+    &:not(:hover)
+      background-color: inherit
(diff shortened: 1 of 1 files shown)


kekkler:
Can we merge this?

coolmic:
Merge when?

derykmarl:
Thanks!  Exactly what I was searching for.  The only downside is we're jumping through these hoops for people with accessibility needs (rightly so) but when you're actually pressing tab to cycle through items this means there's no visual cue when the focus is resting on the burger menu - this removes the warning but reintroduces difficulties for keyboard users and thus negates half the reason we started on this journey of making the burger menu a button in the first place :)

On the other hand, letting the focus outline appear looks a bit ugly when you clicked on it.

Eventually, hopefully Chrome will adopt [focus-visible](https://css-tricks.com/keyboard-only-focus-styles/) in order to make the outline only appear when focused with the keyboard (you can still use this method if you don't mind using a polyfill.  Personally I'd rather use them sparingly)

I'm new to quite a lot that's going on here, but for me (with a dark navbar) I reached a reasonable compromise by adding one more line to the focus selector in your patch: 
`background-color: darken($navbar-background-color, 3%)`
This just makes it a touch darker than the hover when selected with tab key or clicked on.  Downside is it may be a bit subtle if the user also has a visual impairment, but it's a compromise that I read gmail uses, for one.

PR #2522, opened 2019-06-05
Add .is-title-adjacent helper -- support for arbitrarily moving elements closer to titles (Fixes #2346)

### LINKED ISSUE #2346: FR: Add modifier class to allow any arbitrary element to appear closer to a title (not just subtitle)
This is about **Bulma**.

### Overview of the problem

I'm using Bulma **version** 0.7.2
My **browser** is: Safari 12.0.2 on macOS Mojave 10.14.2

I am sure this issue is **not a duplicate**.

### Description

Right now, when you have a title and a subtitle touching, the subtitle's margin-top will automatically reduce itself so that they are closer together. However, if you want to have supporting information close to a title, like tags, or a byline that would not make sense to be spaced as far as a normal paragraph would, you currently have to manually create that CSS.

A new modifier class that could be attached to any arbitrary number of items following a title-classed element that would make those elements appear closer to the title (or, indeed, a title + a subtitle) would help alleviate a lot of other manual work that would have to be done to, for instance, easily put things like tags or other elements underneath a title without technically making them subtitles.

purrloftruth:
This is a **new feature**.

Hi. This is my first time opening a pull request in Bulma! 👋🏻 The .is-title-adjacent helper adds the ability to arbitrarily move any element which is below a title or a subtitle closer to that title or subtitle, in the same way that a subtitle automatically moves closer to a title.

### Proposed solution

Some websites like to have things like bylines, post datetimes, or other things kept close to the post title area. This helper makes it easier to do this in Bulma without writing custom code.

It fixes Issue #2346, which I filed in February.

### Tradeoffs

The way this is written, it shouldn't impact those who have already written custom code to do this same thing, so that's thankfully *not* a tradeoff here. I don't see any reason why it adds any significant tradeoffs, though, of course, for more specific use cases, some users will continue to create their own custom classes.

### Testing Done

The changes I've made are small, so they seem to work as intended. As I'm a first-time contributor, I wanted to submit this pull request with the intent of making sure that there isn't something I'm overlooking, but given the tests done both on my personal website and while updating the docs, things seem to look good.

On my personal website, before the changes:
Screenshot 2019-06-05 12 58 57

My personal website, after the changes:
Screenshot 2019-06-05 12 55 56

Screenshot of my modifications to the documentation:
Screenshot 2019-06-05 13 49 20

### Changelog updated?

Yes, but not with PR # since I don't have one yet.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Upcoming release
 
+### New features
+
+* Add .is-title-adjacent helper
+
 ### Bug fixes
 
 * Fix #2031, Fix #2483 -> Invalid output when declaring a custom shade map
--- a/docs/documentation/elements/title.html
+++ b/docs/documentation/elements/title.html
@@ -59,6 +59,20 @@ meta:
 <p class="subtitle is-5">Subtitle 5</p>
 {% endcapture %}
 
+{% capture adjacent %}
+<p class="title is-1">Title 1</p>
+<p class="subtitle is-3">Subtitle 3</p>
+<p class="is-size-5 is-title-adjacent">Written by <strong>Jane Doe</strong></p>
+<div class="tags is-title-adjacent">
+  <span class="tag is-primary">Tag 1</span>
+  <span class="tag is-primary">Tag 2</span>
+  <span class="tag is-primary">Tag 3</span>
+</div>
+<div class="content">
+  <p>A few days ago, I thought strongly about taking a walk down the road. That thought led me into a journey into my mind, which in the end, ended up almost as strenuous as that walk would have been on its own!</p>
+</div>
+{% endcapture %}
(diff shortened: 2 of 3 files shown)


purrloftruth:
I'd like to hear what people think about the helper's class name. 

Also, as it works now, it still works with is-spaced -- while subtitles don't need an is-spaced helper, each element with an is-title-adjacent helper does. I'm unsure if this is of much utility, but maybe someone else can think of something that this would be suited for. Something else that I *can* potentially see a use for is a separate helper that simply adds a margin-bottom $block-spacing to an element, which has the same effect as doing this without all the hassle and potential weirderies/side-effects. To my knowledge, at least, a helper like that doesn't exist yet. Should I remove the is-spaced support? Should I potentially add in this separate helper?

PR #2538, opened 2019-06-21
no classes for margin / padding

dlishtvan:

### Proposed solution
add rules for margin/padding, user can use it as className (margin-bottom-xs, margin-auto, margin-none...)

### Changelog updated?

No.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,6 @@ _site
 dev
 node_modules
 test/output/
+branch_structure.json
+temp_auto_push.bat
+temp_interactive_push.bat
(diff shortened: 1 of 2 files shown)


nicostouch:
I too was surprised at a lack of utility classes around padding. 

I would propose adding:
- is-paddingless-left
- is-paddingless-right
- is-paddingless-top
- is-paddingless-bottom 

If I get time later I might open a PR.

.gitignore
```diff
@@ -15,3 +15,6 @@ _site
 dev
 node_modules
 test/output/
+branch_structure.json
```
guest20, line 18: What are those?

guest20:
Does this need a patch in `/docs/` too?

PR #2416, opened 2019-03-19
Add Visited Link Variable

curiouscrusher:
Included link visited variable for styling visited links.

This is a **bugfix**.

### Proposed solution

It fixes #2415 where `$link-visited` does not appear to be used. 

### Tradeoffs

None

### Testing Done

- [x] Pull the latest `master` branch
- [x] Make sure your Sass code is compliant with the [Bulma Sass styleguide](https://github.com/jgthms/bulma/blob/master/.github/CONTRIBUTING.md#bulma-sass-styleguide)
- [x] Make sure your PR only affects `.sass` or documentation files
- [x] [Try your changes](https://github.com/jgthms/bulma/blob/master/.github/CONTRIBUTING.md#try-your-changes).

Confirmed visited links show as the default `$purple` variable or as any color set in a customization override as explained in the [docs](https://bulma.io/documentation/customize/variables/)

Issue #2415, opened 2019-03-19
Link Visited Variable Not Used


curiouscrusher:
This is about **Bulma**.

### Overview of the problem

This is about the Bulma **CSS framework**

I'm using Bulma **version** [0.7.4] _Via NPM_
My **browser** is: All

I am sure this issue **was previously reported approximately 1yr ago** in #1888 

### Description

It appears that `$link-visited` is [defined](https://github.com/jgthms/bulma/blob/50fb74d6c8e74772a5a31e94457bbf35462a092f/sass/utilities/derived-variables.sass#L56) but never [used](https://github.com/jgthms/bulma/blob/master/sass/base/generic.sass#L62) unlike `$link-hover`, which is both [defined](https://github.com/jgthms/bulma/blob/50fb74d6c8e74772a5a31e94457bbf35462a092f/sass/utilities/derived-variables.sass#L58) & [used](https://github.com/jgthms/bulma/blob/master/sass/base/generic.sass#L69). I found [no other reference](https://github.com/jgthms/bulma/search?q=%24link-visited&unscoped_q=%24link-visited) to the `$link-visited` variable throughout the SASS files. 

### Steps to Reproduce

1. Install Bulma v0.7.2
2. Add HTML markup with a link
```HTML
...
<a href="somegreatlink.com">Visit Me</a>
...
```
3. Customize the link variables
```SCSS
$link-visited: #CCCCCC;
$link-hover: #EFEFEF;
```
4. See only hover state change on link

### Expected behavior
Link `:visited` color to change.

### Actual behavior
No change in `:visited` color of link.

curiouscrusher:
Added PR #2416 that addresses this.

PR #2565, opened 2019-07-18
Feature/link pseudo selectors

gatherben:
This is a **improvement**.

As noted in https://github.com/jgthms/bulma/issues/2415 not all defined a tag pseudo-class colors are used. This PR adds support for all a tag psuedo-classes along with utility classes to emulate their functionality for testing purposes. Per @jgthms suggestion in https://github.com/jgthms/bulma/pull/2416 these enhancements have been placed behind a flag which is false by default, so there should be no change for existing users.

### Proposed solution

See description.

### Tradeoffs

The main tradeoff is whether these styles should be applied by default, as some users do not want to have to override these styles once in place. I believe that placing this enhancement behind a flag eliminates this issue.

### Testing Done

Did a test builds with the flag off and on. Verified that no additional css is rendered when the flag is off, and the expected css is rendered when it is on.

Yes, please see the testing done section for details.

### Changelog updated?

No. If I'm reading this correctly this should only be done for new features. However, I have updated the docs to reference the new flag.


gatherben:
@jgthms would you mind having a look at this? I'd love to have support for this going forward.

PR #2587, opened 2019-08-05
Make fields fullwidth in panel-blocks (Fixes #2501)

### LINKED ISSUE #2501: I want to use "field has-addons" in panel-block
This is about Bulma.
I'm using Bulma v0.7.4.

About panel block definition
The following is defined, but it does not apply width(100%) when using "field has-addons".

```css
.panel-block > .control {
  flex-grow: 1;
  flex-shrink: 1;
  width: 100%;
}
```

Similarly, please define the following.

```css
.panel-block > .field {
  flex-grow: 1;
  flex-shrink: 1;
  width: 100%;
}
```

thx.

Geo1088:
This is a **bugfix**. (#2501)

### Proposed solution

Currently, when using a `.field` as a child of a `.panel-block`, the `.field` doesn't take up the full width available to it. This causes the `.is-expanded` modifier to do nothing on controls within the `.field`, which is undesirable.

This behavior applies the same styles to `.field`s as to `.control`s, which already work as intended in this setup.

### Tradeoffs

None that I can think of, as fields without any `is-expanded` controls will behave the same as before.

### Testing Done

None, WIP.

### Changelog updated?

No, WIP.

### DIFF
--- a/sass/components/panel.sass
+++ b/sass/components/panel.sass
@@ -76,7 +76,8 @@ $panel-icon-color: $text-light !default
   padding: 0.5em 0.75em
   input[type="checkbox"]
     margin-right: 0.75em
-  & > .control
+  & > .control,
+  & > .field
     flex-grow: 1
     flex-shrink: 1
     width: 100%


Geo1088:
Could I get some eyes on this? Should be a simple fix.

Geo1088:
@jgthms anything? This shouldn't require much review, it's just a one-line change.

Geo1088:
@jgthms Bump, should be an easy merge and it's been a while... Would love to have this fixed for a couple projects I'm working on.

PR #2615, opened 2019-09-04
Fix Textarea

apolokaklab:
Before:
Fix1

After:
FIX2

This on Firefox latest and Chromium latest with Kali Linux.

### DIFF
--- a/sass/form/input-textarea.sass
+++ b/sass/form/input-textarea.sass
@@ -14,6 +14,7 @@
       &.is-focused,
       &:active,
       &.is-active
+        outline: none
         box-shadow: $input-focus-box-shadow-size rgba($color, 0.25)
   // Sizes
   &.is-small
@@ -31,6 +32,7 @@
     width: auto
 
 .input
+  @extend %control
   @extend %input-textarea
   &.is-rounded
     border-radius: $radius-rounded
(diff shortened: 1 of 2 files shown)


sass/form/shared.sass
```diff
@@ -26,7 +26,6 @@ $input-icon-active-color: $grey !default
 $input-radius: $radius !default
 
 =input
-  @extend %control
```
jgthms: Why move this? It's already applied:

```sass
%input-textarea
  @extend %input
```

sass/form/input-textarea.sass
```diff
@@ -31,6 +32,7 @@
     width: auto
 
 .input
+  @extend %control
```
jgthms: It's already in `%input-textarea`.

sass/form/input-textarea.sass
```diff
@@ -45,6 +47,10 @@
 
 .textarea
   @extend %input-textarea
+  -webkit-appearance: none
+  box-shadow: none
+  border-width: 1px
+  border-style: solid
```
jgthms: You're duplicating a lot of properties. These are already applied by `%input-textarea`.

apolokaklab:
@jgthms Because this bug is on `utilities/controls.sass` but for this time only textarea, so I patch it for added textarea without `%control`, if the PR are less effective I will update it with other solution, maybe like pure textarea without class :)

PR #2629, opened 2019-09-18
Margin and padding system

Christian64:
Inspirated in bootstrap Padding and margin system, more friendly for boostrap users that want to make the change to Bulma.
Margin and padding system take the values of is-size-

This is a **new feature 

### Proposed solution

Alternative way (The more you choose the better )  to spacing and friendly

### Testing Done

Yes.

### DIFF
--- a/css/bulma.css
+++ b/css/bulma.css
@@ -481,6 +481,236 @@ table th {
   overflow: hidden !important;
 }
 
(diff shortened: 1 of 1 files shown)


jgthms:
You can't edit the `.css` directly. You need to edit only the `.sass` files.

PR #2635, opened 2019-09-28
Fix index.js when Vimeo does not load

auscompgeek:
This is a **documentation bugfix**.

### Proposed solution

This fixes the homepage JS to not break spectacularly if Vimeo's JS doesn't load for any reason.

### Testing Done

None.

### Changelog updated?

No.

### DIFF
--- a/docs/lib/index.js
+++ b/docs/lib/index.js
@@ -8,7 +8,7 @@ document.addEventListener('DOMContentLoaded', function () {
   var introIframe = document.getElementById('introIframe');
   var npmClipboard = new Clipboard('#npmCopy');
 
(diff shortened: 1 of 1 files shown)


jgthms:
I don’t see the difference?

auscompgeek:
The add/remove column buttons don't work at all if Vimeo is blocked.

jgthms:
Ok I thought `Vimeo` was equivalent to `window.Vimeo`.

The file you changed is generated automatically. Can you change `/docs/_javascript/index.js` instead, and run `npm run js-build` from the `/docs` folder?

PR #2638, opened 2019-10-01
Improvement disable resize in textarea disabled and ignore jekyll-cache

lucasferreiralimax:
This is a ** bugfix and documentation fix**.
This is an improvement for textarea disabled.
This is an improvement ignore .jekyll-cache

### Proposed solution
* Disabled textarea can resize, I disabled resizing with CSS.
* Ignore jekyll-cache in repository
Testing resize on textarea disabled

### Testing Done

Testing resize on textarea really disable resize
Testing resize on textarea really disable

### Changelog updated?

No.


jgthms:
I agree with removing `.jekyll-cache` but I'm not sure about removing the resize option. For example, the disabled textarea could be filled with useful information to the user, acting like a read-only version of the textarea. So the user might want to resize it.

Can you change this PR and only remove the `.jekyll-cache`?

lucasferreiralimax:
But is disabled, user is not access a feature with a disable
@jgthms  Thinking for to case disabled or not?

jgthms:
Yeah but they can still read it, even if it's disabled.

lucasferreiralimax:
@jgthms But also you see that with scroll

Follow example with resize disabled

Example

PR #2664, opened 2019-10-20
fix(panel): use $panel-colors variable

Gomah:
This is a **bugfix**.

When importing the panel sass file individually, it throws an error as the `$message-colors` variable is not defined.

### Proposed solution

 I think `panel.sass` should have the same logic as `message.sass` & declare a `$panel-colors` variable on top.

### Changelog updated?

No.


pokonski:
It's been two months since this fix was merged, do you have plans to release it @jgthms ?

PR #2665, opened 2019-10-20
Remove empty, archived repos from Related projects

connectkushal:

Cleaned up links with empty repos and archived projects.

### Changelog updated?

No.


jgthms:
I think read-only repos are still useful because the code is still there to be used. And you can create a fork and revive the project. Maybe just add a `(read-only)` label.

As for Vue Admin which looks dead, maybe add `(dead)`. But people can still find forks of the project and read the issues.

As long as the URL works, a project can still be useful.

connectkushal:
@jgthms done

PR #2679, opened 2019-10-30
Image tag generics rule duplicates minireset's

FMCorz:
This is an **improvement** that removes a rule duplicated rule between `generics` and `minireset`.

### Proposed solution

Minireset already declares the following, therefore the _generics_ rule for `img` is not needed, and can be removed.

```
img,
video
  height: auto
  max-width: 100%
```

### Tradeoffs

Custom builds that include `generics` and not `miniresets` will lose the `img` rule.

### Testing Done

Checked that the rule is no longer duplicated, but still present.

### Changelog updated?

No.

### DIFF
--- a/sass/base/generic.sass
+++ b/sass/base/generic.sass
@@ -93,10 +93,6 @@ hr
   height: $hr-height
   margin: $hr-margin
 
-img
-  height: auto
-  max-width: 100%
-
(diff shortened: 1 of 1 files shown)


jgthms:
Some people might not include the minireset for some reason, or use another reset. So I reckon we should keep this.

PR #2726, opened 2019-12-03
Navbar improvement

jhuettinger:
This is a **improvement**.

### Proposed solution

Currently the navigation menu dropdown can only be toggled via JavaScript. With the use of a checkbox it is possible to avoid using JavaScript which leads to faster load time and better performance. 
The current solution with toggling the `.is-active` class also works.

### Usage
1. Add an `input` with the attributes `type="checkbox" class="navbar-checkbox" id="navbar-checkbox"` on top of the navbar

2. Change the `<a>` tag of the navbar-burger to a `<label>` tag and add the attribute `for="navbar-checkbox"` 

It should look like this:
```
<nav class="navbar" role="navigation" aria-label="main navigation">
    <input type="checkbox" class="navbar-checkbox" id="navbar-checkbox">
    <div class="navbar-brand">

        <label class="navbar-burger burger" aria-label="menu" for="navbar-checkbox"
           data-target="navbarBasicExample">
            <span aria-hidden="true"></span>
            <span aria-hidden="true"></span>
            <span aria-hidden="true"></span>
        </label>
    </div>

    <div id="navbarBasicExample" class="navbar-menu">
        ...
    </div>
</nav>
```

### Tradeoffs

None.

### Testing Done

Testing succeeded with the proposed changes to the basic example from the bulma docs.

### Changelog updated?

No.


jgthms:
Thanks for the suggestion. I know about the checkbox technique but I'm not a big fan of adding CSS hacks to Bulma.

al222:
Hi, Why not at least add it to the documentation alongside the JavaScript solution, with some appropriate warnings? I was very happy to find it elsewhere, and was coming here make exactly the same suggestion than jhuettinger ;) It's always good to have a choice, and a lot of people come to Bulma because of the "no JavaScript" motto, pure CSS menus are really neat.
Since it's my first message, let me thank you also for your awesome work.

PR #2737, opened 2019-12-11
Fix #2471: Page padding for fixed spaced navbar

### LINKED ISSUE #2471: navbar is-fixed-top together with is-spaced incorrect padding for has-navbar-fixed-top
Its a bug

### Overview of the problem

 I'm using Bulma **version** [current]
My **browser** is: Chrome

### Description

navbar **is-fixed-top** together with **is-spaced** leads to incorrect padding

### Steps to Reproduce

1. Add is-spaced *and* is-fixed-top to navbar element classlist
2. Add has-navbar-fixed-top to body or html element
3. See what happens

### Expected behavior

That correct padding is applied to page, including the extra needed padding for is-spaced (1rem = 2 * $navbar-padding-vertical)

### Actual behavior

The extra padding introduced by the is-spaced class is not added either body or html

syn-zeta:
This is a **bugfix | documentation fix**.

- `has-spaced-navbar-fixed-top/bottom` is missing from documentation
- it only specifies padding for html/body on desktop and above

https://jsbin.com/buxaletika/edit?html,output
navbar

### Proposed solution
The solution here will apply correct padding to html/body by using only `has-spaced-navbar-fixed-top/bottom`.

### Tradeoffs
None that I can think of.

### Testing Done

navbar_fix

### Changelog updated?

Yes.

### DIFF
--- a/sass/components/navbar.sass
+++ b/sass/components/navbar.sass
@@ -124,9 +124,11 @@ $navbar-breakpoint: $desktop !default
 
 html,
 body
-  &.has-navbar-fixed-top
+  &.has-navbar-fixed-top,
+  &.has-spaced-navbar-fixed-top
     padding-top: $navbar-height
-  &.has-navbar-fixed-bottom
+  &.has-navbar-fixed-bottom,
+  &.has-spaced-navbar-fixed-bottom
     padding-bottom: $navbar-height
 
 .navbar-brand,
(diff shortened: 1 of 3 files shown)


sass/components/navbar.sass
```diff
@@ -124,9 +124,11 @@ $navbar-breakpoint: $desktop !default
 
 html,
 body
-  &.has-navbar-fixed-top
+  &.has-navbar-fixed-top,
+  &.has-spaced-navbar-fixed-top
     padding-top: $navbar-height
```
josemotanet: I just saw this and I thought if `$navbar-padding-vertical * 2` should be added? Thoughts?

josemotanet: To explain the suggestion above: when a navbar is fixed on top and it is spaced, content digs under because it won't consider the navbar's padding, regardless of whether it exists or not.

Does this make sense? Can I help?

a-kriya: Already done on lines 422 and 424 in this file below.

`is-spaced` only adds extra padding from `$navbar-breakpoint` which defaults to `$desktop` by default. So, the changes here add the regular padding for viewports under that for `has-spaced-navbar-fixed-(top|bottom)` class.

tomas-kulhanek:
is possible to merge or fixit?

a-kriya:
@jgthms Issue remains as of v0.9.3. Any pointers on how I can improve the PR?

PR #2740, opened 2019-12-14
Fix #2739 add variables size for layout hero

### LINKED ISSUE #2739: Hero Sizes should be configurable variables
Like done with section sass these hero sizes should be variables: 

https://github.com/jgthms/bulma/blob/5381f2649d3466487585a0f0dfa8f177c8f42b1d/sass/layout/hero.sass#L71-L85

See section sass:
https://github.com/jgthms/bulma/blob/5381f2649d3466487585a0f0dfa8f177c8f42b1d/sass/layout/section.sass#L1-L3

I can make a pull request on it if this seems like something many others would also want.

IvanFrescas:
This is an improvement|documentation fix.

Fixes issue #2739 adding three sass variables to hero layout

### Proposed solution
Screen Shot 2019-12-14 at 9 31 39

### Testing Done
It was tested locally and the layout hero is still working the same way as before.

### Changelog updated?

No.

### DIFF
--- a/docs/_data/variables/layout/hero.json
+++ b/docs/_data/variables/layout/hero.json
@@ -1,0 +1,25 @@
+{
+  "by_name": {
+    "$hero-padding": {
+      "name": "$hero-padding",
+      "value": "1.5rem 0rem",
+      "type": "size"
+    },
+    "$hero-padding-medium": {
+      "name": "$hero-padding-medium",
+      "value": "9rem 0rem",
+      "type": "size"
+    },
+    "$hero-padding-large": {
+      "name": "$hero-padding-large",
+      "value": "18rem 0rem",
+      "type": "size"
+    }
+  },
+  "list": [
+    "$hero-padding",
+    "$hero-padding-medium",
+    "$hero-padding-large"
+  ],
+  "file_path": "layout/hero.sass"
+} 
--- a/docs/documentation/customize/concepts.html
+++ b/docs/documentation/customize/concepts.html
@@ -14,7 +14,7 @@ breadcrumb:
 
 <div class="content">
   <p>
-    Bulma is highly customizable thanks to <strong>415 Sass variables</strong> living across <strong>28 files</strong>.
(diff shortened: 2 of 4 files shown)


jgthms:
So this is a breaking change because you're removing the horizontal padding on the `hero-body` defined [here](https://github.com/jgthms/bulma/blob/master/sass/layout/hero.sass#L143):

```sass
.hero-body
  padding: 3rem 1.5rem
```

I would actually created 4 variables:

```sass
$hero-body-padding
$hero-body-padding-vertical-small
$hero-body-padding-vertical-medium
$hero-body-padding-vertical-large
```

IvanFrescas:
Thanks for your answer, I'll work on it.

IvanFrescas:
What do you think of this solution? 
I also noticed that in the documentation there is no reference to the is-small class. If you want I can make another issue of that.

jgthms:
I'd keep `padding-top` and `padding-bottom` for the `small` `medium` and `large` versions. Otherwise, you're repeating `1.5rem` 4 times.

IvanFrescas:
I relied on the file `section.sass` where it also has the sizes for medium and large.
do you have any suggestions to avoid repeating `1.5rem`?

hammadzz:
> I'd keep `padding-top` and `padding-bottom` for the `small` `medium` and `large` versions. Otherwise, you're repeating `1.5rem` 4 times.

That would be cleaner.


> I relied on the file `section.sass` where it also has the sizes for medium and large.
> do you have any suggestions to avoid repeating `1.5rem`?

Just remove the 1.5rem from medium and large as it is an unnecessarily overriding the same value from hero-body. But I am new to SASS so @jgthms will have to confirm.

jgthms:
Actually you’re right. Doing like `section.sass` is fine actually. Let’s go with that.

IvanFrescas:
Awesome.
So, should I leave the changes as they are? Or should I modify something? 🤔

PR #2742, opened 2019-12-17
Update navbar.html

gautomdas:
Slight modification needed for nav-bar brand image example (forward slash).

### Proposed solution
Adding a forward slash for the image in the navbar-brand  examples.

### Tradeoffs
None.

### Testing Done
Checked local compilation. It is just a simple forward slash.

### Changelog updated?
No


jgthms:
In HTML5 the slash is not present.

PR #2743, opened 2019-12-17
Fixes #2741: Update form inputs derived variables

### LINKED ISSUE #2741: Can't use inherit value anymore since 0.8.0
This is about **Bulma**.

### Overview of the problem

I'm using Bulma **version** [0.8.0]

This is a **Sass** issue: I'm using version [0.8.0]

### Description

Before version 0.8.0, we were able to use `inherit` values for colors. I used to override `$text-strong` variable so it takes an inherited color instead of a hardcoded one.  

```
$text-strong: inherit
```

Since 0.8.0, using inherit value for this particular variable throws the following error when compiling with Sass:

```
ERROR in ./resources/sass/app.scss
Module build failed (from ./node_modules/css-loader/index.js):
ModuleBuildError: Module build failed (from ./node_modules/sass-loader/dist/cjs.js):

$input-placeholder-color: rgba($input-color, 0.3) !default
                         ^
      $color: inherit is not a color.
  ╷
6 │ $input-placeholder-color: rgba($input-color, 0.3) !default
  │                           ^^^^^^^^^^^^^^^^^^^^^^^
  ╵
  node_modules/bulma/sass/form/shared.sass 6:27  @import
  node_modules/bulma/sass/form/_all.sass 3:9     @import
  node_modules/bulma/bulma.sass 6:9              @import
  stdin 9:9                                      root stylesheet
```

I rolled back to Bulma 0.7.5 and Sass compiles without throwing any error.  

An issue has been posted about this more than two years ago. @wrabit suggested to use `inherit` as a workaround but this is not working anymore. https://github.com/jgthms/bulma/issues/821#issuecomment-539953034

### Steps to Reproduce

1. Override `$text-strong` variable with the value `inherit`
2. Compile with Sass

### Expected behavior

We should be able to use `inherit` values without Sass throwing an error.

### Actual behavior

Sass throws an error when compiling.

marcbelletre:
Use `$grey-darker` and `$grey` as default instead of `$text-strong` and `$text-light` which may contain the "inherit" value (fixes issue #2741)

This is an **improvement / bugfix**.

I ran into a compiling error with Bulma 0.8.0 when trying to use `inherit` value for the `$text-strong` variable. I found out this is due to breaking changes in commit 0f988eadd9cbca055966e4418a704b85e19999ff in which a lot of default colors have been modified. This particular line was throwing the error when using `inherit` value for `$text-strong`:

```
$input-placeholder-color: rgba($text-strong, 0.3) !default
``` 

I think we may often use this particular value for the `$text-strong` and `$text-light` variables because there are a lot of cases where a `<strong>` tag in a text should just inherit its parent color. 
Also I think it doesn't really make sense to use the strong text color as default in any other derived variable.

### Proposed solution

In this PR I just put back the old default values, replacing `$text-strong` by `$grey-darker` and `$text-light` by `$grey`. By the way this is what [the docs](https://bulma.io/documentation/form/input/#variables) actually says.

### Tradeoffs

There are still other derived variables which use the `$text-strong` and `$text-light` ones as default. All of these changes have been made in commit 0f988eadd9cbca055966e4418a704b85e19999ff.  
Maybe it could be a better idea to roll back the whole commit but I'm not really sure about that.

### Testing Done

I did test this PR in my project. Using `$text-strong: inherit` does not throw a compiling error anymore.

### Changelog updated?

No.


afwn90cj93201nixr2e1re:
@jgthms

jgthms:
The idea with using `$text-strong` is that you only need to update once if you want to make a dark mode for example.

The fix would be to instead handle the `rgba()` case where `inherit` is used.

tomekr:
@jgthms with the use of inherit (https://github.com/jgthms/bulma/blob/0.8.1/sass/form/shared.sass#L1) it looks like in a hero, the input will inherit the text color from a heading/subheading, causing the text to be white.

Screen Shot 2020-03-25 at 2 11 22 PM
Screen Shot 2020-03-25 at 2 11 25 PM

PR #2747, opened 2019-12-20
#2729 - Add level align modifiers

### LINKED ISSUE #2729: new class: align-bottom
This is about **Bulma**.
This is about a **FEATURE**
This is **not a bug**.

### Overview of the problem

Bulma is currently not opinioned on how to align elements.  This causes problems with layout tools like `.level-item` where a button is placed next to a series of horizontally laid out controls:

```css
.align-bottom {
    align-self: flex-end; /* Push level-item and following elements to the bottom */
} 
```

### Description

Bulma has no opinion on how to align items within a `level` layout rule.  This requires all Bulma users to then create a non-standard style name/approach to solving the problem (usually, searching Stack Overflow for "flexbox align button to bottom")

### Steps to Reproduce

```
<section class="section">
    <div class="container is-fluid">
        <form action="@Url.Action((SearchController x) => x.GetSearchResults(null))">
            <div class="level">
                <div class="level-left">
                    <div class="level-item">
                        <div class="field">
                            <label for="AssigneeDisplayName" class="label">Assignee</label>
                            <div class="control">
                                <input type="text" name="AssigneeDisplayName" />
                                <input type="hidden" asp-for="AssigneeSid" />
                            </div>
                        </div>
                    </div>

                    <div class="level-item">
                        <div class="field">
                            <label for="Status" class="label">Status</label>
                            <p class="control">

                                <input type="text" name="Status" />
                            </p>
                        </div>
                    </div>
                    <div class="level-item align-bottom">
                        <div class="field">
                            <p class="control">
                                <button type="reset" class="button is-warning">Clear</button>
                                <button type="submit" class="button is-link">Search</button>
                            </p>
                        </div>
                    </div>
                </div>
            </div>
        </form>
    </div>
</section>
```

### Expected behavior

Add more helpers

### Actual behavior

No helper available; needed to go to Stack Overflow to get help on best way to do this in flexbox, and the top voted suggestion `auto-to

abnersajr:
This is a **new feature**.

### Proposed solution

This aims to fix #2729 adding new 4 classes to help item alignment.
It uses `align-self` to determine items position.

### Tradeoffs

The only tradeoff is that this uses flexbox. So, there isn't a tradeoff.

### Testing Done

Tested locally to check if it works as expected.

### Changelog updated?

Yes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -224,6 +224,7 @@ Fix #1979 -> Correct loading spinner color when a button is:
 * #592 -> Give arbitrary elements access to the image/ratio classes
 * #1682 Fix #1681 -> Adds disabled styles for `<fieldset disabled>`
 * #2201 Fix #1875 -> `.buttons` and `.tags` group sizing (`.are-small`, `.are-medium`, `.are-large`)
+* #2747 Fix #2729 Adds item alignment classes
 
 ### Improvements
 
--- a/docs/documentation/modifiers/helpers.html
+++ b/docs/documentation/modifiers/helpers.html
@@ -71,5 +71,22 @@ breadcrumb:
       <td><code>is-relative</code></td>
       <td>Applies <code>position: relative</code> to the element.</td>
     </tr>
+    <tr>
+      <th rowspan="4">Item Alignment</th>
+      <td>align-start</td>
+      <td>The element is positioned at the <strong>beginning</strong> of the container</td>
+    </tr>
+    <tr>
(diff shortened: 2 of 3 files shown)


jzabroski:
Can we have this merged so I can close my request?  This seems to work fine.

jgthms:
Having “align-bottom” equal to “flex-end” is incorrect as it’s only true for a “flex-direction” of “row”. It’d be better to keep the actual CSS value in the name.

PR #2752, opened 2019-12-30
Add cursor helper

balping:
This is a **new feature**.

This PR adds a new helper `.has-cursor-pointer` which sets the cursor to pointer on a given element. I find myself creating similar helper classes all the time.

### Tradeoffs

None.

### Testing Done

None.

### Changelog updated?

No.

### DIFF
--- a/sass/base/helpers.sass
+++ b/sass/base/helpers.sass
@@ -279,3 +279,6 @@ $displays: 'block' 'flex' 'inline' 'inline-block' 'inline-flex'
 
 .is-relative
   position: relative !important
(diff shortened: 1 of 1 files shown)


nirnejak:
I believe it should be something like, so that it can work with disabled states
```sass
$cursors: ("default", "pointer", "not-allowed", "context-menu")
@each $cursor in $cursors 
  .has-cursor-#{$cursor} 
    cursor: #{$cursor}
    &:disabled
      cursor: not-allowed
```

PR #2759, opened 2020-01-01
fix bulma.sass path in documentation

tshirtman:
the bulma archive has bulma.sass directly at the root of the bulma-$version/ directory, not in a bulma subdirectory. contradicting the documentation at https://bulma.io/documentation/customize/with-sass-cli/

### Proposed solution

update the bulma_path variable

### Tradeoffs
can't think of any
### Testing Done

None.

### Changelog updated?

No.


jgthms:
Good catch. Thanks.

Issue #2760, opened 2020-01-04
[Feature Request] Circular Progress Bar


kkesley:
This is about **Bulma**.

### Overview of the feature request

This is about the Bulma **CSS framework**
I'm using Bulma **version** [0.8.0]

### Description

Would you consider adding a circular progress bar to the [current progress bar](https://bulma.io/documentation/elements/progress/)?

Maybe add a class `is-circle` to the element? e.g. `<progress class="progress is-circle" value="15" max="100">15%</progress>`

Example:

![circular-progress-bar-example](https://user-images.githubusercontent.com/14907878/71759756-0e08fa00-2f06-11ea-86bf-2aca00f69cd1.PNG)

suterma:
I have found a very great circular progress bar here: https://gist.github.com/nicbet/49a681d036ab8c2886e3419a6e8c6efb
It works well, and looks quite similar to the proposal here. Maybe this could get added more or less like in the gist?

PR #2775, opened 2020-01-18
Circular progress bar

francesco-loreti:
This is a new feature for issue #2760

### Proposed solution

This is my solution for circular progress bar required into issue #2760.
You can use `is-circle` with `process` class.
For example:
```
<div class="progress is-circle is-info has-text-info is-40">
<div class="hide"></div>
<div class="active"></div>
<div class="active2"></div>
</div>
```
After `is-circle` you can append all standard Bulma classes for text colors(has-text-$color), bar color(is-info, is-danger), font width(is-size-1), etc...
Default size of bar is 0. You can change this size with class `is-$dimension` from 0 to 100. 

This solution is not compatible with `<process>` tag.

### Tradeoffs

The issue reported was using the

### Testing Done

Manually tested with Mozilla Firefox

### Changelog updated?

No.

Any suggestions?


u89012:
Oh no! Is this feature abandoned? I've a need for this just about now!

PR #2776, opened 2020-01-18
#2760 - Add feature for circular-progress-bar

francesco-loreti:
This is a new feature for issue #2760

### Proposed solution

This is my solution for circular progress bar required into issue #2760.
You can use `is-circle` with `process` class.
For example:
```
<div class="progress is-circle is-info has-text-info" value="40">
<div class="hide"></div>
<div class="active"></div>
<div class="active2"></div>
</div>
```
After `is-circle` you can append all standard Bulma classes for text colors(has-text-$color), bar color(is-info, is-danger), font width(is-size-1), etc...
Default size of bar is 0. You can change this size with value tag attribute `value="40"` from 0 to 100. 

This solution is not compatible with `<process>` tag.

### Tradeoffs

The issue reported was using the

### Testing Done

Manually tested with Mozilla Firefox and Microsoft Edge.
immagine

### Changelog updated?

No.

Any suggestions?

### DIFF
--- a/sass/elements/progress.sass
+++ b/sass/elements/progress.sass
@@ -60,6 +60,129 @@ $progress-indeterminate-duration: 1.5s !default
   &.is-large
     height: $size-large
 
+  &.is-circle
+    $progress: 0
+    $color-active: #2ecc71
+    $color-inactive: #eee
+    $color-bg: $white
+    $color-font: #eee    
+
+    // background: $color-inactive
+    border-radius: 50%
+    position: relative
+    font-family: 'Lato', sans-serif
+    height: 150px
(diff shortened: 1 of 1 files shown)


francesco-loreti:
immagine

jgthms:
I don't think adding this circular progress bar is a good idea because it doesn't use Bulma's `<progress>` bar. This is like adding a whole other library to Bulma.

Also, you're adding 2,194 lines to the CSS! This increases the filesize of Bulma by 26%! For a niche feature like this, it's way too much and this will bloat Bulma unnecessarily.

PR #2778, opened 2020-01-19
Update options.html

tbmreza:
Typo.


jgthms:
“than” is actually correct because you’re saying “more than”.

PR #2788, opened 2020-01-31
Fix multiple selectors in ':not'

### LINKED ISSUE #2597: hero navigation color issue - only on safari
image

Worked at version 0.7.1 (not on 0.7.5)

Problem is the line https://github.com/jgthms/bulma/blob/master/sass/layout/hero.sass#L20

image

### LINKED ISSUE #2785: link color bug on safari
## Description

On Safari only, the link color is not correctly computed when inserted in a hero with some color modifier.

## To reproduce

```
<section class="hero is-info">
  <a>Hello world</a>
</section>

<section class="hero">
  <a>Hello world</a>
</section>
```

Result:

| Correct behavior (Chrome) | Wrong behavior (Safari) |
| --- | --- |
| correct behavior | wrong behavior |

I expect that the link color should be `$link-color` as on Chrome.

## Reason

When not working (on Safari only), the following rule is triggered:

https://github.com/jgthms/bulma/blob/master/sass/layout/hero.sass#L24-L26

I cannot explain why this occurs only on Safari, but for sure, inheriting the color (i.e. take the parent color) cannot lead to a good result here.

DCjanus:
This is a **bugfix**.

### Proposed solution

fix #2785
fix  #2597

### Changelog updated?

No.

### DIFF
--- a/docs/documentation/elements/icon.html
+++ b/docs/documentation/elements/icon.html
@@ -111,7 +111,7 @@ meta:
 
 <div class="content">
   <p>
-    The Bulma <code>icon</code> container comes in <strong>4 sizes</strong>. It should always be <em>slightly bigger</em> than the icon it contains. For example, Font Awesome 5 icons use a font-size of <code>1em</code> by default (since it inherits the font size), but provides <a href="https://fontawesome.com/how-to-use/svg-with-js#additional-styling" target="_blank">additional sizes</a>.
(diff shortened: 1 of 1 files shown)


jgthms:
Actually, `.pagination-link` should inherit because it has a transparent background. But `.is-current` shouldn't.

DCjanus:
I'm not frontend developer, would you mind fix this directly?

PR #2804, opened 2020-02-10
Notification.html - simpler JS to close notifications

alessio-signorini:

### Proposed solution
Simple javascript in the `onClick` attribute of the `delete` button to close notifications.

### Tradeoffs
None.

### Testing Done
Standard javascript, should work in all browsers. Works in last Chrome/Safari/Firefox at least.

### Changelog updated?
No.


alessio-signorini:
@jgthms - no interest?

jgthms:
I would add the example as an alternative instead of a replacement.

PR #2873, opened 2020-03-14
outdated cards with new word in it changed

kaangokdemir:

### Proposed solution

In website, 0.5.3 version features (which published 2 years ago) were still had `New!` tag with it. They were replaced into `Since` tags like most of 0.6.1 version features had.

### Tradeoffs

N/A

### Testing Done

None.

### Changelog updated?

No.

### DIFF
--- a/docs/documentation/components/card.html
+++ b/docs/documentation/components/card.html
@@ -134,10 +134,7 @@ meta:
   </ul>
 </div>
 
-<div class="tags has-addons">
-  <span class="tag">Since</span>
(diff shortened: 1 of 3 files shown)


jgthms:
Actually, you can use the `new-tag.html` HTML component like this:

```jekyll
{% include elements/new-tag.html version="0.7.2" %}
```

This will automatically handle the `New/Coming Soon/Since` text.

kaangokdemir:
Tags imported from `elements/new-tag.html` instead @jgthms

PR #2874, opened 2020-03-14
Fixes #2862

### LINKED ISSUE #2862: Fix example about Twitter
This is about **the Docs**.

### Overview of the problem

The Twitter brand guideline says We cannot use the "Twitter icon" with "Twitter Text".

https://about.twitter.com/content/dam/about-twitter/company/brand-resources/en_us/Twitter_Brand_Guidelines_V2_0.pdf

But you show an example of it.
https://bulma.io/documentation/elements/button/

This example leads us to do violating usage.
Can you fix it?

fmmricardo:
This is a improvement.

Why:
* According Twitter brand guideline says we cannot use the "Twitter icon" with "Twitter text".

This change addresses the need by:
* Replacing the span text from Twitter for @Bulma

### DIFF
--- a/docs/documentation/elements/button.html
+++ b/docs/documentation/elements/button.html
@@ -221,7 +221,7 @@ meta:
     <span class="icon">
       <i class="fab fa-twitter"></i>
     </span>
-    <span>@Bulma</span>
+    <span>@jgthms</span>
   </button>
   <button class="button is-success">
     <span class="icon is-small">


jgthms:
I’d put “@jgthms” or the verb “Tweet” because we don’t manage “@bulma”.

fmmricardo:
@jgthms ok sounds fair to me.

jgthms:
Thanks!

PR #2879, opened 2020-03-22
Remote duplicate code

nguyenyou:
This is an **improvement**.

The `.navbar-item` is already declared as `display: block` at line 162.

https://github.com/jgthms/bulma/blob/f75ae3db01cb3be96dde517811f4950540f4c76f/sass/components/navbar.sass#L162


jgthms:
Good catch. Thanks.

PR #2882, opened 2020-03-23
Fix #6 RTL support with Logical properties

### LINKED ISSUE #6: Don't forget about RTL direction please

jgthms:
Fixes #6.

This is a **new feature**.

### Proposed solution

A lot of help from https://www.rtlstyling.com/

This PR adds **RTL support** by making use of [CSS Logical Properties](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Logical_Properties), specifically:

* `margin-inline-start`
* `margin-inline-end`
* `padding-inline-start`
* `padding-inline-end`
* `border-inline-start`

Because other properties like `border-start-start-radius` or `inset-inline-start` are still poorly supported (< 75%), this PR also introduces 4 mixins:

* `=ltr` which only outputs content if `$rtl` is false
* `=rtl` which only outputs content if `$rtl` is true
* `=inset-inline-start($spacing)` as a replacement for `inset-inline-start`
* `=inset-inline-end($spacing)` as a replacement for `inset-inline-end`

And a new flag: `$rtl` set to `false` by default.

This means you can build a RTL version of Bulma by simply setting the flag to `true` and importing Bulma:

```sass
@charset "utf-8"
$rtl: true
@import "bulma"
```

As a result, 3 new CSS files exist:

* `bulma-rtl.css`
* `bulma-rtl.css.map`
* `bulma-rtl.min.css`

### Tradeoffs

Where `margin-right` was, there is now `margin-inline-end`. So compatibility is slightly reduced, but still > 90%.

Only properties highly supported by browsers were used directly.
image

### Testing Done

Adding `<html dir="rtl">` to pages and seeing the result.

### Changelog updated?

No.


Nemesis19:
Hello, any ETA on this feature? Thanks!

cN3rd:
Any updates on this? it wasn't updated since late march.

jgthms:
RTL has been merged through another PR.

PR #2887, opened 2020-03-25
removed  bower

nainardev:

### Proposed solution

Bower is deprecated see https://bower.io
removed it from readme

### Testing Done

None.

### Changelog updated?

No.


jgthms:
Thanks but while Bower is deprecated, it is still maintained. It doesn't cost much to keep it, and it's up to the user to choose npm/yarn.

PR #2896, opened 2020-04-02
removed bower

RDIL:
Bower is deprecated and shouldn't be used anymore.


jgthms:
Thanks but while Bower is deprecated, it is still maintained. It doesn't cost much to keep it, and it's up to the user to choose npm/yarn.

PR #2901, opened 2020-04-06
Using `auto` instead for `overflow-y` on html selector

dhonx:


jgthms:
It’s on purpose, to prevent horizontal “jumps” when navigating from a page higher than the viewport and one shorter.

PR #2941, opened 2020-05-13
documentation fix | form/general.html updated button label

ChiragSukhala:
The button label was out of sequence as compared to rest of buttons in the group, and was bugging me. 
I believe its the same with anyone reading the docs carefully.

### Changelog updated?

No.


jgthms:
Thanks,  good catch.

PR #2955, opened 2020-05-16
Fix issue when there's only one is-toggle tag

Neo-Zhixing:
This is a bugfix.

Currently tags.is-toggle was displayed incorrectly when there's only one list item in .tabs.is-toggle.

image

This is due to to border-radius shorthand overriding each other.
image

### Proposed solution

Replace border-radius short-hand with individual properties in .tags.is-toggle li:last-child a

### Tradeoffs
None.

### Testing Done

After making those changes, .tags.is-toggle with only one child can be rendered correctly.

image

### Changelog updated?

No.


jgthms:
Thanks.

Issue #2749, opened 2019-12-22
[Feature/Question] How to use CSS variables?


super2ni:
This is about Bulma.

### Overview of the problem

This is about the Bulma **CSS framework**

I'm using Bulma 0.7.5.
My **browser** is: Chrome

### Description

I do not know if it's a Feature request or a Question, but I would like to be able to override the Sass variables by CSS variables. The final goal is to offer the end user to be able to customize the theme of the site by updating some CSS variables using Javacript.
Right now, I'm pretty sure it's not possible, but I am ok do discuss the subject, offer solutions and even implement them.

### Steps to Reproduce
1. Set CSS variables, for instance:
`
:root {
    --primary: red;
  }
`
2. Using SCSS, override Bulma variables. For instance:
`$primary:var(--primary, blue) !default;`
`
3. Import Bulma SASS files:
`@import "~bulma/sass/utilities/_all";`

### Expected behavior

Be able to compile and update CSS variables using Javascript syntax:
`document.documentElement.style
    .setProperty('--primary', 'pink');`

### Actual behavior

Compile and see a Sass-loader error.

Grovkillen:
#1837 and scroll down

Issue #1837, opened 2018-04-28
Feature: Support CSS Custom Properties (Variables)


sandrina-p:
### Overview of the problem

This is about the Bulma **CSS framework**

I'm using Bulma **version** 7.0.0

I am sure this issue is **not a duplicate**?

### Request:

I've noticed it's [on Bulma road v1](https://github.com/jgthms/bulma/projects/1) the support to CSS Variables. Do you have in mind a deadline for it?
Thank you.

jgthms:
It’s already used for variable width columns.

You could also set it with your own Sass variables.

sandrina-p:
Not really. If I want $primary to be `var(--primary)`, being `--primary` `#008080` it doesn't work.

``` 
>> Error: argument `$color` of `darken($color, $amount)` must be a color
>>         on line 104 of node_modules/bulma/sass/base/helpers.sass, in function `darken`
>>         from line 104 of node_modules/bulma/sass/base/helpers.sass
>>         from line 5 of node_modules/bulma/sass/base/_all.sass
>>         from line 5 of frontend/simple/sass/bulma_overrides/base/_index.scss
>>         from line 3 of frontend/simple/sass/bulma_overrides/index.scss
>>         from line 9 of frontend/simple/sass/main.sass
>> >>       color: darken($color, 10%) !important; } }
```

Grovkillen:
Totally agree with OP.

https://caniuse.com/#search=custom%20properties

http://lea.verou.me/2016/09/autoprefixing-with-css-variables/

https://www.youtube.com/watch?v=2an6-WVPuJU

jordanranson:
+1 Would also like to be able to use CSS variables when customizing Bulma.

AdsonCicilioti:
+1

sandrina-p:
This issue was closed, can you @jgthms open it again and Maybe add the label `needs work` and `feature` ?

northamerican:
CSS variables are reactive and would be much easier to use, especially with frameworks that have you load Sass variables first for reuse anywhere in the app (like Nuxt)
+1

gamichalski:
CSS variables please! it's perfect!!!

eirikhm:
Any updates on this?

jgthms:
I’m trying to find a way to make both Sass variables and CSS variables coexist. It seems that some developers only want the former, some the latter, and some a mix of both. 

[...]

Issue #1775, opened 2018-04-05
Customize all colors without sass?


DarrylD:
Wondering if anyone found a decent way to override all of the colors without needing sass to override variables.

Looking to change the colors dynamically via js upon page load. Trying to get around having multiple "theme" stylesheet overrides. 

I may be reaching but, figured I'd ask.

jgthms:
Try using CSS variables:

```sass
$green: var(--green)
```

```css
:root {
  --green: #23d160;
}
```

whawker:
Would love this functionality too, unfortunately IE11's lack of support for CSS variables kills it 😢

sandrina-p:
I've tried that, but then it doesn't know the variable's real value (color) is...
```
Running "sass:dev" (sass) task
>> Error: argument `$color` of `red($color)` must be a color
>>         on line 46 of node_modules/bulma/sass/utilities/functions.sass, in function `red`
>>         from line 46 of node_modules/bulma/sass/utilities/functions.sass, in function `colorLuminance`
>>         from line 59 of node_modules/bulma/sass/utilities/functions.sass, in function `findColorInvert`
>>         from line 15 of node_modules/bulma/sass/utilities/derived-variables.sass
>>         from line 25 of frontend/simple/sass/bulma_overrides/utilities/variables.scss
>>         from line 1 of frontend/simple/sass/bulma_overrides/utilities/_index.scss
>>         from line 2 of frontend/simple/sass/bulma_overrides/index.scss
>>         from line 9 of frontend/simple/sass/main.sass
>> >>   $color-rgb: ('red': red($color),'green': green($color),'blue': blue($color
```

Is there a workaround?

sandrina-p:
> Would love this functionality too, unfortunately IE11's lack of support for CSS variables kills it 😢

@whawker Not really... 
```
.foo {
  color: red;
  color: var(--primary);
}
```

You can use fallback when the browser doesn't support css variables, it will read the previous matched property. Bulma could follow this approach. What do you think @jgthms ?

wtho:
@sandrina-p It would bloat up the css by adding the fallback to every color in css, which is still reasonable in my opinion.

More critical is that
* it would make the colorscheme easier changeable by browser plugins or end-users with little effort
[...]

PR #2981, opened 2020-06-05
Optional CSS Variables and multi-theme ability

Tofandel:
Over the years, there have been tons of demands and tries to get css variables into bulma
https://github.com/jgthms/bulma/issues/1775
https://github.com/jgthms/bulma/issues/2749
https://github.com/jgthms/bulma/issues/1837
#3028 

This PR does this successfully with a lot of ingeniosity by maximizing backward compatibility, the framework can be compiled exactly like before with no css variables, it can be compiled with all the css variables for outside theming or it can be compiled with a set of themes for bundled theming

This is a **new feature**.

### Proposed solution
Wrapping every variable in a few custom functions and to register them all beforehand
A theme mixin to output the used variables and the modified ones on successive call

I also revisited the build process from a simple cli command to a more complete nodejs script

### Tradeoffs
The only tradeoff of this solution is that because of the wrapping of each variable, the git diff is large and will cause a lot of merge conflicts and will require future PR to follow the new syntax (though it's not a hard syntax, just slightly repetitive)
There will also be a minor BC break because some variables have been modified (Only the map `$colors` that was renamed to `$colors-list` to keep BC and print a deprecation warning and the map `$shades` that are now only a list of names)
The other variables will stay compatible as long as you build without theming enabled (by default)

### Todos
- [x] Check that the HSL transformation corresponds to the output of lighten, saturate and adjust-hue
- [x] Update doc
- [x] Add a theme switcher in the doc
- [x] Update changelog
- [x] Cleanup 
- [ ] Remove files according to guidelines and squash commits


jgthms:
Interesting, thanks.

It seems quite overkill at first sight. I do like the idea of having a separate themable css file but I think most of the JS functions could have been done in Sass/Ruby.

For example the v() function doesn’t need 2 parameters as you could simply infer the name from the variable name.

Also there’s duplication in the css vars: if the hsl values are available seperately, we could use them instead of having a literal value as well. Because then we’d only need to change a color in a single location.

I also think the final output is missing the point of having the css vars, as the initial variable names are completely gone. For example “—bulma-success” should have a value of “var(—bulma-green)”. Same for “ dropdown-item-hover-color” for example. It has a literal value in the output, but should actually reference “var(—text-strong)” which itself should reference “var(—grey-darker)” which has a value of “#0a0a0a”.

Because right now if you want to update the grey-darker value you need to rebuild the whole CSS file with the compiler, so it’s missing the point of having CSS variables in the first place.

I have a branch with an initial CSS vars implementation but I might take inspiration of a few things here.

Tofandel:
I tried infering the variable name from the v function but it's not possible with sass, have a look also at the color loops and you will understand

For the duplication I was just thinking about only using hsl if available

Referencing the variables in the declaration would also be possible

I'll make nesting v() functions possible so it will do the trick

Tofandel:
@jgthms I've given some thoughts over your points during the weekend and found a way to solve them all

For the v function I can make it sass only and one param only just as long as I make a register function and call it on every variable before use

Basically the solution can be pure sass 

The only missing point is the one I talked about regarding BC breaks with HSL and non HSL values

One option could be to make every color HSL only but that may increase the size of the css quite a bit

I'll implement everything tomorrow, I think you'll be pleased 

It'll be ready to ship very soon

css/bulma.css
```diff
@@ -10032,12 +10005,12 @@ label.panel-block:hover {
 }
 
 .hero.is-white.is-bold {
-  background-image: linear-gradient(141deg, #e6e6e6 0%, white 71%, white 100%);
+  background-image: linear-gradient(141deg, #e8e3e4 0%, white 71%, white 100%);
```
Tofandel, line 10008: This is due to the adjust hue being executed at the same time than the adjust lightness, while before the adjust lightness was executed after the adjust hue (which does noting on white)

I consider this a bug fix

css/bulma-rtl.css
```diff
@@ -6839,7 +6817,7 @@ label.panel-block:hover {
 
 .tabs.is-boxed a {
   border: 1px solid transparent;
-  border-radius: 0 0 4px 4px;
```
Tofandel, line 6842: There was a bug during previous rtl conversion since non rtl it is 4px 4px 0 0, it should have been 4px 0 0 4px (left and right are reversed, but currently top and bottom also got reversed)

Tofandel:
All features to ship are now tested and ready

css/bulma-rtl.css
```diff
@@ -6839,7 +6821,7 @@ label.panel-block:hover {
...
```
Tofandel, line 6842: Bug introduced in rtl version since border-radius is top-l top-r bottom-r bottom-l there was no need to change this property
image

css/bulma.css
```diff
@@ -5997,7 +5979,7 @@ body.has-navbar-fixed-bottom {
 }
 
 .navbar-burger:hover {
-  background-color: rgba(0, 0, 0, 0.05);
```
Tofandel, line 6000: This was a hardcoded black and was changed to scheme-invert for themeability reasons

css/bulma.css
```diff
@@ -8121,17 +8103,11 @@ label.panel-block:hover {
 }
 
 .columns {
-  margin-left: -0.75rem;
-  margin-right: -0.75rem;
-  margin-top: -0.75rem;
-}
-
-.columns:last-child {
```
Tofandel, line 8129: Selectors are completely opposites and could be simplified, so they were

css/bulma.css
```diff
@@ -8726,13 +8702,7 @@ label.panel-block:hover {
 }
 
 .tile.is-ancestor {
-  margin-left: -0.75rem;
-  margin-right: -0.75rem;
-  margin-top: -0.75rem;
-}
-
-.tile.is-ancestor:last-child {
-  margin-bottom: -0.75rem;
```
Tofandel, line 8735: Same optimisation as columns

css/bulma.css
```diff
@@ -5049,10 +5044,7 @@ button.dropdown-item.is-active {
 
 .message.is-dark .message-body {
   border-color: #363636;
-}
-
-.message.is-primary {
-  background-color: #ebfffc;
```
Tofandel, line 5055: The background color of a message applies only to the message body, so it was moved there as a small optimization

css/bulma.css
```diff
@@ -3136,6 +3143,7 @@ a.tag:hover {
 
 .subtitle {
   color: #4a4a4a;
+  font-family: inherit;
```
Tofandel, line 3146: Some variables were displayed conditionally which doesn't work great with css variables I made the default become inherit as I though that would make more sense but if it's a problem it is possible to invalidate the css property by giving it a wrong value like `-`

jgthms:
Thanks. It’s a complex PR. I’ll try to have a more in-depth look.

I like the idea of having a separate themable Bulma CSS file. I’m not super keen of introducing a JS build process but it seems like the only solution for a few of the problems. 

The major issue I see right now is how most of the initial and derived variables become deprecated, even for non-themable setups. That’s not something I can introduce as it will break far too many setups. These variables are part of the core of Bulma and deprecating them is not possible.

I also think it’s not necessary to transform all Bulma variables into CSS variables. For theming purposes and dark mode support, colors would largely suffice. 

I think the way to go will be a toned-down version of the ideas you’ve introduced.

Tofandel:
@jgthms I actually solved the issue of the js build, it's not necessary anymore, I made a postcss plugin that takes care of optimizing the variables, after generation so it works with webpack and cli as well, though the js build is way faster than other methods

For non themeable setups, variables still work the way I did it, I marked it as deprecated only for theme makers/publishers

The advantage of this is really it's flexibility
If you have a look at the generated css for the docs you will see that only the variables that were used get converted to css variables

That's not something doable with a toned-down version

Tofandel:
Also I didn't mention it but I don't plan on having the themeable file included in the css dir, it's there as a demonstration and will be removed from the default build script

The way to use this feature would really be the one mentioned in the Themes section of the doc I added and this file has really no point except to show all the variables you can change and you could get

wtho:
Should the PostCSS-Plugin `postcss-var-optimize` be included in this repository? I think this would be best if it's use beyond bulma is not expected. It could still be published as own npm package.

I think a fully-themeable, minified version of bulma is also highly interesting to some use cases (especially for users which are not comfortable using node/sass). In my opionion it should be deployed via CDNs and included in the published npm package.

-----

Some features I would like to see (does not have to be at the first throw):
* optional addition of fallback variables for IE11 support:
  ```
  .element {
    color: #ad4359;
    color: var(--primary);
  }
  ```
* optional addition of transition elements:
  ```
  .element {
    color: var(--primary);
    background-color: var(--primary-invert);
    transition: color 500ms ease-in-out, background-color 500ms ease-in-out;
  }
  ```

Tofandel:
@wtho 
"Optional addition of fallback variables": IE11 is not to be supported by bulma but you can already do that with a post css plugin: [postcss-custom-properties](https://github.com/postcss/postcss-custom-properties)

"Optional addition of transition elements": If this PR is ever review and accepted (:roll_eyes:) you could make a PR or your own post css plugin for that

I'm not too much for the idea of having the full file deployed by default because it's not optimized and people might over use it when they could make an optimized version instead, on the other end the only use case is dynamic theming by the user where you'd need to make everything customizable via js, but as I said there wouldn't be enough use of this to have it deployed by default, that's why I'd recommend people to build it themselves because this version assumes you know what you're doing.

LoopsGod:
It's insane that this is not being merged / worked on. I appreciate your work @Tofandel

chrishumphrey1985:
This is definitely a feature that Bulma needs. What's holding it up?

adriansalvatori:
How can I use this PR instead of main branch? I really need this feature.

stellarbear:
It's 2021. Bulma still do not support css variables 🤔

acallaghan:
Would this just be better as a fork of Bulma, instead of mainline? Given that this changes the direction of the project quite a bit & hasn't had support from the creator?

chrishumphrey1985:
I'm not even sure if Bulma is being actively developed anymore.

Tofandel:
Days of work have been put into this PR which didn't get support from the creator

I stopped using bulma so will not be resolving the conflicts on this PR and in the end this work will just be lost and serve no purpose, so if anybody actively using bulma wants to take it over, feel free to fork my PR, as it was finished and now only needs conflict resolution

Tofandel:
@daniil4udo Thanks, it seems you as well. As I explained I'm not using bulma in any of my projects anymore so I won't go into a thorough review, though I have one feedback, the fact that you're using a mix of css variables and scss vars makes it very hard to debug, read and maintain, my solution for this was a custom library to handle all this with scss vars and mixins https://github.com/Tofandel/bulma/blob/fbfe63286ddafe70d3b51ef091c578bf327f5dda/sass/utilities/functions.sass

You need to register each variable with +register('var-name', value)

And use it with v('var-name'), each util like darken had its own counterpart (`vDarken('var-name', value)`) to create a new derived variable when used

In the end the theme mixin just outputs all the used variables at the root, this allowed for tree shaking and then for a postcss plugin I made to interpret all variables in the css so if you had only one theme, you would not have any variable because they would all be converted back. But if you had multiple themes only the variables that change in the theme would be kept, making it super optimized and lightweight, I had the full documentation written out in the PR

It might be a bit hard to change that now that it's already all written, but my recommendation would really be to use this custom lib I made because it's just magical engineering.

I even planned to extract it as a separate project because it can be used to convert any scss lib to a variable css lib super easily (really after this functions were developed I just did a replace all with regex in the project)

And I definitely still will when I get some time

PR #2996, opened 2020-06-13
Fixes #2367 add optional ability to push down footer for short content

### LINKED ISSUE #2367: Footer behavior
This is a question or a bug about **Bulma**.

### Overview of the problem

I'm using Bulma **version** 0.7.2
My **browser** is: Firefox 65.0
I am sure this issue is **not a duplicate**, at least I made an extensive search.

### Description

One always want the footer to be the last displayed element of a page. But when the previous container/section is too short, there is still some blank after the footer.

### Steps to Reproduce

1. Create a page ending with a footer.
2. Put few content on the page.
3. Be sure the footer is the last element of the body.

image

### Expected behavior

Add spacing to fill the page like for full height hero so there will be nothing after the footer.

See **Actual behavior**: the expected behavior is case B even when we match condition of Case A.

### Actual behavior

**Case A**:

When the page is filled enough, there is not space behind the footer.

image

Another example: [this page](https://themes.gohugo.io//theme/bulma/post/hugoisforlovers/).

**Case B**:

When the page is not filled enough, there is space behind the footer.

image

Another example: [this page](https://themes.gohugo.io//theme/bulma/about/).

xdivby0:
This is a **improvement**.
Having the ability to push down the footer for short content pages has been requested by many. This pull request is referencing issue #2367.

### Proposed solution

The solution adds an optional class for footer-parents to push the footer down by growing the second-last section to fill the remaining space.

Without this improvement:
grafik

With this improvement:
grafik

### Tradeoffs

As this is optional, there are no drawbacks by simply not using this class. The drawback _when_ using the class, is that at the moment only sections are working as the element to be growed. 

### Testing Done

None.

I tried it in a project and pasted screenshots above - it seems to work though I am not 100% sure it works all the time, maybe there are special cases where this does not work.

### Changelog updated?

No.

### DIFF
--- a/sass/layout/footer.sass
+++ b/sass/layout/footer.sass
@@ -7,3 +7,11 @@ $footer-padding: 3rem 1.5rem 6rem !default
   padding: $footer-padding
   @if $footer-color
     color: $footer-color
+
+.has-pushed-down-footer-child
+  height: 100vh
+  display: flex
+  flex-direction: column
+
+.has-pushed-down-footer-child>section:nth-last-child(2)
+  flex-grow: 1


xdivby0:
Do I have to tell someone about opening this request or how does this work? I am new and I don't understand why my help is not seen

PeterSherk:
bumping this for relevance now that 1.0 is out

xdivby0:
@PeterSherk  so nice of you, but this PR is 4 years old now, I forgot I ever made it and it could very well be that I'm not accomplishing the goal with the correct tools, maybe stuff changed.

PR #2997, opened 2020-06-13
Change tablet breakpoint start width from 769px to 768px.

badalsurana:
This is a **improvement**.
Most tablet widths start from 768px including iPad and it is very confusing when tablet breakpoint ignore it.

### Proposed solution
Changes tablet breakpoint from 769px to 768px

### Tradeoffs
None

### Testing Done

None.

### Changelog updated?

No.


jgthms:
It's meant to be "tablet landscape". So if the value was `768px`, the styles would be applied on "tablet portrait" as well, which in Bulma's case is considered "mobile". Not the best name, but it's always been like this. It would break too many setups to change it now.

PR #3049, opened 2020-07-18
A11y: Enhance tabs docs

Muhnad:
Enhancing the implementation of tabs by following the instructions of [WAI ARIA tab pattern](https://www.w3.org/TR/wai-aria-practices/#tabpanel).

applying the new improvements will help buefy team to fix the [buefy tab issue](https://github.com/buefy/buefy/issues/2602) they mentioned a couple of days ago.

References:
[wai aria practices](https://www.w3.org/TR/wai-aria-practices/#tabpanel)
[inclusive components](https://inclusive-components.design/tabbed-interfaces/)

### DIFF
--- a/docs/documentation/components/tabs.html
+++ b/docs/documentation/components/tabs.html
@@ -16,60 +16,78 @@ meta:
 
 {% capture tabs_example %}
 <div class="tabs">
-  <ul>
-    <li class="is-active"><a>Pictures</a></li>
-    <li><a>Music</a></li>
-    <li><a>Videos</a></li>
-    <li><a>Documents</a></li>
+  <ul role="tab">
+    <li role="presentation" class="is-active">
+      <a href="#picture-panel" id="picture-tab"
+         role="tab" aria-controls="picture-panel" aria-selected="true">Pictures</a>
+    </li>
+    <li role="presentation">
+      <a href="#music-panel" id="music-tab"
+         role="tab" aria-controls="music-panel" aria-selected="false">Music</a>
+    </li>
+    <li role="presentation">
+      <a href="#video-panel" id="video-tab"
+         role="tab" aria-controls="video-panel" aria-selected="false">Videos</a>
+    </li>
+    <li role="presentation">
+      <a href="#docs-panel" id="docs-tab"
+         role="tab" aria-controls="docs-panel" aria-selected="false">Documents</a>
+    </li>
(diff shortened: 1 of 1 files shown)


docs/documentation/components/tabs.html
```diff
@@ -16,60 +16,78 @@
...
 {% capture tabs_example %}
 <div class="tabs">
-  <ul>
-    <li class="is-active"><a>Pictures</a></li>
-    <li><a>Music</a></li>
-    <li><a>Videos</a></li>
-    <li><a>Documents</a></li>
+  <ul role="tab">
+    <li role="presentation" class="is-active">
+      <a href="#picture-panel" id="picture-tab"
```
Muhnad, line 21: Using Links vs Buttons?
- by using links will make the components work well if there's no JS loaded.

docs/documentation/components/tabs.html
```diff
@@ -16,60 +16,78 @@
...
+    </li>
+    <li role="presentation">
+      <a href="#docs-panel" id="docs-tab"
+         role="tab" aria-controls="docs-panel" aria-selected="false">Documents</a>
+    </li>
   </ul>
 </div>
+
+<section id="picture-panel" role="tabpanel" tabindex="-1" aria-labelledby="picture-tab">Pictures Content</section>
+<section id="music-panel" role="tabpanel" tabindex="-1" aria-labelledby="music-tab" class="is-hidden">Music Content</section>
```
Muhnad, line 40: to make the components work well if there's no JS we need to append the `is-hidden` class by JS. not in plain HTML.

pez-wb:
Why not include adding buttons to the tab structuring as part of this?

Tabs - as a general rule – are used to navigate between sections on the same page/document. 
Meaning that the clicking a tab is an "on page" interaction and NOT a "link to a separate document".

`<a>` tags are fine as a option - but from a purely HTML a11y and semantically correct POV - the `<button>` element is should be the default for any tabbed UI.

FWIW — I am happy to help implement/PR this change.

pez-wb:
It looks like someone opened a PR related to this [here](https://github.com/jgthms/bulma/pull/2003) but it's old and IMHO is too broadly scoped. I'd focus on Tabs for now and then find any other "on page" interaction elements that require use of the `<a>`.

https://github.com/jgthms/bulma/pull/2003

PR #3081, opened 2020-08-13
refactor: split out responsiveness mixins

elfin-sbreuers:
This is an improvement.

This allow to make the grid be usable as standalone element by just
including

@import "utilities/initial-variables";
@import "utilities/mixins-responsiveness";
@import "grid/columns";

### Tradeoffs

I don't see any since from my point of view the `utilities/_all.sass` would be used anyways for every one who wants to use the whole framework.

### Testing Done

Manually.

I imported the relevant files in my sass. The grid system could be generated and used without dragging in more dependencies than necessary. The build worked as well.

### Changelog updated?

No.

### DIFF
--- a/sass/utilities/_all.sass
+++ b/sass/utilities/_all.sass
@@ -6,4 +6,5 @@
 @import "derived-variables.scss"
 @import "animations.sass"
 @import "mixins.sass"
+@import "mixins-responsiveness.sass"
 @import "controls.sass"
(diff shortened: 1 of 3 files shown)


jgthms:
The problem will be for setups that only import `mixins.sass` and except the responsive mixins to be there. You might try Dart Sass and import only what you need: https://sass-lang.com/documentation/at-rules/use

PR #3086, opened 2020-08-21
Allow each component to have its own colors and default to global ones

aredridel:
This is a **new feature**.

### Proposed solution

Tags and buttons did not have enough contrast for our use case, and we wanted to use a dark grey body text in most situations such as messages, not a darker flavor of the background color. It is simple enough to add intermediate variables as customization points for the colors of each element or component, allowing us to do this. It's not perfect — buttons still make more assumptions about how colors are related than we'd like — but it goes a long way toward making Bulma customizable for our uses.

### Tradeoffs

- It adds more variables
- It's not complete. Buttons in particular have a lot of assumptions about how colors relate for use in various states. They're a complex element!

### Testing Done

- Base build and confirm colors work as they did
- Patched bulma as we use it (via sass include and our own build) and noted that we can customize the colors for each component

### Changelog updated?

No. It's unclear how to do so for unreleased changes, and it appears it's maintained more periodically in practice than with each PR.

### DIFF
--- a/sass/elements/table.sass
+++ b/sass/elements/table.sass
@@ -23,6 +23,8 @@ $table-row-active-color: $primary-invert !default
 $table-striped-row-even-background-color: $scheme-main-bis !default
 $table-striped-row-even-hover-background-color: $scheme-main-ter !default
 
+$table-colors: $colors !default
+
 .table
   @extend %block
   background-color: $table-background-color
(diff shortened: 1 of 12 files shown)


sass/elements/table.sass
```diff
@@ -23,6 +23,8 @@ $table-row-active-color: $primary-invert !default
 $table-striped-row-even-background-color: $scheme-main-bis !default
 $table-striped-row-even-hover-background-color: $scheme-main-ter !default
 
+$table-colors: $colors !default
```
tim-reynolds-thegrommet, line 26: I don't see this variable being used in the PR

tim-reynolds-thegrommet:
I definitely think this would be a nice addition. Full disclosure, I work with @aredridel so I am probably biased, but objectively having more control if a developer wants it seems like a win to me.

PR #3047, opened 2020-07-17
Flexbox helpers

mrm:
This is a **new feature**.

When using Bulma I often use `is-flex` but the modifiers you'll want to use are missing such as `justify-content: space-between`. I found a couple of issues/PRs such as the most recent #3008 which suggest adding flexbox helpers. 

### Proposed solution

I've added most flexbox options such as `is-justify-content-space-between`. Perhaps naming can be improved as some are quite long (the suggested `is-justified-space-between` or `{has-,is-}justify-between` are probably be better), and perhaps some hardly used values can be removed.

So it's not done yet, but wanted to get some feedback early on.

### Testing Done

None.

### Changelog updated?

No.

PR #3087, opened 2020-08-23
Add Flex Helpers

jjohnson1994:
Add flex helpers

This is a **new feature**.

### Proposed solution

Though flex helpers would be a good addition to the helper sass files. I wanted to align right without using float and prefer using flex. Noticed there were no flex helpers so I've added some.

### Tradeoffs

- Adds more classes

### Testing Done

Built and tested locally

### Changelog updated?

No.

p.s. I've not updated the Changelog or docs yet. I wanted to see what people thought of the new helpers first and wasn't sure about version numbers.

### DIFF
--- a/sass/helpers/flex.sass
+++ b/sass/helpers/flex.sass
@@ -1,52 +1,52 @@
 .is-flex {
-  display: flex
+  display: flex !important
 }
 
 .is-flex-row {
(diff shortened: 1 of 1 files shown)


jgthms:
This was fixed in #3047

PR #3105, opened 2020-09-07
Fix z-index of disabled input/select

lionzeye:
This is a **bugfix**.

### Example

A disabled input/select has no border, so it overlaps the border of neighbouring fields, buttons and selects, thus breaking the expected style.

disable input overlap

```html
<div class="field">
   <label class="label" for="offerCostPrice0">Kostprijs</label>
      <p class="control">
         <a class="button is-static has-text-weight-bold">€</a>
      </p>
      <p class="control is-expanded">
         <input name="offerCostPrice[]" id="offerCostPrice0" value="12.75" class="input" disabled>
      </p>
   </div>				
</div>
```

### Proposed solution

Inputs, buttons and selects in a grouped field should have a base z-index of 1 and disabled inputs/select 0, so disabled inputs/selects which are borderless, don't overlap the border of neighbouring fields, buttons and selects anymore.

fixed z-index input and select

### Tradeoffs

None

### Testing Done

Works fine for various combinations of combined inputs, buttons and selects

### Changelog updated?

No.

### DIFF
--- a/sass/form/tools.sass
+++ b/sass/form/tools.sass
@@ -69,6 +69,7 @@ $label-colors: $form-colors !default
       .button,
       .input,
       .select select
+        z-index: 1
         &:not([disabled])
           &:hover,
           &.is-hovered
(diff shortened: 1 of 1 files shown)


lionzeye:
Will rebase this if https://github.com/jgthms/bulma/pull/3192 is merged

PR #3109, opened 2020-09-14
prune Not Fonud

FutureLearnSole:
This is a **chore** done.

### Tradeoffs
List is pruned,

### Testing Done
404 pings for 
1. https://buefy.github.io/
2. http://sketch.fordesignrs.com/

### DIFF
--- a/docs/_data/expo.json
+++ b/docs/_data/expo.json
@@ -366,6 +366,11 @@
       "url": "https://www.oneday.design/",
       "date": "8 Dec 2016"
     },
+    "Buefy": {
+      "name": "Buefy",
+      "url": "https://buefy.org/",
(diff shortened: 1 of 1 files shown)


docs/_data/expo.json
```diff
@@ -372,11 +366,6 @@
       "url": "https://www.oneday.design/",
       "date": "8 Dec 2016"
     },
-    "Buefy": {
-      "name": "Buefy",
-      "url": "https://buefy.github.io/",
```
service-paradis, line 377: This should be updated to https://buefy.org/ instead

roiLeo (approved):

jgthms:
Thanks but I think it would be better to just remove the links from the expo so that visitors can still see the screenshots.

PR #3118, opened 2020-09-22
add variables to configure the darken and lighten amounts

ccondry:
This is an improvement. I wanted to be able to make the hover and active colors of Bulma buttons a darker shade. I found that this is done with a 2.5% and 5% static value, so I made them configurable variables like almost everything else in Bulma. I did a search through the code and found a few more of these static percentages, along with a few 10% values, so I created initial variables to cover all of these values.

### Proposed solution

Solution is to create and use variables to make the darken and lighten amounts used in Bulma configurable.

### Tradeoffs

I really can't think of any tradeoffs here. The only other alternative is to overwrite some SASS in the project that is using Bulma, but I feel like this is supposed to be in the Bulma code base.

### Testing Done

I started from the main (master) branch, ran `npm install` and `npm run build` to make sure it builds normally. I also tested the values in my project to make sure they worked as initial values and when configured to a different value.

### Changelog updated?

No - I was not sure how to add content to changelog without it being in a new version section. But it would be like this:

### New features

* New variables `$darken-half`, `$darken-full`, `$darken-double`, `$lighten-full`, `$lighten-double`, `$button-darken-half`, `$button-darken-full`


aleavikraman:
this would be nice to have. Is it being considered to merge in?

PR #3119, opened 2020-09-22
Update block.html - Clarify examples

dangowans:
The example of text inside `div`s without the `.block` class felt awkward.

### Proposed solution

In the examples where the `.block` class is not used, don't use sample text that says the `block` class is used.

### Tradeoffs

None.

### Testing Done

None.

### Changelog updated?

No.

### DIFF
--- a/docs/documentation/elements/block.html
+++ b/docs/documentation/elements/block.html
@@ -28,13 +28,13 @@ meta:
 
 {% capture no_block_example %}
 <div>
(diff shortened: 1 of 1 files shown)


jgthms:
Of course! Thanks for catching that so quickly.

dangowans:
Honestly, I didn't realize it was so fresh.  Just landed on that page looking for a way to do something. : )

PR #3121, opened 2020-09-23
Update generic.sass

AkmarNafi:
updated $body-overflow-y to 'auto'. 

This is a improvement

The previous value  for 'overflow-y' was 'scroll' which causes a scroll bar to appear even in there is no need to scroll.  When the property is set to 'auto' , the scroll bar will automatically appear if there is an overflow.
Having a scrollbar when it is not need is annoying and does not look good

### Tradeoffs

there are no drawbacks to this solution

### Testing Done

None.

### Changelog updated?

No.

### DIFF
--- a/sass/base/generic.sass
+++ b/sass/base/generic.sass
@@ -4,7 +4,7 @@ $body-min-width: 300px !default
 $body-rendering: optimizeLegibility !default
 $body-family: $family-primary !default
 $body-overflow-x: hidden !default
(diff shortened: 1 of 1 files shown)


jgthms:
So this is intentional as to avoid the page to “jump” when navigating between pages of different height. If a user navigates from a page with scroll bar to a page without, the whole content can jump 15px horizontally which is bad UX.

PR #3141, opened 2020-10-04
Add helpers responsiveness + borderless helper

CreativeBulma:
This PR adds responsiveness to new Flexbox and Spacing helpers and also new is-borderless helper


drewalth:
Nice addition.

PR #3176, opened 2020-11-07
Use imports everywhere

### LINKED ISSUE #2797: Unable to fully use Bulma Sass with the new Sass modular system (@use)
This is about **Bulma**.

### Overview of the problem

I'm using Bulma **version** [0.8.0] 

This is a **Sass** issue: I'm using version [1.25.0] (Dart Sass)

### Description

I was refactoring code to start using the new Sass modular system and at first it was ok, but once I tried to use navbar.sass directly the compiler started complaining:
`
SassError: Undefined variable.
    1 │ $navbar-background-color: $scheme-main !default
` 
I tried several options like placing the variable in the file I am calling this from, pushing the variable though with the 'with' option, calling @import on the derived-variables but nothing worked.
I feel that the Bulma structure would need to be updated to support this.

### Steps to Reproduce
1. `@use "~bulma/sass/components/navbar.sass" with (
  $scheme-main: "red"
);`
2. compile

### Expected behavior

It compiles

### Actual behavior
Get an error
`
SassError: Undefined variable.
    1 │ $navbar-background-color: $scheme-main !default
`

jgthms:
This is an **improvement**.

### Proposed solution

Instead of relying on the ancestor Sass file to import dependencies, each component file imports its own dependencies.
Fixes #2797

### Tradeoffs

No real tradeoff, unless some old Sass version does not support importing the same file multiple times.

### Testing Done

Tested with latest versions of:

* Dart Sass `npm sass@1.29.0`
* Node Sass (LibSass) `npm node-sass@5.0.0`
* even Ruby Sass `sass-3.7.4.gem`

### Changelog updated?

Yes.


lionzeye:
There is a lot of duplicate css in the output file after this was merged. Selector
`.image.is-1by3 .has-ratio, .modal, .modal-background, .is-overlay, .hero-video`
appears in the output file 40 times now. See the output in https://github.com/jgthms/bulma/blob/38aa8b9299cb70f165783858122bcc7f6e3d6d35/docs/css/bulma-docs.css for example

jgthms:
You're right, I've noticed that too. It's how `@extend` rules are populated. I thought cssnano might catch it and optimize it, but it didn't. I will create a new `extends.sass` file in the `/utilities` folder which will only be imported once.

PR #3180, opened 2020-11-11
Adds is-required to label

moladukes:
Adds .is-required class to label adding a required star. 
Uses the $danger color variable.

This is a **improvement**.

### Proposed solution

Add a required class option to form labels

Screen Shot 2020-11-11 at 8 20 39 AM

Screen Shot 2020-11-11 at 8 25 44 AM

### Tradeoffs

May want to style your required labels differently than the classic star pattern and or move the star as an :after element. 

### Testing Done

Tested on Standard and Horizontal form labels.

### Changelog updated?

No.

### DIFF
--- a/sass/form/tools.sass
+++ b/sass/form/tools.sass
@@ -21,6 +21,9 @@ $label-colors: $form-colors !default
     font-size: $size-medium
   &.is-large
     font-size: $size-large
+  &.is-required:before
+    content: "*"
+    color: $danger
 
 .help
   display: block


citrus:
+1 for the `is-required` class to be on the field vs the label. This would play well with the [buefy field](https://buefy.org/documentation/field/) component.

yamin-oanda:
I just implemented the `is-required` class on my own and guessed that others might be using similar naming. Here's mine:

```css
select:invalid, input:invalid, input:invalid + .file-cta {
  border-left-color: var(--bulma-danger);
}

select.is-required, input.is-required, input.is-required + .file-cta {
  border-left-width: 4px;
}
```

(Note: I am by no means a CSS guru, so there's probably a better way to express this.)

It looks like:

Screenshot 2024-10-10 at 09 46 07

I decided to go with a thicker left-border on the input itself instead of an asterisk on the label. Imho an asterisk on the left of the label looks awkward, and on the right of the label messes up alignment. A slightly thickened left border seems like a good option.

WDYT?

t1:
Wouldn't it be nicer to directly use the [`required` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/required) and the `:required` pseudo class?

t1:
If we also add `textarea`, that would be:

```css
select:required, input:required, input:required + .file-cta, textarea:required {
    border-left-width: 4px;
}
```

Issue #3016, opened 2020-06-25
Screen size dependent spacing helpers


ambrt:
Is it a bug/feature/question or do you need help? **Feature**

All helpers from https://bulma.io/documentation/helpers/spacing-helpers/ like `pb-1` could have their counterparts of `is-mobile` like `pb-1-mobile` - so padding would be added only on mobile screen size.

nedjo:
Breakpoint support didn't make it into the commits in #451 where spacing support was added; see https://github.com/jgthms/bulma/commit/a2a9564ca9efb643cb3992a83f9df802159449c4#diff-638e23261295b3336ed9b3561484828d. From @jgthms in [that issue](https://github.com/jgthms/bulma/issues/451#issuecomment-612836802):

> I'm not sure how to handle responsiveness, in order to have mobile/tablet/desktop/widescreen/fullhd version of each spacing helper. If following the above approach, you would have `m-0-mobile` and `m-2-tablet` that could be combined

guillaumeprevost:
Bumping this feature request :)
 
Just a thought, another possibility could be to have a way to declare that a given spacing should be "responsive/fluid" so that when hitting a breakpoint it becomes smaller or larger automatically by a given percentage.
For example if my spacing "3" is 32px (so mt-3 is a margin-top: 32px), I could set the class "mt-3-fluid" so that it is 32px on desktop, 32*80% on tablet, 32*60% on mobile). 
Not sure whether this is the Bulma way...

PR #3184, opened 2020-11-12
Responsive spacing helpers v2

lionzeye:
This is an **improvement**.

This is a fixed version of https://github.com/jgthms/bulma/pull/3179

### Proposed solution

Responsive spacing helpers are a much desired improvement.
https://github.com/jgthms/bulma/issues/3016
https://github.com/jgthms/bulma/issues/451#issuecomment-647723514

This pull request adds responsive spacing helpers for all the current spacing helpers.

Example of use:
.is-sidebar-menu has a default padding of 2.5 rem. If I want a more compact menu on a certain range of screen sizes, by reducing the padding, I'd immediately have to start writing exceptions in my css. I'd much rather add a single class to the component, like: .p-2-tablet, to get the desired result. If I want similar behaviour on 3 components, I'd have 3 exceptions in my css, when that could be replaced by simply adding classes. If I want different behaviour in a one case I can simply remove the class on that specific page instead of having to create an exception on the already created exception.

### Tradeoffs

The only trade-off I see, is potentially bloating the bulma css output. I personally think the improvement outweighs this drawback, but I'm very eager to hear more opinions.

### Testing Done

I've tested multiple responsive spacing helpers per spacing type and direction (or axis).

### Changelog updated?

No.

### DIFF
--- a/sass/helpers/spacing.sass
+++ b/sass/helpers/spacing.sass
@@ -29,3 +29,174 @@ $spacing-values: ("0": 0, "1": 0.25rem, "2": 0.5rem, "3": 0.75rem, "4": 1rem, "5
       .#{$shortcut}#{$spacing-vertical}-#{$name}
         #{$property}-top: $value !important
         #{$property}-bottom: $value !important
+  +mobile
+    @each $name, $value in $spacing-values
+      // All directions
+      .#{$shortcut}-#{$name}-mobile
+        #{$property}: $value !important
+      // Cardinal directions
+      @each $direction, $suffix in $spacing-directions
+        .#{$shortcut}#{$suffix}-#{$name}-mobile
+          #{$property}-#{$direction}: $value !important
+      // Horizontal axis
+      @if $spacing-horizontal != null
+        .#{$shortcut}#{$spacing-horizontal}-#{$name}-mobile
+          #{$property}-left: $value !important
+          #{$property}-right: $value !important
+      // Vertical axis
+      @if $spacing-vertical != null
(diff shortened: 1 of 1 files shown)


mralexho:
In case anyone wants to use this as a patch file -

```
"scripts": {
  ...
  "postinstall": "patch --forward node_modules/bulma/sass/helpers/spacing.sass < patches/spacing.patch"
}
```

[Patch file on gist](https://gist.github.com/mralexho/5b6b7727272f3c7ec853af6fa58c73e1)

jgthms:
I like the idea but there is a lot of repetition (which means room for mistakes). Could you make a mixin out of it? Something like this:

```scss
@mixin spacingHelpers($suffix = "") {
  @each $property, $shortcut in $spacing-shortcuts {
    @each $name, $value in $spacing-values {
      // All directions
      .#{$shortcut}-#{$name}#{$suffix} {
        #{$property}: $value !important;
      // etc.
}
```

lionzeye:
Thanks for the feedback. I have limited experience with SASS, but have updated my patch accordingly and I'm doing some final checks.

lionzeye:
Ok, I've checked and tested the output CSS and it works as intended. I've excluded padding helpers with value 'auto', because that isn't valid and padding should be either a length or a percentage. I'll open a separate pull request to get that fixed for the standard spacing helpers.

lionzeye:
Got rid of some code duplication and renamed the mixin to better fit the bulma conventions

Sogl:
Why hasn't it been merged yet?

PR #3192, opened 2020-11-18
Fix #3188 Inconsistent grouped button border behavior

### LINKED ISSUE #3188: [bug] Inconsistent button border behavior when grouped with .has-addons class
This is about **Bulma**.

Is it a bug/feature/question or do you need help? : It is a Bug
If it's a bug, is it a browser bug?: No

### Overview of the problem

I'm using Bulma **version** [0.9.1]
I am sure this issue is **not a duplicate**? None that I could find

### Description

When 3 or more buttons are grouped (with `.has-addons`) certain patterns result in visual border bugs. I've found one of these patterns to be a when button with a border style is grouped with buttons without a border style.

### Steps to Reproduce

Simple example with addons
```html
<div class="buttons has-addons">
  <button type="button" class="button is-primary is-light">
    <span>a</span>
  </button>
  <button type="button" class="button">
    <span>b</span>
  </button>
  <button type="button" class="button is-primary is-light">
    <span>c</span>
  </button>
</div>
```

Also in fields
```html
<div class="field">
  <div class="field-body">
      <p class="control">
        <button type="button" class="button is-danger is-light">
          <span>a</span>
        </button>
      </p>
      <p class="control">
        <button type="button" class="button">
          <span>b</span>
        </button>
      </p>
      <p class="control">
        <button type="button" class="button is-primary is-light">
          <span>c</span>
        </button>
      </p>
    </div>
  </div>
</div>
```

https://codepen.io/MarceDev/pen/NWrJqpM?editors=1011

### Expected behavior

The left border should not be visible.

### Actual behavior

The left border is visible while the right border is not visible, so this is inconsistent.

lionzeye:
This is a **bugfix**.

Fixes https://github.com/jgthms/bulma/issues/3188 which is an issue where we concluded that the z-index of borderless, grouped buttons should be lower than the z-index of grouped, bordered buttons.

### Proposed solution

Set the z-index of grouped, borderless buttons to a lower value than grouped, bordered buttons. While bumping the z-index values of :hover and :focus states etc.

### Tradeoffs

None that I'm currently aware of.

### Testing Done

Tested with example from issue https://github.com/jgthms/bulma/issues/3188 and different grouped button combinations.

### Changelog updated?

No.

### DIFF
--- a/sass/elements/button.sass
+++ b/sass/elements/button.sass
@@ -45,6 +45,8 @@ $button-static-border-color: $border !default
 
 $button-colors: $colors !default
 
+$button-borderless-colors: ('white','light','black','text','primary','link','info','succes','warning','danger') !default
+
 // The button sizes use mixins so they can be used at different breakpoints
 =button-small
   border-radius: $radius-small
@@ -307,6 +309,10 @@ $button-colors: $colors !default
       +button-large
   &.has-addons
     .button
+      z-index: 2
+      @each $name, $pair in $button-borderless-colors
+        &.is-#{$name}
+          z-index: 1
       &:not(:first-child)
         border-bottom-left-radius: 0
         border-top-left-radius: 0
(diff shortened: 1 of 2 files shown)


suterma:
I want to raise awareness of a potential painting performance degradation when changing the z-index on hover. See https://stackoverflow.com/questions/79130607/why-does-the-bulma-css-framwork-change-the-z-index-on-hover

It might be unrelated to the issue #3188, because as far as I understand the reported error is not related to hover, althoug in the proposed solution above, hovering seems to be affected or introduced. If the proposed solution for this issue is the correct one, this could answer the stackoverflow question. Feel free to answer accordingly if so.

PR #3208, opened 2020-12-06
Do not override is-rounded with button-small. Fixes #3163.

### LINKED ISSUE #3163: button is-rounded does not work with buttons are-small
This is about Bulma

### Overview of the problem

I'm using Bulma **v0.9.1**
My **browser** is: Firefox

### Description
The class for "buttons are-small" overrides the "button is-rounded" functionality.

### Steps to Reproduce
1. Create a div with classes "buttons are-small"
2. Create div with class "button is-rounded" inside buttons div
[JSFiddle](https://jsfiddle.net/92sou5cd/)

### Expected behavior

Buttons are small and rounded

### Actual behavior
Buttons are small, but not rounded

raineorshine:
This is a bugfix for #3163.

### Proposed solution

`button-small` currently assumes the button is not rounded and overrides its border-radius from `$radius` to `$radius-small`. This is not valid however if `is-rounded` is applied to the button. 

This PR adds `&:not(.is-rounded)` to avoid this. This solution is preferable to #3164 which uses `!important`.

### Tradeoffs

None

### Testing Done

Tested with the following code:

```html
<p>This button should be small and rounded:</p>
<div class="buttons are-small">
  <button class="button is-rounded">Hello</button>
</div>
```

Bulma `v0.9.1` incorrectly renders the button not rounded. Test and confirmed that this PR causes the button to be correctly rendered small and rounded.

### Changelog updated?

No.

### DIFF
--- a/sass/elements/button.sass
+++ b/sass/elements/button.sass
@@ -47,7 +47,8 @@ $button-colors: $colors !default
 
 // The button sizes use mixins so they can be used at different breakpoints
 =button-small
-  border-radius: $radius-small
+  &:not(.is-rounded)
+    border-radius: $radius-small
   font-size: $size-small
 =button-normal
   font-size: $size-normal


jgthms:
Thanks, I like this solution.

PR #3233, opened 2021-01-05
Allow input colors on contenteditable

raineorshine:
This is a **bugfix**.

### Proposed solution

[Input color classes](https://bulma.io/documentation/form/input/#colors) like `is-danger` should also apply to [contenteditable](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/contenteditable) elements, which are functionally similar to inputs and textareas.

This solution only applies the border color and does not assume anything else about how the contenteditable is styled.

### Tradeoffs

It is possible that existing developers would be using `is-danger` on a `contenteditable` with different semantics.

### Testing Done

Tested:

```html
<div class="is-danger" contenteditable>Hello</div>
```

### Changelog updated?

I can add a changelog and update documentation upon request.

### DIFF
--- a/sass/form/input-textarea.sass
+++ b/sass/form/input-textarea.sass
@@ -4,14 +4,7 @@ $textarea-min-height: 8em !default
 
 $textarea-colors: $form-colors !default
 
-%input-textarea
-  @extend %input
-  box-shadow: $input-shadow
-  max-width: 100%
-  width: 100%
-  &[readonly]
-    box-shadow: none
-  // Colors
+@mixin input-color-border
   @each $name, $pair in $textarea-colors
     $color: nth($pair, 1)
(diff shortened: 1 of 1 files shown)


jgthms:
Good idea, but why not use `@extend %input-textarea` on the `[contenteditable]`?

raineorshine:
`input-textarea` contains a few additional properties that affect the default element, rather than just enabling input color classes (`is-danger` etc):

```css
  box-shadow: $input-shadow
  max-width: 100%
  width: 100%
  &[readonly]
    box-shadow: none
```

Setting the width doesn't seem desirable since `[contenteditable]` could be an inline element. This contrasts with textareas which are basically always block.

PR #3241, opened 2021-01-20
fix: increase contrast between disabled input text and background flood

13protons:
This is a **improvement**.

The current color contrast for disabled form inputs is below legibility standards (WCAG AA) of 4.5:1: 

Screen Shot 2021-01-19 at 11 29 07 PM

Screen Shot 2021-01-19 at 11 29 24 PM

### Proposed solution

Use `$grey-darker` for the text color of disable inputs, as the lack of interactivity, disabled cursor, and gray background all work to reinforce that the input is, in fact, disabled. 

The outcome still looks disabled but is now more legible and has better objective contrast:

Screen Shot 2021-01-19 at 11 59 31 PM

Screen Shot 2021-01-19 at 11 42 38 PM
 

### Tradeoffs

This seems like a straightforward accessibility win. There may be a minor risk of users attempting to interact with the disabled field, but it should present much less of legibility issue especially for users with less-than-perfect eyesight. 

### Testing Done

Verified build output with codepen by pasting in the output of `npm run build` (bulma.min.css) and rendering code samples from input and text area docs: 

[https://codepen.io/alanguir/pen/qBavxGK](https://codepen.io/alanguir/pen/qBavxGK)

### Changelog updated?

No.

### DIFF
--- a/sass/form/shared.sass
+++ b/sass/form/shared.sass
@@ -18,7 +18,7 @@ $input-focus-border-color: $link !default
 $input-focus-box-shadow-size: 0 0 0 0.125em !default
 $input-focus-box-shadow-color: bulmaRgba($link, 0.25) !default
 
-$input-disabled-color: $text-light !default
+$input-disabled-color: $grey-darker !default
 $input-disabled-background-color: $background !default
 $input-disabled-border-color: $background !default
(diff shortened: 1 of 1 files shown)


jgthms:
Sounds good. Can you use `$text-strong` instead?

fabiofdsantos:
ping @13protons

PR #3245, opened 2021-01-23
Docs: Corrected sample text in typography examples

WhitWaldo:
This is a documentation fix

### Proposed solution
Examples in typography helpers reference the font-size value and not the appropriate size class. Corrected so example text shows up properly.

### Tradeoffs
Nearly inconsequential amount of additional maintenance if changes are made to typography going forward. Otherwise, none.

### Testing Done

None. Can't figure out how to build documentation locally, but since the class will use the appropriate class value now.

### Changelog updated?

No.


jgthms:
Thanks but the previous PR #3244 already fixed it. Great minds think alike. 😉

PR #3278, opened 2021-02-25
Fix modifiers link on main page

gavvvr:

### Proposed solution

Currently "See modifiers syntax" link on landing page redirects to itself.
This PR should fix the problem, but I did not test.

### Testing Done

None.

### Changelog updated?

No.


jgthms:
Thanks.

PR #3294, opened 2021-03-30
[Add] explicit self-closing tags to inputs

dotconde:
This is a improvement documentation fix.

### Proposed solution

Add self-closing tags to inputs. Code snippets from current documentation doesn't work in React.

### Tradeoffs

Instead of: ```<input>```
It should be: ```<input />```

### Testing Done

None.

### Changelog updated?

No.


jgthms:
In HTML 5 you don’t need to self close the tags.

PR #3310, opened 2021-04-17
add missing semi-colon to css import commands

apuchitnis:
This is a self-explanatory  **documentation fix**.


jgthms:
I believe it’s supposed to be outside the double quotes.

apuchitnis:
Thanks @jgthms - it's fixed now 🙂

Issue #3333, opened 2021-05-21
[Feature] SASS division without slash


thefoxie:
This is about **Bulma**.

### Overview of the problem

This is about the Bulma **CSS framework**
I'm using Bulma **version** [0.9.2]
This is a **Sass** issue: I'm using version [1.33.0]
I am sure this issue is **not a duplicate**

### Description
Usage of slash division will be deprecated in sass 2.0.0 and now my console gets polluted with deprecation warnings.
https://sass-lang.com/documentation/breaking-changes/slash-div

### Steps to Reproduce

1. Update to latest sass 1.33.0
2. Run  project

### Expected behavior
No warnings.

### Actual behavior
Warning for every slash used for division.
`DEPRECATION WARNING: Using / for division is deprecated and will be removed in Dart Sass 2.0.0.`

Niek:
Same here, might be a good idea to run https://github.com/sass/migrator on the Sass files?

tcitworld:
It seems bulma still makes use of `node-sass` instead of `dart-sass`, and `math:div` is only supported in `dart-sass`: https://sass-lang.com/documentation/modules/math

Also I encountered [this issue](https://github.com/sass/dart-sass/issues/1320) when trying to migrate, so it's probably safer to downgrade `sass` to `1.32.0`.

francesco-loreti:
Recommendation: math.div($i, 12)

robertmarney:
Full error example:

```bash
: Using / for division is deprecated and will be removed in Dart Sass 2.0.0.

Recommendation: math.div($button-padding-horizontal, 4)

More info and automated migrator: https://sass-lang.com/d/slash-div

   ╷
88 │       +ltr-property("margin", $button-padding-horizontal / 4)
   │                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   ╵
    node_modules\bulma\sass\elements\button.sass 88:31  @import
    node_modules\bulma\sass\elements\_all.sass 5:9      @import
    node_modules\bulma\bulma.sass 5:9                   @import
    resources\sass\buefy.scss 166:9                     @import
    stdin 1:9                                           root stylesheet
```

lashchivskyi:
> Повний приклад помилки:
> 
> ```shell
> : Використання / для  розділення застаріло і буде видалено  в Dart Sass 2.0.0.
> 
> Рекомендація: math.div ( $ button -padding-horizontal, 4)
> 
[...]

PR #3335, opened 2021-05-21
Change build method to use dart sass instead of node-sass

tcitworld:
This is a **new feature**. It is needed as `node-sass` [is being depreciated](https://sass-lang.com/blog/libsass-is-deprecated), and compatibility between projects using legacy `node-sass` and projects using `sass` (Dart) is going to cause issues (such as #3333). By switching to the Dart version we can then better support it.

### Proposed solution
It replaces `node-sass` with `sass` as the main way to build Bulma.

### Tradeoffs
In the long term support for `node-sass`/`libsass` will have to be removed.

### Testing Done

* Build the documentation and quick check for UI issues, and check that it doesn't reference `node-sass` as a preferred way to use Bulma anymore.
* Checked that the difference in the CSS files consists of very small changes (order changes)

### Changelog updated?

Yes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Bulma Changelog
 
+## Unreleased
+
+### Breaking change
+
+Bulma is now build with [`sass` (Dart Sass)](https://sass-lang.com/dart-sass) instead of [`node-sass`](https://github.com/sass/node-sass) ([which is depreciated](https://sass-lang.com/blog/libsass-is-deprecated)).
+
+In the future, Bulma may make use of Sass syntaxes that are currently only available for the Dart Sass version, so it is advised to migrate to `sass` instead of `node-sass` if that's what you are using to compile the source sass files. This will not affect you if you are using the CSS files directly.
 ## 0.9.2
 
 ### Breaking change
--- a/docs/_data/links.json
+++ b/docs/_data/links.json
@@ -114,13 +114,13 @@
       "icon": "cogs",
       "path": "/documentation/customize/variables"
     },
-    "customize-node-sass": {
-      "name": "With node-sass",
-      "subtitle": "Use npm/yarn and node-sass",
+    "customize-sass": {
+      "name": "With sass",
+      "subtitle": "Use npm/yarn and sass",
       "color": "danger",
       "icon_brand": "true",
       "icon": "npm",
-      "path": "/documentation/customize/with-node-sass"
+      "path": "/documentation/customize/with-sass"
     },
     "customize-sass-cli": {
       "name": "With Sass CLI",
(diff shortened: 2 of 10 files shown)


tcitworld:
Wouldn't it be easier to bump to version to 1.x with this breaking behaviour and backport everything in 0.x versions? Instead of having distinct projects, if that's what you meant? Since it seems node-sass/libsass is going to be completely abandoned in the end.

jgthms:
There is a plan of a v1 already in the pipelines. It’s not using dart, but actually it might be a good opportunity for that.

service-paradis:
> One solution would be to create 2 versions of Bulma. There could be a “bulma-dart” version of the framework, while “bulma” remains in node. And then do the switch at some point so that “bulma” would be in dart and “bulma-node” would exist as a legacy version.

I understand the point where you want to support the most user, but I'm not sure about this. Supporting 2 versions and directing people to use the correct version depending on their setup would be time consuming. Also, there are some project out there that uses Bulma under the hood (Buefy for example). They will also need to offer 2 versions?

Asking people that uses Bulma with Dart to switch package could result with a loss too. It would be more easier to use Bulma for people that uses deprecated packages. Sounds wrong to me.

> Wouldn't it be easier to bump to version to 1.x with this breaking behaviour and backport everything in 0.x versions? Instead of having distinct projects, if that's what you meant? Since it seems node-sass/libsass is going to be completely abandoned in the end.
> > There is a plan of a v1 already in the pipelines. It’s not using dart, but actually it might be a good opportunity for that.

I think this is a pretty good solution, but when is it planned? There are already a lot of people that uses the latest Dart package and have ignore a lot Deprecation warning. Maybe you could have a @next version short term (would be a good compromise with your idea of having a different project)?

cokkto:
Agree with [comment](https://github.com/jgthms/bulma/pull/3335#issuecomment-848479939) - keep 0.x version node_sass based, and make 1.x dart based. All current users will be able to continue using bulma, and when they eventually switch to dart, all what they need would be to switch to new major release

tcitworld:
Do you still want this @jgthms ?

lamka02sk:
We recently migrated back to node sass from dart sass, because dart sass is almost twice as slow. For us, 7s build times in dev mode were unacceptable. You should keep 0.x version node sass based as @cokkto suggested. 
Let's hope there will be some progress on using dart vm executable with webpack very soon.

stafyniaksacha:
> We recently migrated back to node sass from dart sass, because dart sass is almost twice as slow. For us, 7s build times in dev mode were unacceptable. You should keep 0.x version node sass based as @cokkto suggested.
> Let's hope there will be some progress on using dart vm executable with webpack very soon.

It may be because of using dart sass in js instead of dart vm : https://itnext.io/the-css-preprocessor-dilemma-node-sass-or-dart-sass-32a0a096572

tcitworld:
This has been changed a while ago.

PR #3336, opened 2021-05-21
Replace depreciated division operator with math.div function

### LINKED ISSUE #3333: [Feature] SASS division without slash

tcitworld:
This is a **bugfix**. 

It requires #3335 and will fix #3333, by replacing uses of the [depreciated division operator](https://sass-lang.com/documentation/breaking-changes/slash-div).

### Proposed solution

Replace uses of the division operator with the `math.div` function.

### Tradeoffs

Main drawback is that Bulma now requires `sass >= 1.33` to be build and used. Earlier versions of `sass` won't be able to build and `node-sass` won't be either. Since this is a very hard change, I'd recommend waiting as long as possible before merging and releasing this¹.

### Testing Done

Build the documentation and quick check for UI issues

### Changelog updated?

Will do after #3335 is merged.

¹ The depreciated message from `sass` is quite annoying, however.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Bulma Changelog
 
+## Unreleased
+
+### Breaking change
+
+Bulma is now build with [`sass` (Dart Sass)](https://sass-lang.com/dart-sass) instead of [`node-sass`](https://github.com/sass/node-sass) ([which is depreciated](https://sass-lang.com/blog/libsass-is-deprecated)).
+
+In the future, Bulma may make use of Sass syntaxes that are currently only available for the Dart Sass version, so it is advised to migrate to `sass` instead of `node-sass` if that's what you are using to compile the source sass files. This will not affect you if you are using the CSS files directly.
 ## 0.9.2
 
 ### Breaking change
--- a/docs/_data/links.json
+++ b/docs/_data/links.json
@@ -114,13 +114,13 @@
       "icon": "cogs",
       "path": "/documentation/customize/variables"
     },
-    "customize-node-sass": {
-      "name": "With node-sass",
-      "subtitle": "Use npm/yarn and node-sass",
+    "customize-sass": {
+      "name": "With sass",
+      "subtitle": "Use npm/yarn and sass",
       "color": "danger",
       "icon_brand": "true",
       "icon": "npm",
-      "path": "/documentation/customize/with-node-sass"
+      "path": "/documentation/customize/with-sass"
     },
     "customize-sass-cli": {
       "name": "With Sass CLI",
@@ -682,7 +682,7 @@
     "customize": [
       "customize-concepts",
       "customize-variables",
-      "customize-node-sass",
+      "customize-sass",
       "customize-sass-cli",
       "customize-webpack"
     ],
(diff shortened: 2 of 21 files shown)


service-paradis:
Maybe some of those could use multiplication instead of a division. It would be compatible with older libsass while removing a lot of deprecation warning for dart. Ex.

```sass
margin-right: $block-spacing / 2
```

Would become:

```sass
margin-right: $block-spacing * 0.5
```

Instead of:

```sass
margin-right: math.div($block-spacing, 2)
```

EtienneBruines:
@service-paradis but if it only applies to "some", then measures still need to be taken for those cases where division cannot be replaced by multiplication.

service-paradis:
> @service-paradis but if it only applies to "some", then measures still need to be taken for those cases where division cannot be replaced by multiplication.
> 
> (And it therefore would not make too much sense, replacing division by multiplication _some_ of the cases.)

Sure, but it would (at least) reduce some of the warning waiting for a solution.

Another way to reduce the number of warnings would be to create a custom function like:

```sass
@function bulmaRgba($number, $div)  
  @return $number / $div
```

A single warning would be issued and it would be easier to replace with `math.div` when @jgthms wants to offer a Dart version.

ghiscoding:
> Sure, but it would (at least) reduce some of the warning waiting for a solution.

in latest version `1.34.0` of dart-sass, they added a flag `--quiet-deps` to remove warnings from other lib/project/dependencies, I just tried it and it works fine. After running a build, I see warnings from my lib but without seeing all the ones from Bulma (while previous version `1.33.0` was displaying **a lot** of warnings from every dependencies it was a nightmare to parse through but now it's fixed, they had lot of people complaining and they addressed it quickly). See latest dart-sass [v1.34.0](https://github.com/sass/dart-sass/releases/tag/1.34.0)

service-paradis:
> in latest version 1.34.0 of dart-sass, they added a flag --quiet-deps to remove warnings from other lib/project/dependencies, I just tried it and it works fine. After running a build, I see warnings from my lib but without seeing all the ones from Bulma (while previous version 1.33.0 was displaying a lot of warnings from every dependencies it was a nightmare to parse through but now it's fixed, they had lot of people complaining and they addressed it quickly). See latest dart-sass v1.34.0

The warning are annoying and I'm glad `dart-sass` provides a way to quiet those messages, but this is not a good workaround. I dont like to see those warning, but I dont want to silence them. They are a reminder that I wont be able to update to the next major version of `dart-sass` if Bulma does not offer a way to be compatible with the updated syntax.

service-paradis:
Bootstrap have a PR around this problem that might be a goop option for Bulma as well:
https://github.com/twbs/bootstrap/pull/34245

tcitworld:
Closed in favor of #3362

PR #3368, opened 2021-06-28
Fixes #3316 | fixes $custom-shades

### LINKED ISSUE #3316: custom shades are returning as a whole map and not working properly
Hi everyone
As there are certain prebuilt bulma shades given at utilities/derived-variables.sass

**Overview of the problem**
`$shades: mergeColorMaps(("black-bis": $black-bis, "black-ter": $black-ter, "grey-darker": $grey-darker, "grey-dark": $grey-dark, "grey": $grey, "grey-light": $grey-light, "grey-lighter": $grey-lighter, "white-ter": $white-ter, "white-bis": $white-bis), $custom-shades) !default`

there is scope to define custom shades same as custom-colors but it is returning a whole list due to automatically taking all four values (normal.dark,light,invert)

**Steps to Reproduce**
Define new variable named `$grey-custom` and `$custom-shades` and assign map value to that before importing bulma sass library.
```
$grey-custom: hsl(0, 0%, 25%) !default
$custom-shades: ("grey-custom": $grey-custom)
```

**Expected behavior**
class **has-background-grey-custom** should output only
`background-color: #404040 !important;`

**Actual behavior**
`background-color: #404040, #fff, whitesmoke, #878787 !important;`

niuage:
This is a bug fix.

Issue: https://github.com/jgthms/bulma/issues/3316

### Proposed solution

It fixes 1 issue, and adds 1 improvement.

#### The issue:

You can read about the issue here: https://github.com/jgthms/bulma/issues/3316, but the tldr is that the function `mergeColorMaps` wasnt working properly when used with `$shades` and `$custom-shades`.

The function was merging a list for custom shades: `$value: ($color-base, $color-invert, $color-light, $color-dark)`.
So, when it was used in `colors.sass`, and the list was the value of the `background` property, it was breaking.

```
@each $name, $shade in $shades
  .has-text-#{$name}
    color: $shade !important
  .has-background-#{$name}
    background-color: $shade !important
```

The proposed solution is to leave the function `mergeColorMaps` as is, but to take into account that the values could be lists when using it with $shades.

```
@each $name, $shade in $shades
  // NEW
  @if type-of($shade) == "list"
     $shade: nth($shade, 1)

  .has-text-#{$name}
    color: $shade !important
  .has-background-#{$name}
    background-color: $shade !important
```

#### Improvement

Another tiny improvement is the addition of `$grey-lightest` in `$shades`. My guess is that it was an involuntary omission, but if it wasn't I can simply revert that change.

### Tradeoffs

No trade-offs.

### Testing Done

I tested that the generated classes for $custom-shades work properly. I also checked that the classes generated for the default $shades work too.

I tested that the generated classes for $custom-shades work properly in my own project by modifying the code in node_modules directly. I also checked that the classes generated for the default $shades work too.

### Changelog updated?

No.

Thanks!

### DIFF
--- a/sass/helpers/color.sass
+++ b/sass/helpers/color.sass
@@ -33,6 +33,8 @@
       background-color: $color-dark !important
 
 @each $name, $shade in $shades
+  @if type-of($shade) == "list"
+     $shade: nth($shade, 1)
   .has-text-#{$name}
     color: $shade !important
   .has-background-#{$name}
--- a/sass/utilities/derived-variables.sass
+++ b/sass/utilities/derived-variables.sass
@@ -109,6 +109,6 @@ $custom-shades: null !default
 
 $colors: mergeColorMaps(("white": ($white, $black), "black": ($black, $white), "light": ($light, $light-invert), "dark": ($dark, $dark-invert), "primary": ($primary, $primary-invert, $primary-light, $primary-dark), "link": ($link, $link-invert, $link-light, $link-dark), "info": ($info, $info-invert, $info-light, $info-dark), "success": ($success, $success-invert, $success-light, $success-dark), "warning": ($warning, $warning-invert, $warning-light, $warning-dark), "danger": ($danger, $danger-invert, $danger-light, $danger-dark)), $custom-colors) !default
 
-$shades: mergeColorMaps(("black-bis": $black-bis, "black-ter": $black-ter, "grey-darker": $grey-darker, "grey-dark": $grey-dark, "grey": $grey, "grey-light": $grey-light, "grey-lighter": $grey-lighter, "white-ter": $white-ter, "white-bis": $white-bis), $custom-shades) !default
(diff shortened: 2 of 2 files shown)


jgthms:
Shouldn't we fix the `mergeColorMaps` function instead?

niuage:
The way I was looking at it was that the `mergeColorMaps` function shouldn't have to worry about the way it was going to be used, its only responsibility is to merge 2 maps, and to generate the color variations of each of the given map values.

If it was me, I would either:
- leave it as is, and let the different places where it's used worry about which of the color variations they care about,
- or create another function altogether that would be drastically simpler and to the point, as it would know that the map values are simple values and not lists.

Adding an option to the `mergeColorMaps` to indicate we dont want to generate the color variations (or something to this effect) is definitely a possibility, but I thought it would complicate the function even more.

But really, it's up to you, and if you let me know which you prefer, I can try to implement it, or you can knock it out yourself real quick. All I care about is for this bug to be fixed :)

niuage:
@jgthms what's your opinion?

niuage:
@jgthms do you have an ETA on when you'll be to have time to look at that?

PR #3406, opened 2021-09-10
Remove auto padding spacing helpers

lionzeye:
This is a **bugfix**.

I came across this issue while working on https://github.com/jgthms/bulma/pull/3184
The bulma css output contains spacing helpers for padding with the value of 'auto'.
This is not a valid value, as it should be either a length or a percentage.

### Proposed solution

This pull request makes sure that the spacing SASS skips spacing helper generation for auto padding spacing helpers. I'm not 100% convinced that this is the cleanest solution, so I'm of course open to improvements.

### Tradeoffs

None

### Testing Done

Build works, outputs are identical, with the exception of the invalid helpers.

### Changelog updated?

No.


lionzeye:
Now included in https://github.com/jgthms/bulma/pull/3184

PR #3413, opened 2021-09-20
feath: bulma jquery

rastclick:
I wrote the jquery code for bulma
Bulma-jQuery is extension for the Bulma CSS framework that include all jquery code you need.

https://github.com/rastclick/bulma-jquery


jgthms:
Ok thanks but I don't think you intended to make a PR. I wll close this.

PR #3467, opened 2022-01-02
@use instead of @import. Modules using * rather than scoping.

jrobens:
This is a ** improvement **.

Use @use instead of @import to help facilitate the customisation and extension of bulma. This allows you to use @use to bring bulma into the site stylesheet. This provides better visibility of what is being imported. Bulma components are scoped as modules.

### Proposed solution

Replaces the @import with @use 

### Tradeoffs

- Minimal changes using "*" instead of scoping modules or specifically collecting variables that should/could be used in a site stylesheet e.g. initial_variables are likely to be overridden and derived_variables consumed in an expected derivative theme.
- Renamed _all to all_modulename instead of scoping 

### Testing Done

Local site testing. Conceptual start, discuss implementation and how this makes using Bulma better.

### Changelog updated?

No.


jgthms:
This requires Dart Sass, which most Bulma users don't have yet. But it's planned to move to that at some point.

daniil4udo:
@jrobens that has been implemented in [@bulvar/bulma](https://github.com/daniil4udo/bulvar/issues/38). install `@bulvar/bulma@0.10.1-beta.2` and be my guest to test it

PR #3476, opened 2022-01-23
Display the result of using text transformers

code913:

### Proposed solution

Instead of showing all the transformed text as bold (default styling of `<strong>`) it might be better to use the appropriate class name to show the effect of using that class.

### Testing Done

None.

### Changelog updated?

No.


jgthms:
Nice one. Thanks.

PR #3481, opened 2022-01-26
Typo fix

jbierfeldt:

### Proposed solution

Typo fix

### Changelog updated?

No.

### DIFF
--- a/docs/documentation/form/general.html
+++ b/docs/documentation/form/general.html
@@ -833,7 +833,7 @@ variables_keys:
 
 <div class="content">
   <p>
-    To keep the design evenly balanced, Bulma provides a very useful <code>control</code> container with which you can <strong>wrap</strong> the form controls.
+    To maintain an evenly balanced design, Bulma provides a very useful <code>control</code> container with which you can <strong>wrap</strong> the form controls.
     <br>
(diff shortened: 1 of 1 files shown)


jgthms:
“maintain” here means to keep going, continue, so I think it’s valid.

jbierfeldt:
If you want to keep maintain, then you could change it to: "To maintain an evenly balanced design, ..."

The current version "To maintain the design evenly balanced" sounds very strange to my ear (U.S. American English), and reads like a typo. But of course maybe it sounds fine in other types of English! Just wanted to raise awareness in case it was a typo.

jgthms:
You're right, the original sentence wasn't great. We could go with `To provide an evenly balanced design` since the sentence is not really a continuation of a prior statement anyway. What do you think?

jbierfeldt:
I think that's perfect!

jbierfeldt:
I actually think that "maintain" is the best option here, since "provide" is used immediately in the next clause. Also, the use of "maintain" in ```To maintain an evenly balanced design``` has the connotation of ```(To allow you, the user,) to maintain an evenly balanced design```.

jgthms:
Ok let's go with that. Thanks for your help.

PR #3513, opened 2022-04-26
Fixes #3383 ('Variables' sections on docs page)

### LINKED ISSUE #3383: 2 Variables sections on the Variables docs page
This is about **the Docs**.

### Overview of the problem

This is about the Bulma **Docs**
My **browser** is: Firefox

There are 2 `Variables` sections on the [variables page](https://bulma.io/documentation/customize/variables/#variables) with the same ID's.

### Description

There is a mistake on the [variables page](https://bulma.io/documentation/customize/variables/#variables). There are 2 variables sections with the  same ID's but different content. There are 2 `Variables` buttons in the right hand menu which both go to the first variables section but only the bottom one light up when scrolling onto either of the variables sections.

Either they should be given different ID's (and some spacing added between them) or they could be merged together into one section.

I believe [this is the related page on GitHub](https://github.com/jgthms/bulma/blob/master/docs/documentation/customize/variables.html).

### Steps to Reproduce

1. Open the Bulma [variables page](https://bulma.io/documentation/customize/variables)
2. Click one of the two `Variables` buttons in the menu on the right hand side of the page
3. or scroll down to one of the `Variables` sections

### Expected behavior

The two sections should be treated as separate sections (with padding between them) or be merged into one.

Screenshot_20210723_133418

avt613:

Related to issue  #3383

There are 2 Variables sections on the [variables page](https://bulma.io/documentation/customize/variables/#variables) with the same ID's and no gap between them. The right hand menu shows 2 Variable sections which both take you to the first one of the 2 variable sections.

### Proposed solution

1. Give each variables section a different name, by using the `anchor_name` as the section title
2. Add gap between consecutive variable 'divs' (identified by class="bd-vars")

### Tradeoffs

None that I can see.

### Testing Done

Some...
Fixes problem on the [variables page](https://bulma.io/documentation/customize/variables/#variables), but may cause unintended issues on other pages which use the modified file if they have an `anchor_name` set which should not be used for the section title.

### Changelog updated?

No.

### Screenshot
image

### DIFF
--- a/docs/_includes/components/variables.html
+++ b/docs/_includes/components/variables.html
@@ -15,7 +15,7 @@
 
 {% if variables_size != '0' or variables_keys_size != '0' %}
   <div class="bd-vars">
-    {% include elements/anchor.html name="Variables" %}
+    {% include elements/anchor.html name=anchor_name %}
 
     <div class="bd-var bd-is-head">
       <div class="bd-var-name">
--- a/docs/_sass/components/variables.scss
+++ b/docs/_sass/components/variables.scss
@@ -16,6 +16,10 @@
   }
 }
 
+.bd-vars + .bd-vars {
+  margin-top: var(--docs-inner);
+}
+
 .bd-var {
   display: flex;
   flex-wrap: wrap;


avt613:
Thanks for merging.
The scss for the [Bulma docs](https://bulma.io/documentation/customize/variables/#generic-variables) has not been recompiled yet. 
Is this done automatically or does it need to be done manually? (The other change has already taken affect)

PR #3531, opened 2022-05-27
Remove full width inconsistency

kylet91:

This was / is annoying the hell out of me when browsing the documentation, when I would go to "hero" or "section" pages, the whole layout would break because it was fullwidth.

Very tiny PR, but also very annoying problem.

### DIFF
--- a/docs/documentation/layout/container.html
+++ b/docs/documentation/layout/container.html
@@ -1,7 +1,6 @@
 ---
 title: Container
 layout: documentation
-fullwidth: true
 hide_categories: true
 doc-tab: layout
 doc-subtab: container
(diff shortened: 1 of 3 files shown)


jgthms:
These components are fullwidth so the documentation needs to allow them to be fullwidth.

kylet91:
That makes sense, but it looks (IMO at least) like the page is broken when the nav bar suddenly disappears. Where with the nav bar still there, it still gets the point across that this takes up the entire width, but doesn't look broken.

image

Obviously if this is intentional and you don't think it's a problem, no stress, happy to close this PR. :)

jgthms:
Yes this issue was brought up from time to time but I think visitors get used to it in the long run. 

The problem with keeping the sidebars is that the components don’t have enough space to look how they should. It would also give the wrong impression of how they would actually be used (spanning the full width of the viewport).

PR #3542, opened 2022-06-11
Add flex gap helper : Fixes #3414

### LINKED ISSUE #3414: Flexbox helpers : is-flex-gap-* proposal
This is about **Bulma**.

It's a feature proposal

### Overview of the problem

I'm using Bulma **version** [0.9.1]

### Description

I'm using flexbox helpers pretty often and only one thing is missing that force me to define classes and add specific CSS to my code is the flex gap ! 

It would be really useful to take advantage of $spacing-values to define a variety of gaps for flexbox helper.

Technically, it would be as simple as adding the equivalent of this loop in flexbox helper: 

```scss
$spacing-values: ("0": 0, "1": 0.25rem, "2": 0.5rem, "3": 0.75rem, "4": 1rem, "5": 1.5rem, "6": 3rem) !default;

@each $name, $value in $spacing-values {
  .is-flex-gap-#{$name} {
    gap: $value !important;
  }
}
```

to generate helpers like `is-flex-gap-0` to `is-flex-gap-6`

Resulting example 
```html
<header class="modal-card-head is-flex-direction-column is-align-items-flex-start is-flex-gap-4">
            <div class="is-flex is-align-items-center is-justify-content-space-between is-flex-gap-2">
```

What do you think about this upgrade ? If it makes sens, I can propose a PR if you want. Let me know.

Alii-isk:
This PR adds a "flexbox gap" helper, which can be used like this "is-flex-gap-x"
i also added the corresponding documentation for this.

### Proposed solution
this fixes this issue #3414 

### Testing Done

None.

### How have you confirmed this feature works?
yes, I run `yarn deploy` and tested the `bulma.css` and everything is working as expected.

### Changelog updated?

No.

### DIFF
--- a/docs/documentation/helpers/flexbox-helpers.html
+++ b/docs/documentation/helpers/flexbox-helpers.html
@@ -62,7 +62,8 @@ align-self-values:
 
 <div class="content">
   <p>
-    Combined with <code>is-flex</code>, all of the <strong>Flexbox CSS properties</strong> are available as Bulma helpers:
+    Combined with <code>is-flex</code>, all of the <strong>Flexbox CSS properties</strong> are available as Bulma
+    helpers:
   </p>
   <ul>
     <li><code>flex-direction</code></li>
@@ -73,6 +74,7 @@ align-self-values:
     <li><code>align-self</code></li>
     <li><code>flex-grow</code></li>
     <li><code>flex-shrink</code></li>
+    <li><code>gap</code></li>
   </ul>
 </div>
 
@@ -102,6 +104,7 @@ align-self-values:
 
 {% include elements/anchor.html name="Flex grow and flex shrink" %}
 
+
 <table class="table is-bordered">
   <thead>
     <tr>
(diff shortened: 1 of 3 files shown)


tkvlnk:
are u sure u need `is-flex-gap-x` if there is already [https://github.com/jgthms/bulma/blob/43b879db25d86d630e3e5c6282421b696b31ce33/css/bulma.css#L20751](`is-gap-x`) available

jgthms:
It's been shipped with v1 https://github.com/jgthms/bulma/blob/main/sass/helpers/gap.scss

Alii-isk:
> are u sure u need `is-flex-gap-x`?
> there is already [`is-gap-x`](https://github.com/jgthms/bulma/blob/43b879db25d86d630e3e5c6282421b696b31ce33/css/bulma.css#L20751) available (which is also used for grids)
> 
> PS it would be good to have it mentioned in docs though
> 
> 

This PR was 2 years ago

tkvlnk:
> > are u sure u need `is-flex-gap-x`?
> > there is already [`is-gap-x`](https://github.com/jgthms/bulma/blob/43b879db25d86d630e3e5c6282421b696b31ce33/css/bulma.css#L20751) available (which is also used for grids)
> > PS it would be good to have it mentioned in docs though
> 
> This PR was 2 years ago

ur right
I missed the dates 

but anyways it should've been closed and not contribute to issues counter 🤷

PR #3544, opened 2022-06-17
Fix usage of ltr/rtl padding for notifications (Fixes #3025)

### LINKED ISSUE #3025: Overriding the $notification-padding variable only works when also overriding the notification-padding-ltr and notification-padding-rtl variables as well
This is about **Bulma**.

### Overview of the problem

This is a **Sass** issue: I'm using version [0.9.0]

### Description

It seems that adjusting the notification padding back to the default from previous versions of Bulma (I believe before 0.8.0 when all the paddings/margins changed w/ the larger form controls update) no longer works. I have customized Bulma quite a lot and have plenty of custom colors, variables and so on so I know my SASS file is working fine, but since upgrading to 0.9.0 I am unable to update the notification padding without also updating the variables for `ltr` and `rtl`. Should the ltr & rtl variables be derived perhaps? 

### Steps to Reproduce
1. Create a sass file per the docs in order to customize Bulma
2. Override the `$notification-padding` variable, a la `$notification-padding: 0.375rem 0.75rem;`
3. Import Bulma as normal: `@import "~bulma/bulma";`

### Expected behavior

Notification padding should be set to `0.375rem 0.75rem`

### Actual behavior

Notification padding remains at the default value in `~bulma/bulma/sass/elements/notification.sass`: `$notification-padding: 1.25rem 2.5rem 1.25rem 1.5rem !default`

This can be overcome by also overriding the other padding variables in this file, but this is not immediately clear to people who had adjusted it prior to the addition of the `ltr` and `rtl` variables:
```
$notification-padding: 0.375rem 0.75rem;
$notification-padding-ltr: 0.375rem 0.75rem;
$notification-padding-rtl: 0.375rem 0.75rem;
```
Perhaps it would be good to derive the `ltr` and `rtl` variables from the main one, or set a boolean to allow this just so we don't have to override all three. In writing up the details for this issue I figured out the solution, but I believe this should either be made clear or there might be an opportunity for an enhancement at least. Thanks!

shreve:
This is a **bugfix**. Addresses #3025 

### Proposed solution

Unify the usage of ltr and rtl padding values behind one variable, `$notification-padding`.

Currently, `$notification-padding` is set, but isn't used at all. The real padding is written using `+ltr` / `+rtl` mixins, and using the specific values `$notification-padding-ltr` and `$notification-padding-rtl`. 

This obscures what we really want to do: __Offer a configurable value for padding, and set an appropriate default based on the value of `$rtl`.__

To correct this, we create a 4th variable, `$default-notification-padding`, which is populated by `@if $rtl`, then set the default (replaceable) value of `$notification-padding` with this new variable.

### Tradeoffs

- Add a new variable that isn't necessarily worth documenting, and makes this definition more verbose
- Add an insignificant amount of compute time during render

### Testing Done

- [x] Pull the latest `master` branch
- [x] Make sure your Sass code is compliant with the [Bulma Sass styleguide](https://github.com/jgthms/bulma/blob/master/.github/CONTRIBUTING.md#bulma-sass-styleguide)
- [x] Make sure your PR only affects `.sass` or documentation files
- [x] [Try your changes](https://github.com/jgthms/bulma/blob/master/.github/CONTRIBUTING.md#try-your-changes).

The following code blocks are pairs of (test case `bulma.sass`, and the output of `grep -A4 "\.notification {" css/bulma.css` after running `npm run build`):

```bulma.sass
@charset "utf-8"
/*! bulma.io v0.9.4 | MIT License | github.com/jgthms/bulma */
@import "sass/utilities/_all"
@import "sass/elements/notification.sass"
```
```css/bulma.css
.notification {
  background-color: whitesmoke;
  border-radius: 4px;
  position: relative;
  padding: 1.25rem 2.5rem 1.25rem 1.5rem;
```

Here we see the default still stays ready for ltr with a delete button on the right.

```bulma.sass
@charset "utf-8"
/*! bulma.io v0.9.4 | MIT License | github.com/jgthms/bulma */
@import "sass/utilities/_all"
$rtl: true
@import "sass/elements/notification.sass"
```

```css/bulma.css
.notification {
  background-color: whitesmoke;
  border-radius: 4px;
  position: relative;
  padding: 1.25rem 1.5rem 1.25rem 2.5rem;
```

When `$rtl` is enabled, the padding is now updated without any other changes.

```bulma.sass
@charset "utf-8"
/*! bulma.io v0.9.4 | MIT License | github.com/jgthms/bulma */
@import "sass/utilities/_all"
$rtl: true
$notification-padding: 0px
@import "sass/elements/notification.sass"
```

```css/bulma.css
.notification {
  background-color: whitesmoke;
  border-radius: 4px;
  position: relative;
  padding: 0px;
```

Now even with `$rtl` enabled, the output is still the specified value of `$notification-padding`.

### Changelog updated?

No.

### DIFF
--- a/sass/elements/notification.sass
+++ b/sass/elements/notification.sass
@@ -3,10 +3,14 @@
 $notification-background-color: $background !default
 $notification-code-background-color: $scheme-main !default
 $notification-radius: $radius !default
-$notification-padding: 1.25rem 2.5rem 1.25rem 1.5rem !default
 $notification-padding-ltr: 1.25rem 2.5rem 1.25rem 1.5rem !default
 $notification-padding-rtl: 1.25rem 1.5rem 1.25rem 2.5rem !default
 
+$default-notification-padding: $notification-padding-ltr
+@if $rtl
+  $default-notification-padding: $notification-padding-rtl
+
+$notification-padding: $_default-notification-padding !default
 $notification-colors: $colors !default
 
 .notification
@@ -14,10 +18,7 @@ $notification-colors: $colors !default
   background-color: $notification-background-color
   border-radius: $notification-radius
   position: relative
-  +ltr
-    padding: $notification-padding-ltr
-  +rtl
-    padding: $notification-padding-rtl
+  padding: $notification-padding
   a:not(.button):not(.dropdown-item)
     color: currentColor
     text-decoration: underline


jgthms:
I don't see how this is an improvement? It's simply putting the `@if $rtl` logic in the variable definition, rather than in the property definition?

One thing I agree on is to remove `$notification-padding` which isn't used anymore.

shreve:
It re-enables `$notification-padding` and makes its default behave sensibly, but that's only an improvement if that is the goal (which is my preference).

If your preference is to remove `$notification-padding`, I suppose it's not an improvement. If so, I can change this PR to simply remove it from the SASS and the documentation. The real upstream problem is that I was misled by the documentation and either of these options fixes that.

PR #3567, opened 2022-09-02
navbar documentation closing tag added

ebraj:
There is no closing tag or (/) in the navbar docs page in the `image` and `hr` tag.

Screen Shot 2022-09-02 at 9 45 58 PM

Have added it on all the `image` and `hr` tag.


jgthms:
It's not required when using the HTML5 syntax.

PR #3579, opened 2022-09-30
Uppercase text in breadcrumbs

alagaccia:
Sometimes the breadcrumbs appear to be uppercase. It would be nice to make to developer choose this parameter.

### Proposed solution

Just add a `text-transform: $breadcrumb-text-transform` inside the .breadcrumb class
Default value for `$breadcrumb-text-transform` is `none`

### Testing Done

None.

### Changelog updated?

No.

### DIFF
--- a/sass/components/breadcrumb.sass
+++ b/sass/components/breadcrumb.sass
@@ -1,5 +1,6 @@
 @import "../utilities/mixins"
 
+$breadcrumb-text-transform: none !default 
 $breadcrumb-item-color: $link !default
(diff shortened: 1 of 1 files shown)


jgthms:
In which situation would they be uppercase?

alagaccia:
> In which situation would they be uppercase?

A lot of websites and programms use uppercase breadcrumbs because of their tiny stiyle. It'd be just a choice for the developer. The default value would be "none" with no conflicts.

PR #3591, opened 2022-11-10
fix: Fix fixed font size on html tag breaking accessibility Fixes #3588

### LINKED ISSUE #3588: Accessibility issue with bulma reset
This is about **Bulma**.

### Overview of the problem

### Description

Bulma reset file has a `font-size: 16px;` declaration in the `html`, this breaks the accessibility of websites for users who have vision impairment and scale up the font size of the browser.

### Steps to Reproduce

1. set your font with relative units
2. update your browser font-size preferences
3. check your website

### Expected behavior

The font size of the entire application (and all other relative set sizes) should scale up with the chosen size

### Actual behaviour

Nothing changes since the `html` font-size are set in pixels

Eduruiz:
This is a **bugfix**.

Fix #3588

### Proposed solution

Update font size of `html` tag to relative value `1em`

### Tradeoffs

None that I'm aware of 

### Testing Done

Just checked the value after building the file.

### Changelog updated?

No.


package.json
```diff
@@ -1,19 +1,19 @@
 {
-  "name": "bulma",
+  "name": "bulma-no-reset",
```
jgthms, line 2: You can't edit this file.

Issue #2845, opened 2020-02-24
Hero with is-fullheight-with-navbar combined with is-spaced navbar


robbertkl:
A hero with `is-fullheight-with-navbar` only accounts for a normal fixed navbar, not for a `is-spaced` navbar.

I've managed to work around it by adding this to my scss:

```scss
.is-fullheight-with-spaced-navbar {
  @extend .is-fullheight-with-navbar;
}
@media screen and (min-width: 1024px) {
  .hero.is-fullheight-with-spaced-navbar {
    min-height: calc(100vh - 5.25rem);
  }
}
```

Perhaps a class like this could be incorporated into bulma?

maduhaime:
This is my take on the same problem, with the use of scss and variables.

```scss
/* 
Hack for missing modifier for fullheight hero with spaced navbar.
By default, .is-spaced has no effect on Navbar padding.  Only from $tablet.
*/
.hero {
  &.is-fullheight-with-spaced-navbar
    @extend .is-fullheight-with-navbar;
  }
  @include from($tablet) {
    min-height: calc(100vh - $navbar-height + (2 * $navbar-padding-vertical));
  }
}
```

PR #3623, opened 2023-02-10
Added is-fullheight-with-spaced-navbar modifier for navbars that use is-spaced.

rayment:
This is an **improvement**.

### Proposed solution

Currently the hero layout with `is-fullheight-with-navbar` fails to account for `is-spaced` navbars resulting in an extra space at the bottom of the page. This PR adds a new modifier `is-fullheight-with-spaced-navbar` so that the spacing is accounted for on desktop and higher.

This problem has already been addressed in issue #2845 which in turn ended up going stale.

### Tradeoffs

Users must now decide between `is-fullheight-with-navbar` and `is-fullheight-with-spaced-navbar` but we already differentiate between fixed-bars and regular bars so I don't see this as an unnecessary complication.

#### Minimal example:

```
<nav class="navbar is-light is-spaced" role="navigation">
  <div class="navbar-brand">
  </div>
</nav>
<section class="hero is-info is-fullheight-with-spaced-navbar">
  <div class="hero-body">
  </div>
</section>
```

Note how the page does not scroll when `with-spaced-navbar` is used.

### Changelog updated?

No.

### DIFF
--- a/docs/documentation/layout/hero.html
+++ b/docs/documentation/layout/hero.html
@@ -183,6 +183,9 @@ meta:
   <p>
     If you are using a <a href="{{ site.url }}/documentation/components/navbar/#fixed-navbar">fixed navbar</a>, you can use the <code>is-fullheight-with-navbar</code> modifier on the hero for it to occupy the viewport height minus the navbar height.
   </p>
+  <p>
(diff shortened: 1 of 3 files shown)


rayment:
@jgthms ping?

PR #3624, opened 2023-02-12
Fix form field-label alignment for rtl

idank:
This is a bugfix for a small rtl issue with field-label. It should align the other way.

Before:

image

After:

image

### Testing Done

Visual inspection locally.

### Changelog updated?

No.


idank:
ping

PR #3632, opened 2023-03-16
Improve text weight typography helpers docs display

rrosenbergpax8:

When using the helpers I wasn't sure which one I wanted because they were all just wrapped in `strong`.
Instead I think it would be nice for folks to be able to see the difference in text weight in the docs and see the helper applied in action.

### Proposed solution

Make the weight name match the weight helper it's aligned with.

CleanShot 2023-03-16 at 11 53 17

### Tradeoffs
Could possible affect readability.

### Changelog updated?

No.


jgthms:
Nice, thank you.

Issue #3604, opened 2022-12-21
Importing bulma module in Nuxt 3 app causes super slow Vite startup


mohasi:
This is about **Bulma**.

### Overview of the problem

This is about the Bulma **CSS framework** 
I'm using Bulma **version** [0.9.4]
My **browser** is: NA
This is a **Sass** issue: I'm using version [0.9.4] 
I am sure this issue is **not a duplicate**? 

### Description

When adding bulma to Nuxt modules, vite server starts up really slowly, up to a minute for some users. It has already been raised as Nuxt and Vite github issues but devs there have said bulma is the cause.

One interesting comment (https://github.com/nuxt/framework/issues/4613#issuecomment-1339379663) says bulma uses old division code from Bootstrap, which is the underlying cause. Please investigate and issue fix if that is the case.

### Steps to Reproduce

Create new Nuxt project
Start vite server, it starts up within a few seconds
Now add bulma to modules
Start vite server, now it takes anywhere from 20 seconds to a minute

Please see https://github.com/vitejs/vite/issues/11221 and https://github.com/nuxt/framework/issues/4613

### Expected behavior

Vite should start up quickly, as that is one of its major features.

### Actual behavior

Vite starts up real slow (possibly due to some division loops within bulma code -- see description)

mohasi:
Anyone look at these issues?

dword-design:
Same issue here!

Jeroen-Cox:
Same issue!

PR #3641, opened 2023-04-27
Replace custom divide function with sass built-in math.div.

SoCuul:
This is an **improvement**.

### Proposed solution

Currently, bulma uses a custom implemented `divide` function in multiple functions. The main downside of using this custom division function, is that it uses multiple loops that severely slow down build speeds on the final app.

This has been reported previously at #3604, and was been investigated in a separate thread.

This pull request replaces the custom `divide` function logic with the built-in sass `math.div` function. This change results in 20x quicker build times with nuxt + vite.

### Tradeoffs

Because node-sass has not been updated, there is no support for the `math.div` function. To work around this issue, I've replaced all references with `node-sass` to the `sass` npm package. The functionality is nearly identical, and there should be no modification necessary to any workflows.

### Testing Done

I have built both the main `css` files along with the `rtl-css` files. Both returned perfect results, and no errors were displayed during building.

### Changelog updated?

No.


jgthms:
So the custom divide was introduced to add compatibility with dart sass, which has another way to make divisions but requires changing the whole codebase. 

Bulma v1 which I’m still working on will still be compatible with node sass, but with v2 it will only be dart sass which means using the built in divide again.

SoCuul:
Is there an estimated release date for Bulma v2?

dword-design:
Also looking forward to this!

lctech-bill:
Could this PR will be merged?
`Nuxt3` with `sass`/`scss` is currently facing slow build times and I'm looking for a solution.

[Slow nuxt startup when importing sass file](https://github.com/nuxt/nuxt/issues/13824#issuecomment-1397319301)

Xaaalera:
+1

PR #3837, opened 2024-06-11
Fix bulma color brightness

bagedevimo:
This is a **bugfix**.

### Proposed solution

Previously, the bulmaColorBrightness function was understating the value of the blue channel when estimating brightness for a specific colour, due to a missing set of brackets. This just corrects the brackets so that all channels of the colour are weighted, instead of just red and green.

Additionally, I've switched from the slash division operation which is deprecated in Sass, to the math.div operator.

### Tradeoffs

Some folks may be relying on the slightly odd behaviour, though with blue being the least significant channel for brightness this will be an edge case and will likely be a producing confusing rules with the current implementation.

### Testing Done

Ran the docs site and visually compared a variety of buttons, panels, box, and theme pages to ensure no difference. Using the default theming values, validated that there are no differences before and after this change in the output css.

### Changelog updated?

No, but happy too if that's a good idea.

### DIFF
--- a/sass/utilities/functions.scss
+++ b/sass/utilities/functions.scss
@@ -184,10 +184,10 @@
 
 @function bulmaColorBrightness($n) {
   $color-brightness: round(
-    (red($n) * 299) + (green($n) * 587) + (blue($n) * 114) / 1000
+   ((red($n) * 299) + (green($n) * 587) + (blue($n) * 114)) / 1000
   );
   $light-color: round(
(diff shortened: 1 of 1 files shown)


jgthms:
Oh wow, good catch! That was definitely an error, since the W3 recommendation I was refering to does have the extra set of brackets: https://www.w3.org/TR/AERT/#color-contrast

The formula should be:

`((Red value X 299) + (Green value X 587) + (Blue value X 114)) / 1000`

This correction will definitely change things. And I wonder if my error did alter my judgment when developing v1.

I'm gonna check out your branch and see how this looks. Many thanks in any case.

Issue #3799, opened 2024-04-23
Columns is-variable not working


JoeyRumble:
This is about **Bulma & the Docs**.

### Overview of the problem

This is about the Bulma **CSS framework**
This is about the Bulma & Docs
I'm using Bulma 1.0
My **browser** is: Chrome

### Description

The columns 'is-variable' class with the different column gap class modifiers are not working. Neither on the document page (https://bulma.io/documentation/columns/gap/).

willoma:
I can confirm the behaviour.

I have tracked this down to https://github.com/jgthms/bulma/blob/72e99d1bf206e0c0a697b926282a0bea5dd23059/sass/grid/_index.scss#L4

This line imports [columns.scss](https://github.com/jgthms/bulma/blob/72e99d1bf206e0c0a697b926282a0bea5dd23059/sass/grid/columns.scss), however the `is-variable` stuff is in [columns-v2.scss](https://github.com/jgthms/bulma/blob/72e99d1bf206e0c0a697b926282a0bea5dd23059/sass/grid/columns-v2.scss).

janosrusiczki:
OK, so I'm not the only one.

It doesn't seem to be working in their documentation either: https://bulma.io/documentation/columns/gap/#variable-gap

I'm not exactly sure what should be happening, probably one should be able to click the .is-x class names and the example below should modify accordingly, but it doesn't work.

So I guess I'll be waiting until this gets fixed.

phil-w:
Yeah I can't get the columns to work very well at all, in general. The GRID seems more like it might be working.

reneedraper:
I did a little investigating...it looks like 

1. it's not being forwarded in `sass/grid/_index.scss` 
2. it's not included in the parse-vars script `docs/_scripts/parse-vars.js`

There's also something with the column-gap not being overwritten by the new assigned values, I'll probably dig into this more next week.

PR #3845, opened 2024-06-21
Fixes 3779

reneedraper:
This is a **bugfix**.

Referenced in issue [here](https://github.com/jgthms/bulma/issues/3799). columns-v2 is not being forwarded and it's a simple solution.

### Proposed solution

Forward columns-v2. This is a fix for the use of is-variable, **but** it doesn't fix the full documentation since it looks like there is a missing script referenced in klmn.html.

### Tradeoffs

N/A
### Testing Done

None, aside from rebuilding it and making sure the variables work as expected

### Changelog updated?

No.

### DIFF
--- a/sass/grid/_index.scss
+++ b/sass/grid/_index.scss
@@ -1,5 +1,5 @@
 /* Bulma Grid */
 @charset "utf-8";
 
-@forward "columns";
+@forward "columns-v2";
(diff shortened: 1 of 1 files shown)


reneedraper:
Issue fixed in commit fbbb85d

Issue #3895, opened 2024-09-06
Web Components Broken with 1.x Due to :root


kbarnesMCC:
### Overview of the problem
This is about the Bulma **CSS framework**
I'm using Bulma **version** [1.0.2]
My **browser** is: Firefox
I am sure this issue is **not a duplicate**

### Description

In the move to Bulma 1.x; it appears a handful of :root CSS selectors were added as part of the functionality.

Unfortunately, this has completely broken the ability to use Bulma as part of Web Components / ShadowDOM since stylesheets within this context don't respect :root (they use [:host](https://developer.mozilla.org/en-US/docs/Web/CSS/:host) instead).

We're a big fan of Bulma due to it's exclusion of Javascript which helps us use it in an highly encapsulated manner allowing for our large disparate teams to work together without clashing on each other. In fact, the lack of reliance on :root is the reason we chose this over other frameworks (e.g. Bootstrap)...However, this change to tightly couple to the root breaks our encapsulation.

Are there any plans to reconcile this to keep Bulma flexible or do we need to either roll our own Bulma (maintenance costs for us) or begin looking into alternatives?

### Steps to Reproduce

Download Bulma 1.x, investigate SASS files.

### Expected behavior

Ideally, observer Bulma is NOT using :root for css selectors OR supports both :root and :host.

### Actual behavior

Observe Bulma is now leveraging :root CSS selectors for core functionality.

gabor-kormany:
> What would the solution look like? Using :host instead of :root?

Can you maybe make it customisable? Make it ':root' or maybe ':root, :host' as default, but leave the ability to have it defined in scss/sass. We're also intending to use it within a micro-frontends where we have scoped import like:

```
.scopeClass {
  @import 'bulma';
}
```

gabor-kormany:
Opened a PR with the proposed fix: https://github.com/jgthms/bulma/pull/3906

gabor-kormany:
@jgthms ☝️

PR #3906, opened 2024-09-16
Issue-3895: Make variables root configurable

gabor-kormany:
This is a **bugfix**.

[Issue-3895](https://github.com/jgthms/bulma/issues/3895)

### Proposed solution
Make the css variables host configurable instead of hardcoded `:root`. This allows the usage where the css is expected to be scoped with a class and also in web components where the `:root` selector is not available, but `:host` is.

### Testing Done

I've tested this be creating a `test.scss` file with the following content: 
```
@use "sass" with (
  $variables-host: ':host'
);
```

Made the output css file variables under `:host` as expected.

### Changelog updated?

No.


gabor-kormany:
@jgthms Any plans to move forward with this?

avucic:
Yeah...It's a pain for me too. Any updates?

PR #3914, opened 2024-09-30
Add `$navbar-breakpoint` as a CSS variable

northeastprince:


northeastprince:
@jgthms could we get this merged?

jgthms:
You can't use a CSS variable in a media query. What your code is trying to output is the following:

```css
@media screen and (min-width: var(--bulma-navbar-breakpoint)) {
  /* Code */
}
```

But this is not possible.

PR #3937, opened 2024-11-21
Add .has-text-weight-extrabold

brendon:
This is an improvement.

We have a variable for `extrabold` font weight so it makes sense to have the typography utility class to use also. 

### Testing Done

None.

### Changelog updated?

Yes.


brendon:
Need anything else on this one? :)

brendon:
@jgthms, would this be able to be merged in?

brendon:
Thanks @jgthms :)

PR #3963, opened 2025-01-22
Improve colour loading versions of outlined buttons

ynte:
This is a ** improvement | bugfix **.

Right now the loading version of outlined buttons is broken for `is-link`, the other versions have been downgraded from from 0.x (in my opinion)

afbeelding

### Proposed solution

Overwrite the `loading-color` variable on buttons with the `is-outlined` modifier

afbeelding

### Tradeoffs

I'm new to the whole variable thing, but I hope overwriting a variable in a subclass is the way to go

### Testing Done

I tested all variants by making temporary changes to the documentation files, with and without `[disabled]`

### Changelog updated?

No.

### DIFF
--- a/sass/elements/button.scss
+++ b/sass/elements/button.scss
@@ -383,6 +383,12 @@ $no-palette: ("white", "black", "light", "dark");
     @include cv.register-vars(
       (
         "button-border-width": max(1px, 0.0625em),
+        "loading-color":
+        hsl(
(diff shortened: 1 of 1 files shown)


ynte:
For reference, this is how `.is-outlined.isloading` looked in `0.9.4`

https://github.com/user-attachments/assets/58fb17ea-d1d3-4441-8515-ae083473d136

PR #4001, opened 2025-07-07
Fix: Increase specificity for icon sizes within controls

### LINKED ISSUE #3985: .control.has-icons-left .icon is overwriting .is-small class
.control.has-icons-left .icon has a specificity of (0,3,0)
.is-small has a specificity of (0,2,0)

So we can't change the icons size by these components classes and have to make a custom class for the size instead.

Just adding a !important to the .is-small values would work, or adding more specificity like a father class and then a son class just like the other one is doing.

`
<div className="control has-icons-left">
    <input
        onChange={handleDaysInputChange}
        type="text"
        name={`${day}-title`}
        id={`${day}-title`}
        value={days[day]?.title || ""}
        className='input is-primary is-radiusless'
    />
    <span className="icon is-small is-left">
        <i class="fa-solid fa-pen"></i>
    </span>
</div>

Image

`

kght6123:
Resolves #3985

The .control.has-icons-left .icon selector (0,3,0) was overriding the .icon.is-small selector (0,2,0) due to higher specificity.

This change introduces more specific selectors for icon sizes within .control.has-icons-left and .control.has-icons-right contexts, ensuring that .is-small, .is-medium, and .is-large classes applied directly to icons within these controls correctly set their dimensions.

For example, .control.has-icons-left .icon.is-small (0,4,0) will now correctly override the base styles for icons in controls.

This is a **bugfix**.

### Proposed solution

This PR fixes a CSS specificity issue where size utility classes (like `.is-small`) applied to an `.icon` element within a `.control.has-icons-left` (or `.control.has-icons-right`) element were being overridden by the default icon styling for that control.

The solution adds more specific CSS rules in `sass/form/tools.scss` for `.icon.is-small`, `.icon.is-medium`, and `.icon.is-large` when they are descendants of `.control.has-icons-left` or `.control.has-icons-right`. These new rules ensure that the utility size classes take precedence, allowing developers to directly control the size of icons within input controls as intended.

### Tradeoffs

No significant tradeoffs are introduced with this solution. The change is highly targeted and increases specificity in a logical way that aligns with the intended use of utility classes. It doesn't add complexity or affect other components.

### Testing Done

The Sass code was successfully compiled using the project's build script (`npm run build-bulma`) to ensure there were no syntax errors. The logical change in CSS specificity directly addresses the reported bug. Visual confirmation would typically be done by inspecting an HTML page using these classes, but the fix is based on fundamental CSS specificity rules.

スクリーンショット 2025-07-09 23 42 25

スクリーンショット 2025-07-09 23 46 05

### Changelog updated?

No. (This is a minor bugfix).


kght6123:
@jgthms Is this possible to merge?

